PR #71, opened 2013-03-31
buffDuration

sebyg666:
this is an addon i made bassed on bTimers but i made it substancially more versitile.
it has an Extend.xml file where it checks for gear that increase spells/ja timers (to be populated by users) and also uses the status.xml for ALL buffs. please check it out


alphaONE2:
Please don't copy resources into your addon directory. Use `lua_base_path..'../../plugins/resources/status.xml'` instead.

While I'm at it, your Extend.xml file is _not_ valid XML.

Since you've modified Nitrous's code you may want to talk to him about merging your changes into his instead of creating an entirely new addon, but I'll leave that up to you.

sebyg666:
the reason Extend isnt really a correct format for xml is because i couldnt figure out how to use the xml.lua which im sure could parse XML's better then i did. i tried to get some help but no one replied so i was forced to use crazy coding to parse the xml's using stringhelper functions. the reason for the status.xml being in the folder is for the same reason as i was planing to edit it to reflect the same standard as extend.xml. BUT if someone can show me how to use xml.lua to parse my XML's im happy to make those changes. As for merging this with the original,, i would be happy to but for the moment id like it to be my own as ive changed so much of the original code that its not really the same addon anymore.

z16:
You should come to the #windower IRC channel ( irc.ffochat.com ), I'm usually available there. I don't check the forums very often I'm afraid.

sebyg666:
i just logged in ther

sebyg666:
just synckin the files up, but i did everything that you asked. i moved extend.xml into a data folder.
i also removed status.xml, and made it look for it using a filepath
and also chnaged alot of code
and tested all my coding too.
thanks for all the help.

alphaONE2:
OK, everything looks good.

PR #110, opened 2013-04-07
new addon: vwhl - void watch highlighter

giulianoriccio:
This plugin redirects the vw nm's vulnerabilities to the "tell" stream so that they can be held in the chat log using the chat filters' hold function (Config > Chat Filters > "Enter" on "Tell" until you get the green "hold" icon).
Moreover, the important informations are highlighted and a number is added to quickly understand the vulnerability level (1, 3, 5).
Supports EN, FR, DE clients.


alphaONE2:
I'm going to merge this since it looks like it will be useful for EN players, but I wanted to warn you that this will most likely fail on FR and DE clients. Diacritics are encoded using a custom extension to Shift_JIS, that SE uses.

giulianoriccio:
I see, then the string given to the event is different from what I expect?
Il giorno 07/apr/2013 04:16, "IryokuChevalier" notifications@github.com
ha scritto:

PR #143, opened 2013-04-11
Update model.lua

nitrous24:


z16:
Updated dpsclock.lua to include `return DPSClock`, which should fix the bug.

PR #151, opened 2013-04-12
Removed empty function

Podginator:


alphaONE2:
Before I merge this, what's up with the rolltracker#OnDisk.lua file?

Podginator:
Oh, sorry mate, that shouldn't be there. I think it was because github overwrote an open file and textwrangler saved two out. I've deleted it now at any Rate. Sorry for the late reply.

PR #164, opened 2013-04-16
Added sections by Krizz. 

Podginator:


z16:
I assume you put character names in the default settings file as an example for others? The default settings should be generic and not have any player-specific values. Change that so people loading the addon without checking the settings XML don't end up with wrong character data. You can put a detailed explanation (with examples, too) in the readme file, or even remind them on load that they need to adjust the settings, if no character info was found.

Also, consider not defining the entire XML in the code itself. The config module was designed specifically to create default settings from a table it gets as an optional parameter. Proper usage is documented [here](http://dev.windower.net/wiki/doku.php?id=dev:lua:writing_addons&#settings_loader).

Podginator:
I don't really know much about what has been added, these are from Krizz, I'm going to close this request and let him open a new one - He should be able to fix any problems. I've forwarded that message on to him. Cheers

PR #169, opened 2013-04-16
new addon: reive

giulianoriccio:
This plugin tracks exp, bayld, momentum scores and bonuses during a reive.

ss


z16:
Looking good, before I merge it: What does the file data/=) do? Is that meant to be there, or an oversight?

giulianoriccio:
it's there to make gitHub create the "empty" folder

z16:
You could put a default settings.xml in there, would probably be better than a garbage file, since a settings.xml is created anyway.

giulianoriccio:
done

giulianoriccio:
just a question about the settings.xml
what will happen when windower will update the addon? will it overwrite a customized settings.xml with the one in the repository?

z16:
No, it won't be overwritten.

PR #182, opened 2013-04-19
New Addon: DynamisHelper v.5

tehkrizz:
Features:
Creates a 30-second timer when a mob is procced.
Tracks currency obtained.
Creates a light luggage profile to lot all currency.

### DIFF
--- a/addons/dynamishelper/README.md
+++ b/addons/dynamishelper/README.md
@@ -1,0 +1,38 @@
+Authors: Krizz
+
+
+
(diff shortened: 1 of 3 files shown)


jerryhebert:
I'll merge this now but don't forget to add this (and cellhelper) to addons/addons.xml so that they get a description in the launcher.

tehkrizz:
Read my mind. I was going to ask how to do that today.

PR #192, opened 2013-04-20
DynamisHelperv1

tehkrizz:
Proc feature added


tehkrizz:
Forgot to add a setting. Will resubmit shortly.

PR #198, opened 2013-04-21
fixed config overwriting and tablehelper type comparison

giulianoriccio:


giulianoriccio:
meanwhile i committed a new plugin

PR #227, opened 2013-04-30
CellHelp Update - Timer

tehkrizz:


addons/cellhelp/cellhelp.lua
```diff
@@ -228,6 +238,13 @@ function initialize()
 	tb_set_text('salvage_box',' Lot order:  \n'..lotorder);
 end
 
+function checkzone()
+	currentzone = get_ffxi_info()['zone']:lower()
```
alphaONE2, line 242: If you pass in to_id and compare against zone IDs, instead of getting the zone name, you'll speed up the check and it will automatically work for jp, de and fr clients if you ever decide to add support; win-win.

I won't block the pull for this though.

tehkrizz, line 242: I'll note that and revisit when I'm implementing the next set of changes. I can also use the same thing on DynamisHelper.

PR #248, opened 2013-05-14
show both current/total bayld/exp

giulianoriccio:


Zireael:
Would it be possible to add a [Text Mode] options to this addon? And probably to Plasmon addon too, if it works the same.
What I mean is, the text box on screen constantly displaying Reive statistics is nice, but some people might not be interested in so much detail and find it distracting/obstructing. For me, a summary at the end of the Reive in chatlog showing current/total Bayld and XP earned would be enough.

Thank you for considering my suggestion.

giulianoriccio:
If you have noticed, there is already a report in the chat log at the end
of a run. It won't be hard to add a param to keep the window hidden.

PR #265, opened 2013-05-20
Addition of Settings file

Podginator:
Added setting file


z16:
Merging this, but hit me up in IRC when you're on, have a few notes :)

PR #276, opened 2013-05-23
added slips lib and slips/tmp items support

giulianoriccio:


giulianoriccio:
i've found a possible bug

PR #303, opened 2013-05-30
Submitting new addon ChatPorter.

Icinoki:


z16:
Just a quick question, this seems to look for both event_incoming_text and event_incoming_chat, and chat falls into both. Wouldn't that cause both to trigger? I'm not very familiar with your code, but couldn't you remove event_incoming_text in this case?

Icinoki:
I have it set to trigger on event_incoming_text for linkshell (self), party (self), tell (outgoing) and event_chat_message for linkshell (others), party (others), tell (incoming).  I split these because event_chat_message also gets player info and I'd have to find that separately with event_incoming_text.

PR #216, opened 2013-04-25
New Addon: TH Tracker

tehkrizz:
Displays latest TH message on screen.

PR #305, opened 2013-06-01
New Addon: Salvage2

tehkrizz:
Simple addon to track pathos remaining in Salvage 2.

### DIFF
--- a/addons/salvage2/README.md
+++ b/addons/salvage2/README.md
@@ -1,0 +1,23 @@
+Author: Krizz
+
+Version: .5
+
+Date: 20130601
+
+Salvage2
+
+Abbreviation: //s2
+
+Commands:
+* help - Shows a menu of commands in game
+* pos <x> <y> - Positions the pathos remaining box. Default location is 1000,250.
+* [hide/show] - Toggle display of the pathos remaining box.
+* timer [start/stop] - Will start or stop the 100 minute zone timer.
+* remove <pathos> - Removes the pathos from the list.
+
+To do:
+* Change zone comparison to use zone IDs.
+
+Known Issues:
+* May not always display the box upon zoning in until a pathos is removed.
+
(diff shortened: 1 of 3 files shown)


z16:
Gonna pull this, but a suggestion about settings. You should define default settings inside the file, and pass them to config.load, then you don't have to worry about the file not being created. In fact, if the file doesn't exist, it will create it with the default settings. And if you ever introduce new settings, it will automatically incorporate them into an existing `settings.xml` without screwing up any other user-defined values. That way you assure that all the values provided in the default table are guaranteed to be in the settings, unless overridden by user settings in the `settings.xml`.

Here's some sample code:

``` lua
defaults = {}
defaults.posx = 1000
defaults.posy = 250

function event_load()
    -- Do char-independent init stuff

    if get_ffxi_info().logged_in then
        initialize()
    end
end

function event_login()
    initialize()
end

-- By having this function called when being logged in, settings will load on a
-- char-dependent basis. This way everything in <global> can be overridden
-- by character tags.
function initialize()
    -- Not passing a file path for settings will default to 'data/settings.xml'.
    -- If default settings are provided, both the directory and the file will be
    -- created, if they don't exist.

    -- This will save the settings. This can be done at any time someone adjust
    -- the settings while the addon is running. This is needed to add new settings
    -- to the XML, such as when the addon is updated to include new features.
    -- If that isn't desired, it can be omitted. There's a chance this will be done
    -- automatically in the future, but as of now it isn't.
    settings:save()
end
```

z16:
Oh another thing:
Please make sure to edit the `addons.xml` file in the root `Windower/Lua/addons` directory. It needs to contain name, author, short description, bug tracker (for example your own GitHub repo) and support link (any FFXI site profile or email).

tehkrizz:
I was working on incorporating the config.load stuff on another addon, but I ran into some issues. I've been slacking lately with my add-ons, so I can't even remember what the problem was.

I will try updating the addons.xml file again. Last time I had a problem, so I still have other add-ons not listed as well. (https://github.com/Windower/Lua/pull/216)

z16:
Do that for those as well, please. That's the only way additional info for them will be displayed in the launcher. In future, addons won't be accepted anymore without that information filled out.

PR #366, opened 2013-06-27
Add color setting for NPC magic casting (enemies, automatons)

Zireael:
Add color setting for NPC magic casting (enemies, automatons). 
New setting is available in colors.xml file:
        < mobspellcol >0< /mobspellcol >
This will change color of spells casted by NPCs (primarily any mob, but also player's automaton).
To set the color of player casted magic, use currently present setting:
        < spellcol >0< /spellcol >

### DIFF
--- a/addons/battlemod/event_action.lua
+++ b/addons/battlemod/event_action.lua
@@ -219,7 +219,11 @@ function event_action(act)
 					if T{252,265,268,269,271,272,274,275,650}:contains(msg_ID) then
 						spell = 'Magic Burst '..spell
 					end
-					spell = color_it(spell,color_arr['spellcol'])
+					if actor_table['is_npc'] then
+						spell = color_it(spell or '',color_arr['mobspellcol'])
+					else
+						spell = color_it(spell or '',color_arr['spellcol'])
+					end
 				elseif table.contains(fields,'ability') then
 					ability = jobabilities[abil_ID]['english']
 					if msg_ID == 379 then ability = 'Magic Burst '..ability end
--- a/addons/battlemod/event_action.lua
+++ b/addons/battlemod/event_action.lua
@@ -220,9 +220,9 @@ function event_action(act)
 						spell = 'Magic Burst '..spell
 					end
 					if actor_table['is_npc'] then
-						spell = color_it(spell or '',color_arr['mobspellcol'])
+						spell = spell and color_it(spell, color_arr['mobspellcol']) or ''
 					else
-						spell = color_it(spell or '',color_arr['spellcol'])
+						spell = spell and color_it(spell, color_arr['spellcol']) or ''
 					end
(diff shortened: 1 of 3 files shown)


addons/battlemod/event_action.lua
```diff
@@ -219,7 +219,11 @@ function event_action(act)
 					if T{252,265,268,269,271,272,274,275,650}:contains(msg_ID) then
 						spell = 'Magic Burst '..spell
 					end
-					spell = color_it(spell,color_arr['spellcol'])
+					if actor_table['is_npc'] then
+						spell = color_it(spell or '',color_arr['mobspellcol'])
+					else
+						spell = color_it(spell or '',color_arr['spellcol'])
```
alphaONE2: I don't like this part, instead of passing an empty string to color_it if spell is nil, color_it should be modified to return an empty string if it's passed nil.

This is not bad enough to block the pull request, but this should still be addressed.

Zireael: Hi Iryoku, I have taken this code from another part of the file where Byrth handles enemy weaponskills this way and re-purposed it for magic.
 
I was wondering a bit about what the null value was doing here, but as my LUA knowledge is pretty much basic at this time, I left it as is, since my testing showed it works for the end user.
I showed my code to Byrth asking for opinion, but as his laptop got fried few days ago, he won't be able to do much with FFXI. He said to just push it.
 
I'll forward him your comment as I guess it will apply to other parts of the code as well, and my knowledge is nowhere near to amend the addon as a whole.
 
Thank you for your time!
 
 
Dnia 28 czerwca 2013 23:28 IryokuChevalier &lt;notifications@github.com&gt; napisał(a):
In addons/battlemod/event_action.lua:
&gt; @@ -219,7 +219,11 @@ function event_action(act)&gt; ifT{252,265,268,269,271,272,274,275,650}:contains(msg_ID) then&gt; spell = 'MagicBurst '..spell&gt; end&gt; - spell =color_it(spell,color_arr['spellcol'])&gt; + ifactor_table['is_npc'] then&gt; + spell =color_it(spell or '',color_arr['mobspellcol'])&gt; + else&gt; + spell =color_it(spell or '',color_arr['spellcol'])I don't like this part, instead of passing an empty string to color_it if spell is nil, color_it should be modified to return an empty string if it's passed nil.
This is not bad enough to block the pull request, but this should still be addressed.

z16: As a quick fix, you can replace it with this:

``` lua
spell = spell and color_it(spell, color_arr['spellcol']) or ''
```

Ideally, this should be integrated into color_it though.

alphaONE2:
OK, with the latest change I feel better about accepting this. I'll go ahead and merge this. I still think the change should be integrated into color_it to reduce code duplication, but at least there wont be spurious zero-length color sequences now.

PR #375, opened 2013-07-05
update addons.xml

Jandella:
Updated addons.xml with the shouthelper description


z16:
Gonna merge this, but you should also include a link to your own issue tracker under `<bugtracker></bugtracker>` (you may need to enable it in your GitHub settings first).

PR #411, opened 2013-08-02
Update texts.lua

nitrous24:
event_mouse = handle_mouse stopped working, changed it to register_event('mouse',handle_mouse)


nitrous24:
the plugin_manager.lua update was to change _addon.name to plugin_manager instead of "Plugin Manager" because it wasn't getting unloaded when doing //lua r plugin_manager (due to it looking for the _addon.name instead of the folder name i guess

PR #420, opened 2013-08-12
Create movable menu

Pr0c3ss0r:


addons/libs/menu.lua
```diff
@@ -0,0 +1,474 @@
...
+   windower.text.set_color('Menu_bar_'..MenuName, 255, 0, 0, 0)
+   windower.text.set_bold('Menu_bar_'..MenuName, 'true')
+   windower.text.set_text('Menu_bar_'..MenuName, ' \\cr')
+   windower.text.set_bg_visibility('Menu_bar_'..MenuName, true)
+   windower.text.set_font('Menu_bar_'..MenuName, 'Courier New', 9)
+   windower.text.set_visibility('Menu_bar_'..MenuName, false)
+
+   -- kick this function every .01 sec. (no return or return false cancel the functin loop)
+   menu.callback_list['onload_Menu.'..MenuName] = {
+      clock = os.clock() + .01,
```
alphaONE2: At this resolution you might as well not use the clock at all, we don't have a timing mechanism this precise.

addons/libs/menu.lua
```diff
@@ -0,0 +1,474 @@
...
+
+   windower.text.set_text('Menu_TB_'..MenuName, tb)
+
+   x,y = windower.text.get_location('Menu_TB_'..MenuName)
+
+   windower.text.set_location('Menu_cur_'..MenuName, x, y + menu.list[MenuName]['cur'] * 14)
+   windower.text.set_text('Menu_cur_'..MenuName, menu.list[MenuName]['Core_Text'][menu.list[MenuName]['cur'] + menu.list[MenuName]['scr']])
+end
+
+function menu.create(MenuName, Caption, opt, x, y, mx, sub_key)
```
alphaONE2: Please make these argument names consistent; use menu_name and caption.

addons/libs/menu.lua
```diff
@@ -0,0 +1,474 @@
...
+   windower.text.set_location('Menu_TB_'..MenuName, x, y)
+   windower.text.set_location('Menu_cur_'..MenuName, x, y + menu.list[MenuName]['cur'] * 14)
+
+   windower.text.set_location('Menu_scr_'..MenuName, x + x2, y)
+   local bar_x,bar_y = windower.text.get_location('Menu_bar_'..MenuName)
+   bar_y = bar_y - y1
+   windower.text.set_location('Menu_bar_'..MenuName, x + x2, y + bar_y)
+end
+
+function menu.mouse_event(ActionType, x, y, is_blocked)
```
alphaONE2: Ditto about the argument names.

addons/libs/menu.lua
```diff
@@ -0,0 +1,474 @@
...
+function menu.Refresh(MenuName)
+   local tb = ''
+
+   for i = 0, menu.list[MenuName]['mx'] do
+      tb=tb..menu.list[MenuName]['Core_Text'][i+menu.list[MenuName]['scr']]
+   end
+
+   windower.text.set_text('Menu_TB_'..MenuName, tb)
+
+   x,y = windower.text.get_location('Menu_TB_'..MenuName)
```
z16: `x, y` should be `local`

addons/libs/menu.lua
z16: I would call it `menu.new` to conform with the other libs we have.

Also, I would not handle them by name at all, but by reference to the object. You should not need to specify a name to work with them. You can solve this by having a copy of objects created with this module locally. See how the texts library does it.

What is `sub_key`?

addons/libs/menu.lua
```diff
@@ -0,0 +1,474 @@
...
+   end
+
+   menu.list[MenuName]['move'] = function(x, y)
+      menu.move(MenuName, x, y)
+   end
+
+   return menu.list[MenuName]
+end 
+
+function menu.combine(set1,set2)
```
z16: This should be a local helper function instead of a member. Members can be accessed from the outside (and should be made with that in mind).

``` lua
local function combine_tables(set1, set2) --[[code here]] end
```

z16: Actually, that should be the case for almost all functions in here. See how the config library handles local functions.

addons/libs/menu.lua
```diff
@@ -0,0 +1,474 @@
...
+
+function menu.move(MenuName, x, y)
+   local x1,y1 = windower.text.get_location('Menu_TB_'..MenuName)
+   local x2,y2 = windower.text.get_extents('Menu_TB_'..MenuName)
+
+   menu.list[MenuName]['x'] = x
+   menu.list[MenuName]['y'] = y
+
+   windower.text.set_location('Menu_Caption_'..MenuName, x, y)
+   y = y + 15
```
z16: Modifying original argument variables is not the best idea. This isn't a dealbreaker, but you should still call it something else.

addons/libs/menu.lua
```diff
@@ -0,0 +1,474 @@
...
+
+         menu.mouse_on = nil
+
+         for key, val in pairs(menu.list) do
+            x1,y1 = windower.text.get_location('Menu_TB_'..key)
+            x2,y2 = windower.text.get_extents('Menu_TB_'..key)
+
+            if x < x1 + x2+9 and x > x1 then 
+               if y < y1 + y2 and y > y1 - 15 then 
+
```
z16, line 358: You should define those offsets somewhere, so they can be easily adjusted. What are `9` and `15` here?

addons/libs/menu.lua
```diff
@@ -0,0 +1,474 @@
...
+      windower.text.delete('Menu_cur_'..key)
+      windower.text.delete('Menu_scr_'..key)
+      windower.text.delete('Menu_bar_'..key)
+      windower.text.delete('Menu_Caption_'..key)
+      menu.list[key] = nil
+   end
+end
+
+windower.register_event('unload', menu.event_unload)
+
```
z16, line 466: In general, for events you can just inline them. In this case it makes sense, since they're supposed to be used for events only and need not be local nor member functions.

``` lua
windower.register_event('unload', function()
    for key, val in pairs(menu.list) do
        --[[ code here ]]
    end
end)
```

addons/libs/menu.lua
```diff
@@ -0,0 +1,474 @@
...
+DISCLAIMED. IN NO EVENT SHALL Ilax / Pr0c3ss0r BE LIABLE FOR ANY
+DIRECT, INDIRECT, INCIDENTAL, SPECIAL, EXEMPLARY, OR CONSEQUENTIAL DAMAGES
+(INCLUDING, BUT NOT LIMITED TO, PROCUREMENT OF SUBSTITUTE GOODS OR SERVICES;
+LOSS OF USE, DATA, OR PROFITS; OR BUSINESS INTERRUPTION) HOWEVER CAUSED AND
+ON ANY THEORY OF LIABILITY, WHETHER IN CONTRACT, STRICT LIABILITY, OR TORT
+(INCLUDING NEGLIGENCE OR OTHERWISE) ARISING IN ANY WAY OUT OF THE USE OF THIS
+SOFTWARE, EVEN IF ADVISED OF THE POSSIBILITY OF SUCH DAMAGE.
+
+
+-- CreateMenu(Menu_name, Caption, Menu_Option_List, x, y, Max_Line_Show, sub_key *optional)
```
z16: What code is this?

addons/libs/menu.lua
```diff
@@ -0,0 +1,474 @@
...
+
+ ** Menu_name need to be unique.
+
+ Example:
+
```
z16, line 35: While examples are good, you should probably move this to the dev wiki. In the code you should comment functions more.

addons/libs/menu.lua
Pr0c3ss0r, line 466: i would assume is the same for mouse event?

alphaONE2, line 466: Correct.

addons/libs/menu.lua
Pr0c3ss0r, line 358: 15 is the size of caption (height), 9 is the width for 'X' (close text button)

alphaONE2, line 358: You should at least make them named constants then; ideally they should be configurable to allow different fonts to be used.

addons/libs/menu.lua
Pr0c3ss0r: menu_list = {
    sub  = {['english'] = "Oneiros Grip"},
    rear = {['english'] = "Cmd. Earring"},
    main = {['english'] = "Terra's Staff"},
    neck = {['english'] = "Twilight Torque"},
    lring= {['english'] = "Dark Ring"},
    rring= {['english'] = "Dark Ring"},
    back=  {['english'] = "Shadow Mantle"},
    legs=  {['english'] = "Crimson Cuisses"},
    waist= {['english'] = "Flume Belt"},
    lear=  {['english'] = "Moonshade Earring"},
    body=  {['english'] = "Bokwus Robe"},
    head=  {['english'] = "Dls. Chapeau +2"},
    feet=  {['english'] = "Estq. houseaux +2"}}

my_menu = menu.new("my_menu_name2", "Menu caption test", menu_list, 20, 20, 8, 'english')

is just in case a sub key exist in the list

Pr0c3ss0r:
All modification done, all function are now local.

my_menu = menu.new("my_menu_name", "Menu caption test", menu_list, 400, 400, 5)

then user can use: my_menu.refresh() or write('x = '..my_menu.x..' y = '..my_menu.y)

or

my_menu.move(100,230)

or add event (on_click, on_close, on_dbl_click, on_load, on_move)

my_menu['on_click']=function(this, index, val)
   write('Menu ['..this.menu_name..'] click ['..index..']('..val..')')
end

PR #449, opened 2013-08-20
4.1 update stuff and fixes

Podginator:


z16:
Will merge as soon as the `config` bug is fixed.

PR #584, opened 2013-12-26
create itemizer+readme

Jayjs20:


addons/itemizer/itemizer.lua
```diff
@@ -0,0 +1,95 @@
+_addon.name = 'itemizer'
+_addon.author = 'Ihina'
+_addon.version = '1.0.0.0'
+_addon.command = 'itemizer'
+
+require 'tablehelper'
+require 'chat'
+require 'logger'
+require 'actionhelper'
+config = require('config')
```
z16: I don't think you're using this anywhere, are you?

z16:
Please also add an entry to the `addons.xml` with all the info there (also do that for _Silence_ please).

PR #651, opened 2014-01-19
Update Battlemod to function more like older version

SnickySnacks:
Includes additional line templates, fixes for resists/misses/etc, update to rolltracker to correctly remove the battlemod roll line, etc.


SnickySnacks:
I don't know if I did this pull request thing right. 
This is the first time I tried doing it. :D

PR #691, opened 2014-01-26
Final updates to PetTP before release, addons.xml updated

SnickySnacks:
No known bugs at this point. I'm sure one will appear soon.

### DIFF
--- a/addons/PetTP/PetTP.lua
+++ b/addons/PetTP/PetTP.lua
@@ -1,27 +1,26 @@
-local config = require 'config'
+local config = require ('config')
 
-_addon = _addon or {}
 _addon.name = 'PetTP'
 _addon.author = 'SnickySnacks'
 _addon.version = '1.0'
 _addon.commands = {'ptp','pettp'}
 
-local petname = 'Unknown'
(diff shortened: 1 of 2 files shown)


addons/addons.xml
```diff
@@ -222,6 +222,13 @@
     <bugtracker>https://github.com/Windower/Lua/issues</bugtracker>
     <support>http://mibbit.com/#windower@irc.ffochat.com</support>
   </addon>
+ <addon>
+    <name>PetTP</name>
+	<author>SnickySnacks</author>
```
z16, line 227: What happened here? ;p

addons/PetTP/PetTP.lua
```diff
@@ -1,27 +1,26 @@
...
+petname = nil
+mypet_idx = nil
+current_hp = 0
+max_hp = 0
+current_mp = 0
+max_mp = 0
+current_hp_percent = 0
+current_mp_percent = 0
+current_tp_percent = 0
+tb_name = 'addon:pettp'
```
z16, line 17: You should switch to using the texts library instead of manually working with text objects.

PR #700, opened 2014-01-30
Submit Treasury

Jayjs20:


addons/Treasury/Treasury.lua
```diff
@@ -0,0 +1,177 @@
...
+    if param:remove(1) == 'treasury' then
+        command1 = param:remove(1)
+        command2 = param:remove(1)
+        passlot(command1, command2, S(param):map(tonumber))
+    end
+end)
+
+windower.register_event('load', function(...)
+    code.pass = settings.pass:map(table.get+{ids}..string.lower)
+    code.lot = settings.lot:map(table.get+{ids}..string.lower)
```
z16, line 37: These two should be outside of the load event. This is important so relogging will not screw up the sets.

addons/Treasury/Treasury.lua
```diff
@@ -0,0 +1,177 @@
...
+    forceCheck()
+end)
+
+windower.register_event('addon command', function(...)
+    local param = L{...}
+    if tostring(param) == '[clearall]' then
```
z16, line 43: `if param[1] == 'clearall' then`

addons/Treasury/Treasury.lua
```diff
@@ -0,0 +1,177 @@
...
+            code[func]:clear()
+            settings[func]:clear()
+        end
+        settings:save()
+    end
+end)
+
+windower.register_event('unhandled command', function(command1, command2, ...)
+    local param = L{...}
+    local isGlobal = false
```
z16, line 55: We don't use camelCase in our code. This is just a stylistic choice, but I'd advise to use underscores instead (`is_global`). The same goes for the other occurrences of this (`fromPacket`, `findIDs`, etc.).

addons/Treasury/Treasury.lua
```diff
@@ -0,0 +1,177 @@
...
+            settings:save()
+        
+        elseif command2 == 'list' then
+            log(command1, ':')
+            for item in settings[command1]:it() do
+                log(item)
+            end
+        end
+
+    elseif command1 == 'passall' then
```
z16, line 90: Why are these in `unhandled command` instead of `addon command`?

addons/Treasury/Treasury.lua
```diff
@@ -0,0 +1,177 @@
...
+            windower.ffxi.pass_item(slot_index)
+        end
+        
+    elseif command1 == 'lotall' then
+        log('Lot all')
+        for slot_index,item_table in pairs(windower.ffxi.get_items().treasure) do 
+            windower.ffxi.lot_item(slot_index)
+        end
+
+    elseif command1 == 'qq' then
```
z16, line 102: Doesn't make much of a difference, but this shouldn't be in live code ;p

addons/Treasury/Treasury.lua
```diff
@@ -0,0 +1,177 @@
...
+    if name == 'pool' then
+        return getPoolIDs()
+        
+    elseif name == 'seals' then
+        return S{2955, 2956, 1127, 2957, 1126}
+        
+    elseif name == 'currency' then
+        return S{1450, 1455, 1452, 1449, 1453, 1456}
+    
+    elseif name == 'junk' then
```
z16, line 155: I wouldn't call this junk, and I know a lot of people who wouldn't either :)

Should probably split it up in `crystals` and `geodes` or something.

addons/Treasury
```diff
@@ -0,0 +1,177 @@
+_addon.name = 'treasury'
```
z16, line 1: Something is wrong here... it thinks that `addons/Treasury` is a file for some reason. Can you try re-submitting the pull request? I don't wanna screw up our repo because of this.

z16:
This was added independently of this pull request.

PR #747, opened 2014-03-25
Fix all the broken spike messages.

SnickySnacks:
Hopefully no more ERROR 150 spikes.

### DIFF
--- a/addons/battlemod/battlemod.lua
+++ b/addons/battlemod/battlemod.lua
@@ -8,7 +8,7 @@ require 'parse_action_packet'
 require 'statics'
 res = require 'resources'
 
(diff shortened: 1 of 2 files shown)


SnickySnacks:
Added back in the special handling for all the spikes messages (retaliation now shows up properly instead of as "spikes" damage).
Fixed missed counterattacks showing up as ERROR 150.
Flipped the target/actor on all spikes messages.
Removed the act.action==nil error and just let it turn back into a regular message, until I can figure out what it actually is.

SnickySnacks:
Of course I realized this morning that this fix makes all non simplified spikes backwards (actor and target reversed). I'll fix it after work. :(

PR #769, opened 2014-04-17
Update rolltracker.lua

MilkMansKid:
Typo on Ninja and found Bust value for Save TP


z16:
Why was this closed? Is the data incorrect?

MilkMansKid:
The data is corrent. It shouldn't been closed I mis-clicked and couldn't
find a way to "open" it again. Sorry I'm new to the whole git thing.

PR #770, opened 2014-04-21
digger - initial commit

svanheulen:


addons/digger/digger.lua
```diff
@@ -0,0 +1,218 @@
...
+defaults.fatigue.remaining = 100
+defaults.accuracy = {}
+defaults.accuracy.successful = 0
+defaults.accuracy.total = 0
+
+settings = config.load(defaults)
+
+-- global constants
+
+fail_message = {
```
z16: Sets are prettier :)

addons/digger/digger.lua
```diff
@@ -0,0 +1,218 @@
...
+        settings.fatigue.remaining = 100
+    end
+    if count < 1 then
+        settings.accuracy.total = settings.accuracy.total + 1
+    end
+    if count < 0 then
+        settings.accuracy.successful = settings.accuracy.successful + 1
+    end
+    settings.fatigue.remaining = settings.fatigue.remaining + count
+    settings:save('all')
```
z16: Not sure if this was intended, but 'all' will not allow any character-specific settings. It should only be used when you intend to overwrite all other chars' settings.

addons/digger/digger.lua
```diff
@@ -0,0 +1,218 @@
...
+        local message_id = original:unpack('H', 11) % 0x8000
+        if fail_message[message_id] then
+            update_stats(0)
+            display_stats()
+        end
+    end
+end
+
+function digger_command(...)
+    if #arg == 1 and arg[1]:lower() == 'reset' then
```
z16, line 148: `arg` is nowhere defined. Did you forget a `local arg = {...}` at the start of the function?

addons/digger/digger.lua
svanheulen: Sorry, I'm still fairly new to Lua, can you point me to an example?

addons/digger/README
```diff
@@ -0,0 +1,33 @@
...
+
+Donations are welcome in the form of Alexandrite:
+    Acacia on Odin
+
+You can get in contact with me on FFOChat IRC.
+
+=============================== WARNING ===============================
+THIS CODE IS FOR RESEARCH PURPOSES ONLY. USING THIS CODE IS AGAINST THE
+POL/FFXI TERMS OF SERVICE AND MAY GET YOU BANNED. USE AT YOUR OWN RISK.
+=============================== WARNING ===============================
```
z16: This doesn't need to be in here. Using Windower alone is a violation of the ToS.

addons/digger/README
```diff
@@ -0,0 +1,33 @@
...
+
+This addon requires the Timers plugin to display your "recast" timer for
+digging. It will also display statistics about your dig accuracy,
+fatigue and remaining greens after each dig. The fatigue tracking will
+also work properly with the Blue Racing Silks bonus.
+
+To install this addon, create a folder called "digger" inside the
+Windower addons folder and copy the digger.lua file there. Your addon
+folder is most likely "C:\Program Files (x86)\Windower4\addons".
```
z16: This can be omitted, as the addon will be distributed through the launcher. It will only confuse the people who use the launcher, because they'll think they need to do more to install the addon.

addons/digger/README
```diff
@@ -0,0 +1,33 @@
...
+
+To load the addon, in the Final Fantasy XI input line, type:
+    //lua load digger
```
z16: Same as above.

addons/digger/digger.lua
svanheulen: I couldn't find any documentation on the dev wiki about the config library. I just saw other addons using it like this. What should I do to make it character specific?

addons/digger/digger.lua
svanheulen, line 148: It was my understanding that in Lua, the special variable `arg` was automatically defined when defining a function with `...` as the only argument. That allows you to accept a variable number of arguments. It does work like this, but it may not be the best way, do you have a suggestion?

addons/digger/README
svanheulen: That's just my standard "cover-my-ass" disclaimer, but I'll remove it.

addons/digger/digger.lua
nitrous24: just remove the 'all' part and make it settings:save()

addons/digger/digger.lua
z16: `require('sets')` at the top, then replace this with:

``` lua
fail_message = S{7208, 7250, 7227, 7536, 7194, 7682, 7198, 7256, 7216, 7035}
```

And later in the code use it like this:

``` lua
if fail_message:contains(msg) then
    -- Code
end
```

addons/digger/digger.lua
z16, line 148: Look at that, I had no idea, you're right. It's still... hacky, personally I wouldn't use it. I'd define it explicitly at the top of the function with `local arg = {...}`.

addons/digger/digger.lua
svanheulen: Much cleaner, thanks for the info.

PR #775, opened 2014-04-24
Create widescantool.lua

mujihina:


z16:
The addon is incomplete for two reasons:
1. It's missing the appropriate entries in the `addons.xml` file
2. It relies on an ID > name mapping for NPCs that isn't available yet

The first thing you can fix by just adjusting the file and committing it into the same branch. Once you do that I'll leave the pull request open until we find a solution to the mob name problem.

addons/widescantool/widescantool.lua
```diff
@@ -0,0 +1,349 @@
...
+    config.save(settings)
+end
+
+-- Change settings back to default
+function reset_to_default()
+	enable_mode = true
+    settings.global.alerts:clear()
+    settings.global.filters:clear()
+    settings.area.alerts:clear()
+    settings.area.filters:clear()
```
z16: You don't need these four lines if you do `settings:reassign(defaults)` below anyway, so you can remove those.

addons/widescantool/widescantool.lua
```diff
@@ -0,0 +1,349 @@
...
+
+-- Save settings
+function save_settings()
+    config.save(settings)
+end
+
+-- Change settings back to default
+function reset_to_default()
+	enable_mode = true
+    settings.global.alerts:clear()
```
z16: Mixing tabs and 3-space indentation is not a good idea ;p

Use 4 spaces consistently throughout the file.

addons/widescantool/widescantool.lua
```diff
@@ -0,0 +1,349 @@
...
+   	windower.add_to_chat (207, 'wst: \'wst aaa <name or pattern>\': Add Area Alert')
+   	windower.add_to_chat (207, 'wst: \'wst raa <name or pattern>\': Remove Area Alert')   
+   	windower.add_to_chat (207, 'wst: \'wst defaults\': Reset to default settings')   
+   	windower.add_to_chat (207, 'wst: \'wst toggle\': Enable/Disable all filters/alerts temporarily')
+    windower.add_to_chat (207, 'wst: \'wst pet\': Enable/Disable filtering of common mob pets')
+end
+
+-- Parse and process commands
+function wst_command (cmd, ...)
+   if (not cmd) then
```
z16: If you change it to `if (not cmd or cmd == 'h' or cmd == 'help') then` you can remove the block below.

addons/widescantool/widescantool.lua
```diff
@@ -0,0 +1,349 @@
...
+      show_syntax() 
+      return
+   end
+      
+   local zone_id = windower.ffxi.get_info().zone
+   local zone_name = zones[windower.ffxi.get_info().zone].name
```
z16: You should avoid calling `windower.ffxi.get_info()` twice, especially since you only need the value you just created, so you can use that instead.

You should also move this block below to where it's actually needed, this way the functions will always be called, even if the command entered doesn't require them. All `windower.ffxi.get_*` functions that return a table are kinda expensive, since they have to create a table and push all the elements into it one by one, even if you end up only using one of them.

addons/widescantool/widescantool.lua
```diff
@@ -0,0 +1,349 @@
...
+   -- Remove Area Alert
+   if (cmd == 'raa') then
+      windower.add_to_chat(200, 'wst rra: Removing: \"%s\" from area Alerts for %s':format(pattern, zone_name))
+	  if (settings.area.alerts:containskey(zone_id)) then
+		  settings.area.alerts[zone_id]:remove(pattern)
+ 		  windower.add_to_chat (207, 'wst rra: Current alerts for \"%s\": %s':format(zone_name,settings.area.alerts[zone_id]:tostring()))
+ 		  save_settings()
+  	  end  
+      return
+   end
```
z16: You may want to consider using a different syntax for commans, i.e. use full words as commands and allow optional shortening: `//wst area alert remove x`. That will help with remembering what the short syntax stands for, otherwise people just have to remember all the single tokens and tell the two `a`s apart.

z16: And speaking of which, the text you output has different strings than the commands used. Personally if you do it this way I recommend using the logger library and then type `log('Current alerts for "%s": %s':format(zone_name, settings.area.alerts[zone_id])`.

PR #776, opened 2014-04-24
Create mob_names.lua

mujihina:


alphaONE2:
Um... holy crap this one file more than doubles the size of the repository. It can't even be viewed on github. I think we need to come up with some other way to handle this.

z16:
Yeah, this won't go into the addon, although we can probably add it to the resources.

PR #777, opened 2014-04-26
Update addons.xml

mujihina:
Added entry for widescantool


alphaONE2:
In the future please just commit any additional changes to the same branch as your original pull request; any changes made before the pull request is merged or closed will automatically get picked up, so you don't need to submit multiples.

PR #816, opened 2014-05-12
digger - switched to using incoming text for stats

svanheulen:
fixes a bug with messages in non-chocobo zones triggering stat updates, but looses multilingual support.

### DIFF
--- a/addons/digger/digger.lua
+++ b/addons/digger/digger.lua
@@ -33,7 +33,7 @@ SOFTWARE, EVEN IF ADVISED OF THE POSSIBILITY OF SUCH DAMAGE.
 -- addon information
 
 _addon.name = 'digger'
-_addon.version = '1.3.5'
+_addon.version = '1.4.0'
 _addon.command = 'digger'
 _addon.author = 'Seth VanHeulen (Acacia@Odin)'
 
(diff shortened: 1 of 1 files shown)


z16:
I can pull this if you're sure, but imo this is not the right way to approach this. What's the bug with the packet message approach? Imo that should be fixed instead of doing it this way.

addons/digger/digger.lua
```diff
@@ -123,23 +120,21 @@ function check_zone_change(new_zone_id, old_zone_id)
     windower.send_command('timers d "Chocobo Dig Delay"')
 end
 
+function check_incoming_text(original, modified, original_mode, modified_mode, blocked)
+    if original:find('Obtained:') ~= nil and get_chocobo_buff() then
```
z16: Consider using `string.match` instead of `string.find` unless you're actually interested in the position:

``` lua
    if original:match('Obtained:') and get_chocobo_buff() then
```

svanheulen:
Yeah, I'm not sure what I was thinking, haha. It was just because I was getting matches on the message ID in zones where you can't ride a Chocobo. I just check the zone before checking the message ID now.

z16:
This works for me :)

z16:
On a side-note, while that's probably not it, the `% 0x8000` should be `- 0x8000`. It _could_ in theory cause issues, although in reality I don't think that's the case.

PR #885, opened 2014-06-02
Assorted improvements

cimmanon:
- Wardrobe is now checked for overload reduction gear
- Added missing overload reduction gear
- If Activate is down, it will try Deus Ex Automata instead

### DIFF
--- a/addons/autocontrol/autocontrol.lua
+++ b/addons/autocontrol/autocontrol.lua
@@ -168,6 +168,7 @@ function set_attachments_from_autoset(autoset,slot)
                 if windower.ffxi.get_ability_recasts()[205] == 0 then
                     windower.send_command('input /ja "Activate" <me>')
                 elseif windower.ffxi.get_ability_recasts()[115] == 0 then
+                    log('Activate is down, using Deus Ex Automata instead.')
                     windower.send_command('input /ja "Deus Ex Automata" <me>')
                 else
                     log('Unable to reactivate. Activate timer was not ready.')
--- a/addons/autocontrol/autocontrol.lua
+++ b/addons/autocontrol/autocontrol.lua
@@ -167,6 +167,8 @@ function set_attachments_from_autoset(autoset,slot)
             else
                 if windower.ffxi.get_ability_recasts()[205] == 0 then
                     windower.send_command('input /ja "Activate" <me>')
+                elseif windower.ffxi.get_ability_recasts()[115] == 0 then
+                    windower.send_command('input /ja "Deus Ex Automata" <me>')
                 else
                     log('Unable to reactivate. Activate timer was not ready.')
                 end
--- a/addons/autocontrol/maneuver.lua
+++ b/addons/autocontrol/maneuver.lua
@@ -29,15 +29,15 @@ SOFTWARE, EVEN IF ADVISED OF THE POSSIBILITY OF SUCH DAMAGE.
 
 local texts = require 'texts'
 threshItems = T
-    {['16281']=5  -- Buffoon's Collar
-    ,['16282']=5  -- Buffoon's Collar +1
-    ,['11101']=40 -- Cirque Farsetto +2
-    ,['11201']=20 -- Cirque Farsetto +1
-    ,['14930']=5  -- Pup. Dastanas
-    ,['15030']=5  -- Pup. Dastanas +1
-    ,['27960']=5  -- Foire Dastanas
-    ,['27981']=5  -- Foire Dastanas +1
-    ,['28634']=5  -- Dispersal Mantle
+    {[16281]=5  -- Buffoon's Collar
+    ,[16282]=5  -- Buffoon's Collar +1
+    ,[11101]=40 -- Cirque Farsetto +2
+    ,[11201]=20 -- Cirque Farsetto +1
+    ,[14930]=5  -- Pup. Dastanas
+    ,[15030]=5  -- Pup. Dastanas +1
+    ,[27960]=5  -- Foire Dastanas
+    ,[27981]=5  -- Foire Dastanas +1
+    ,[28634]=5  -- Dispersal Mantle
     }
 heat = T{}
 heat.Fire = 0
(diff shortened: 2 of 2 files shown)


addons/autocontrol/autocontrol.lua
```diff
@@ -167,6 +167,8 @@ function set_attachments_from_autoset(autoset,slot)
             else
                 if windower.ffxi.get_ability_recasts()[205] == 0 then
                     windower.send_command('input /ja "Activate" <me>')
+                elseif windower.ffxi.get_ability_recasts()[115] == 0 then
+                    windower.send_command('input /ja "Deus Ex Automata" <me>')
```
nitrous24, line 172: Before i merge this, I think if you're going to add deus ex automata to this, it should announce which it's using. Other than that this all looks fine

Edit: On second thought (Arcon's thought not mine), I think this should be an optional thing, rather than all the time. Adding a value to the settings whether or not to do it would be optimal. Just because of the way an automaton is summoned under it's effects.

cimmanon:
Ok, done.  Looks like I missed the wardrobe addition in the initial PR, so that's there as well.

addons/autocontrol/autocontrol.lua
```diff
@@ -167,6 +167,9 @@ function set_attachments_from_autoset(autoset,slot)
             else
                 if windower.ffxi.get_ability_recasts()[205] == 0 then
                     windower.send_command('input /ja "Activate" <me>')
+                elseif windower.ffxi.get_ability_recasts()[115] == 0 then
+                    log('Activate is down, using Deus Ex Automata instead.')
+                    windower.send_command('input /ja "Deus Ex Automata" <me>')
```
z16, line 172: This should be optional behavior (enabled through the settings file) and default to false. Can leave the chat output if you want, but I guarantee you people will get annoyed by it quickly :)

addons/autocontrol/maneuver.lua
```diff
@@ -227,20 +237,15 @@ function get_threshold()
         local basethreshold = 30
         local equip = T(windower.ffxi.get_items()['equipment'])
         local inv = T(windower.ffxi.get_items()['inventory'])
+        local wardrobe =  T(windower.ffxi.get_items()['wardrobe'])
```
z16: This may not be an issue (I don't know how often the function is called), but this is very expensive. Instead you should do:

``` lua
local items = windower.ffxi.get_items()
```

Then use `items.equipment`, `items.inventory` and `items.wardrobe` instead. This way `get_items` gets called three times, and it's a really expensive function, as it has to push all items in all your inventories on the Lua stack.

addons/autocontrol/maneuver.lua
```diff
@@ -227,20 +237,15 @@ function get_threshold()
...
-        if equip.neck ~= 0 then
-            if threshItems[tostring(inv[equip.neck].id)] ~= nil then
-                bonus = bonus + threshItems[tostring(inv[equip.neck].id)]
+        local slots = {'hands', 'body', 'neck', 'back'}
+        for i, s in ipairs(slots) do
+            if equip[s] ~= 0 then
+                local item = threshItems[tostring(inv[equip[s]].id)] or threshItems[tostring(wardrobe[equip[s]].id)]
+                if item ~= nil then
+                    bonus = bonus + item
+                end
```
z16, line 246: `threshItems` should be changed to have integer keys instead of converting the IDs to strings all the time.

addons/autocontrol/maneuver.lua
```diff
@@ -28,7 +28,17 @@ SOFTWARE, EVEN IF ADVISED OF THE POSSIBILITY OF SUCH DAMAGE.
...
+threshItems = T
+    {['16281']=5  -- Buffoon's Collar
+    ,['16282']=5  -- Buffoon's Collar +1
+    ,['11101']=40 -- Cirque Farsetto +2
+    ,['11201']=20 -- Cirque Farsetto +1
+    ,['14930']=5  -- Pup. Dastanas
+    ,['15030']=5  -- Pup. Dastanas +1
+    ,['27960']=5  -- Foire Dastanas
+    ,['27981']=5  -- Foire Dastanas +1
+    ,['28634']=5  -- Dispersal Mantle
```
z16: If you drop the single quotes around the IDs here, you can access them by ID later, without having to convert it to a string. See note below.

cimmanon: Using strings was part of the original code, but I've converted them to integers anyway.

addons/autocontrol/autocontrol.lua
ghost, line 172: Automatically activating should have been optional behavior in the first place.  But with that not being the case, this change is in line with the current behavior/expectations.  It isn't like there's any other options when activate isn't ready; right now you are always just going to use deus ex when activate isn't up or else you would have been annoyed at it using activate on you anyways.

z16, line 172: You should change it for activating as well then, again default to false.

cimmanon, line 172: I think you spelled "thank you for your contribution" incorrectly.

z16, line 172: I'm pretty sure I didn't, my spelling is rather accurate. If I didn't appreciate your contribution I wouldn't take the time to comment here and just close the pull request instead. But you're adjusting the functional behavior of an addon in common use, hence the new behavior should not be forced on people who don't follow the change log religiously.

cimmanon, line 172: The suggestion you made (make automatic activation configurable and off by default) would be adjusting the functional behavior of an addon in common use.  Every PUP who uses this addon _always_ uses Deus Ex Automata whenever they get the "Activate is down" message because there is nothing else they _can_ do.  The fact that the addon _didn't_ do this before was an inconvenience.

You have a strange way of showing appreciation, expecting me to fix things that weren't even mine to begin with (strings for keys, multiple get items function calls, etc.).

z16, line 172: Neither did I expect you to do it, nor did I blame you for them, I _requested_ you added it, since you were already making adjustments to it and I don't know the addon very well. I now adjusted it myself, if you could confirm that the addon is still working now I'd appreciate it.

I defaulted Deus Ex Automata usage to `false` and Activate usage to `true` to accommodate current users. After a while I can safely change the default for that to `false` and only new users will default with the `false` setting, as previous users will have settings files generated with the Activate toggle being `true`.

PR #924, opened 2014-06-15
initial commit

ProjectTako:


Byrth:
ASE will never be part of the repository.

Byrth:
As a side note, you should really blank the target index in your outgoing standard client update packet so SE can't see you targeting unspawned monsters..

PR #985, opened 2014-07-25
libs/strings.lua todec() neg number handling

Antinym:
Added negative number handling for string.todec()

### DIFF
--- a/addons/libs/strings.lua
+++ b/addons/libs/strings.lua
@@ -219,6 +219,7 @@ function string.todec(numstr, base)
     -- Create a table of allowed values according to base and how much each is worth.
     local digits = {}
     local val = 0
+	local numstr_neg
     for c in ('0123456789ABCDEFGHIJKLMNOPQRSTUVWXYZ'):gmatch('.') do
         digits[c] = val
         val = val + 1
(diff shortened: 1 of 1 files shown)


z16:
In the future please use white spaces instead of `\t` for indentation. I'll fix that and re-commit it this time.

z16:
I removed the entire function because it does the same as `tonumber`. Honestly, I'm not sure why it was even there.

`tonumber` still has an issue with negative numbers and bases other than 10, but I'll fix that today.

Antinym:
Good deal.  I looked at it again and unless there's a need to work with bases higher than 36 I don't see a need for `todec`.  I switched to using `tonumber` a few hours ago.

z16:
Fixed the non-base-10 issues with `tonumber` in _LuaCore 1.7.9.2_, which is currently on dev. Feel free to test it and report if you find anything wrong with it. I had to edit actual Lua code to fix this, so I can't be entirely sure I didn't break something else in the process.

Antinym:
Tested various combinations of bases and string numbers as well as base 10 decimals.  Everything returned the expected results.

z16:
Alright, thanks for that!

PR #1012, opened 2014-08-14
find (Lua version of the Find plugin)

Skudo:


z16:
There already is a Lua version of _Find_ called _FindAll_. It does almost everything _Find_ does but has some additional features, namely searching other characters as well. Does this do anything that _FindAll_ doesn't do? Otherwise I wouldn't wanna merge it, as there's no use to have duplicate addons.

The one feature the _Find_ plugin does that _FindAll_ doesn't yet have is graphical item tracking. But as far as I can see your version doesn't have it either, and tbh I would prefer if it was added to _FindAll_ instead.

Skudo:
Ohay, I will work on adding the graphical tracking feature to _FindAll_ then.

Antinym:
I was/(am in the back of my head) working on a findall/find merge-rewrite.  I wasn't satisfied with what I came up with so it's been on hold.  You're more than welcome to use any/all of what I've got so far.
find has a working tracker
https://github.com/Antinym/Lua-trainwreck/tree/4.1-dev/addons/find 

findAll has various changes to how it's used by the end-user.
https://github.com/Antinym/Lua-trainwreck/tree/4.1-dev/addons/findAll

PR #1015, opened 2014-08-15
AutoRA feature add

banggugyangu:
Added feature to halt at 1000 TP


z16:
This feature should be optional, defaulting to `true`.

PR #1020, opened 2014-08-15
Bugfix

banggugyangu:
Bugfix

### DIFF
--- a/addons/AutoRA/AutoRA.lua
+++ b/addons/AutoRA/AutoRA.lua
@@ -122,6 +122,21 @@ windower.register_event('action',function (act)
 					end
 				end
 			end
+		else
+			if category == 2 then
+				if auto == 1 then
+					if  player.status == 1 then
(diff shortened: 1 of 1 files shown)


addons/AutoRA/AutoRA.lua
```diff
@@ -122,6 +122,21 @@ windower.register_event('action',function (act)
...
+					if  player.status == 1 then
+						auto = 1
+					elseif  player.status == 0 then
+						auto = 0
+					end
+				end
+			
+				if auto == 1 then
+					windower.send_command('@wait 1.5;input /shoot <t>')
+				elseif auto == 0 then
```
z16, line 137: Is this clause meant to be empty?

banggugyangu, line 137: Yeah,  that's as normal  the bug was an oversight forgetting to accomodate
when TP wasn't above 1000

PR #1033, opened 2014-08-25
Update resources.lua

mujihina:
Flag04 = This is the flag to indicate that the item can be send to other chars of same account (even if it's EX)


z16:
I hope we can find a better name for it in the future though ;D

mujihina:
To avoid confusion, it's the same name used by BG for this flag: CanSendPOL

PR #1034, opened 2014-08-27
Update parse_action_packet.lua

mujihina:
Not often....but somewhat regularly I get a runtime error on line 370 (371 now) about pairs() complaining that targs is nil instead of a table.


z16:
Were you able to discern a pattern when this occurs? It would be better to fix the underlying problem than to ignore the `nil` case.

PR #1044, opened 2014-09-06
Update scoreboard.lua

geno3302:
These changes use the change I submitted to the actions.lua file. It will not work without that submission.

Created function FindPetOwnerName with logic to associate pet with owner. Any actions that can be done by a pet now use CreatePlayerName(action) instead of action:get_actor_name()

In spell_finish I changed the id 265 to 264. 264 was the message I was getting on sub targets when casting AoE black magic. This should fix the issue where AoE magic damage was not being processed (at least for black magic, I think pet AoE is still incorrect)

### DIFF
--- a/addons/scoreboard/scoreboard.lua
+++ b/addons/scoreboard/scoreboard.lua
@@ -308,9 +308,12 @@ local function update_dps_clock()
     local player = windower.ffxi.get_player()
     local pet
     if player ~= nil then
-    local pet_index = windower.ffxi.get_mob_by_id(player.id).pet_index
-        if pet_index ~= nil then
-            pet = windower.ffxi.get_mob_by_index(pet_index)
+        local player_mob = windower.ffxi.get_mob_by_id(player.id)
+        if player_mob ~= nil then
+            local pet_index = player_mob.pet_index
+            if pet_index ~= nil then
+                pet = windower.ffxi.get_mob_by_index(pet_index)
+            end
         end
     end
     if player and (player.in_combat or (pet ~= nil and pet.status == 1)) then
@@ -355,9 +358,12 @@ windower.register_event('action', function(raw_action)
     local player = windower.ffxi.get_player()
     local pet
     if player ~= nil then
-        local pet_index = windower.ffxi.get_mob_by_id(player.id).pet_index
-        if pet_index ~= nil then
-            pet = windower.ffxi.get_mob_by_index(pet_index)
+        local player_mob = windower.ffxi.get_mob_by_id(player.id)
+        if player_mob ~= nil then
+            local pet_index = player_mob.pet_index
+            if pet_index ~= nil then
+                pet = windower.ffxi.get_mob_by_index(pet_index)
+            end
         end
     end
     if not player or not (windower.ffxi.get_player().in_combat or (pet ~= nil and pet.status == 1)) then
--- a/addons/scoreboard/scoreboard.lua
+++ b/addons/scoreboard/scoreboard.lua
@@ -358,17 +358,17 @@ windower.register_event('action', function(raw_action)
                     -- hit, crit
                     if subaction.message == 1 or subaction.message == 67 then
                         if subaction.message == 67 then
-                            dps_db:add_m_crit(target:get_name(), CreatePlayerName(action), subaction.param)
+                            dps_db:add_m_crit(target:get_name(), create_player_name(action), subaction.param)
                         else
-                            dps_db:add_m_hit(target:get_name(), CreatePlayerName(action), subaction.param)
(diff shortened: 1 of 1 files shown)


addons/scoreboard/scoreboard.lua
```diff
@@ -457,6 +457,32 @@ windower.register_event('action', function(raw_action)
     end
 end)
 
+function FindPetOwnerName(action)
```
z16: Please try to adhere to the current naming scheme, i.e. use lower case letters and underscore instead of PascalCase.

addons/scoreboard/scoreboard.lua
```diff
@@ -457,6 +457,32 @@ windower.register_event('action', function(raw_action)
...
+	local pet = windower.ffxi.get_mob_by_id(action:get_id())
```
z16: A few lines use tabs instead of spaces, please adjust that to use 4 spaces for each indentation level.

jerryhebert:
There are a couple of other issues besides what Arcon said:
- First, you're essentially changing the 'primary key' of a user, which is its name. This means we'd never be able to uniquely identify a player correctly. Since the key for players is their name, you're making multiple entries for a each player (one for each player+pet or lack of pet combo). A beastmaster would have many instances of player+pet in this scenario and a dragoon would have two. This messes up the db.
- Second, you didn't modify any of the display code and that format is carefully chosen. The columns are going to collide and mess up the readabiltyof the HUD. 

I would not suggest merging this change.

One idea for getting what seems to be the desired effect would simply be to add an option into the config like "combine player and pet damage". Then you could combine them on the fly _only in the display_ so that the separate reports won't be compromised. You could also just aggregate all pets into "pet" so that the HUD would read "Joeschmoe+Pet" or something like that, which would be a more reliable way to preserve the formatting correctness.

geno3302:
The key was a string it's still that same exact string for all players. Genoxd is still Genoxd. What changed was now Shiva is "Shiva (Genoxd)" instead of all the shivas being lumped together under the same string. I haven't changed any of the core code, I only changed what we call pets and the naming is consistent.

Unless the complaint is you would get Player, Shiva (PlayerName), Ifrit (PlayerName) and if so, that was my intent. I want to know how my pets parse and I want to know how each one parses. Also, based off that we currently aren't able to "uniquely identify a player correctly".

geno3302:
I'll work on fixing the display code, any idea what the max character name length is?

addons/scoreboard/scoreboard.lua
```diff
@@ -457,6 +477,31 @@ windower.register_event('action', function(raw_action)
     end
 end)
 
+function find_pet_owner_name(action)
+    local pet = windower.ffxi.get_mob_by_id(action:get_id())
+    local party = windower.ffxi.get_party()
+    
+    local name = ''
```
jerryhebert: Having this default to nil seems slightly more reliable (and having line 500 check nil instead) so that you can determine success/failure based on type rather than value.

addons/scoreboard/scoreboard.lua
```diff
@@ -457,6 +477,31 @@ windower.register_event('action', function(raw_action)
...
+            if member.mob.pet_index and member.mob.pet_index> 0 and pet.index == member.mob.pet_index then
+				name = member.mob.name
+                break
+            end
+        end
+    end
+    return name
+end
+
+function create_player_name(action)
```
jerryhebert: Maybe consider renaming this to create_mob_name() or something since it doesn't only apply to players?

jerryhebert:
My mistake, I got confused by the name of the function and didn't look closely enough. Thanks for the updates, btw. I obviously haven't touched this in an eternity so it's nice to see someone willing to pull up their sleeves and spice things up.

As for the name length, it looks like I went with 16 but I don't remember where I got that. I'll defer to @z16 since he probably knows for sure.

z16:
Player names have a max length of 15, monster names a max length of 23.

z16:
Any update on this? Is it pull-ready?

geno3302:
No sorry, I've been busy at work. I should have time to clean it up this weekend. I'm going to mimic the names for skillchains and do something like Shiva (Gen.) Shiva (Cas.) and so on. Just have to make sure the max mob length (23) + the (123) doesn't mess up the UI.

geno3302:
I think this is ready for merge now, but let me know if there's anything else.
I liked the idea of having a setting for combing master + pet damage into one but that'll mess up the other info like ACC and such. 
Maybe have a setting so it just combines the pets though like
PlayerName
PlayerName's Pets

Not sure when I'll have time to do more, but I've also gotten requests for estimating Helix damage. If someone knows what packet the "effect wears off" message comes in it should be possible to calculate helix damage to within 1 tock.

PR #1058, opened 2014-09-12
Update fields.lua

mujihina:


addons/libs/packets/fields.lua
```diff
@@ -1411,6 +1411,16 @@ fields.incoming[0x034] = L{
     {ctype='data[2]',           label='_junk1'},                                -- 31   Always 00s for me
 }
 
+--- When messages are fishing related, the player is the Actor.
+--- For some areas, the most significant bit of the message ID is set sometimes.
+-- NPC Chat
+fields.incoming[0x036] = L{
+    {ctype='unsigned int',      label='Actor',                fn=id},             -- 04
+    {ctype='unsigned short',    label='Actor Index',          fn=index},          -- 08
+    {ctype='unsigned short',    label='Message ID',           fn=function (val) if val > 0x8000 then return val - 0x8000 else return val end end},           -- 0A
```
z16: This should be adjusted:

``` lua
    {ctype='bit[15]',           label='Message'}                                -- 0A
    {ctype='boolbit',           label='_unknown1'}                              -- 0B
```

I'm only assuming that the MSB is a bool, but that's usually the case in such situations.

addons/libs/packets/fields.lua
```diff
@@ -1411,6 +1411,17 @@ fields.incoming[0x034] = L{
...
 }
 
+--- When messages are fishing related, the player is the Actor.
+--- For some areas, the most significant bit of the message ID is set sometimes.
+-- NPC Chat
+fields.incoming[0x036] = L{
+    {ctype='unsigned int',      label='Actor',                fn=id},             -- 04
+    {ctype='unsigned short',    label='Actor Index',          fn=index},          -- 08
+    {ctype='bit[15]',           label='Message ID'},                              -- 0A
+    {ctype='bit',            label='_unknown1'},                               -- 0B
```
z16: The indent is messed up here :)

PR #1068, opened 2014-09-26
Leading zeros for time display.

Skyrant:
Time is now shown with leading zeros.

7:9 is now 07:09
10:5 is now 10:05

### DIFF
--- a/addons/gametime/gametime.lua
+++ b/addons/gametime/gametime.lua
@@ -132,7 +132,7 @@ defaults.time.bg.red = 100
 defaults.time.bg.green = 100
 defaults.time.bg.blue = 100
 defaults.time.bg.visible = true
-
+defaults.time.zero = 'off'
 
 defaults.days = {}
 defaults.days.pos = {}
@@ -242,6 +242,12 @@ config.register(settings, function()
 		gt.delimiter = '\n'
 	end
 
+	if settings.time.zero == 'on' then
+		gt.zero = 'on'
+	else
+		gt.zero = 'off'
+	end
+
     local info = windower.ffxi.get_info()
     if info.logged_in then
         day_change(info.day)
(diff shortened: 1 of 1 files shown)


addons/gametime/gametime.lua
```diff
@@ -272,8 +272,8 @@ function default_settings()
 end
 
 windower.register_event('time change', function(new, old)
-	gt.hours = (new / 60):floor()
-	gt.minutes = new % 60
+	gt.hours = string.format("%d",(new / 60):floor()):zfill(2)
```
z16: Should change this to the following for efficiency:

``` lua
    gt.hours = '%02d':format((new / 60):floor())
    gt.minutes = '%02d':format(new % 60)
```

This way strings don't have to be reallocated four additional times and it's shorter code. But honestly, you should probably make this toggleable via settings. The game itself does not use leading zeroes, so some people might be confused by this.

Edit: Actually, I now see how this makes sense because of the minutes being truncated otherwise. I'll pull it once you change those two lines.

z16: Actually, now that I think about it, you should revert this back entirely. This changes the type (and value) of `gt.hours` and `gt.minutes`, which used to be numbers but are now strings. You should change the display, not the data, i.e. the point where `gt.hours` and `gt.minutes` are displayed in the text box should be adjusted by this.

Skyrant: Problem is that the windower api reference is incomplete and there is no description anywhere to be found how texts.new() works or the texts class in general. If i knew how it works i could maybe adjust this but since no one seems to care to maintain the dev wiki i just leave this as is for myself as i can not stand the time without leading zeros.

EDIT: Also can not use format on the output because the variables are NIL before the first time tick so it throws and error. The values are converted to strings anyway.

z16: Actually, it's not possible with the `texts` library, I wasn't aware that _GameTime_ uses it. I'll add functionality for that in a bit. In the meantime you can change this to what I said above, that should work as an intermediate solution.

z16:
I rewrote the addon a bit after I added a formatting option to the texts library. By the way, while it's not a great documentation you can check the source code for the texts library. The way `texts.new` works is explained in there and gives an example of how to use it as well.

PR #1074, opened 2014-10-05
posible fix for file_unload function

smd111:
this is my idea for fixing the file unload issue i have been having

Info:
with the current fix that on the dev version;
it works but the job info does not update to the new job till after get_sets is run

so this is my fix i cam up with on my own


Byrth:
I pushed a new refresh.lua to -dev that should make this work the way you think it should.

smd111:
yes the load unload works like i would like on dev

 but
i just tested it and now 

```
the command gs l <file>
```

is not working

Byrth:
Changed it, now it should work.

smd111:
now its working correctly 
cant wait till its pushed live so i can remove the custom text.lua files from my includes as well as the texts = require('includes/gstexts')

Byrth:
We're going to push it live now.

smd111:
sweet
ill give it a test tomorrow and if i find any bugs ill post a new issue

PR #1083, opened 2014-10-09
GearSwap pet_update changes

svanheulen:
I figured out some more information about the 0x67 packet and it's not really a reliable way of determining when you gain or lose a pet. It's really just a status update for you pet but it doesn't guarantee that information about that pet is available yet. The 0x37 packet it a reliable way to get your current pet index and it also allows for the pet_change event to properly trigger when zoning. I also changed the refresh of the global pet table to only set pet.isvalid true when there is actually pet information available because it doesn't make any sense otherwise.


svanheulen:
Also, these changes mean that pet_change wont trigger when calling your fellow npc or trusts. Which makes sense because they're not pets and the pet table wouldn't apply to them anyway.

PR #1087, opened 2014-10-16
Initial Version IRC v0.1

Skyrant:
IRC LUA AddOn for Windower that puts IRC chat functionality into FFXI. Can be used with Twitch.tv to chat with viewers or any other IRC server. Currently only one irc channel is supported.

Early version with basic functionality. Comments and Suggestions are welcome.


addons/irc/data/settings.xml
```diff
@@ -0,0 +1,18 @@
+<?xml version="1.1" ?>
+<settings>
+	<global>
```
z16, line 3: Please convert all tab indents to spaces to conform to our coding standards (in every file).

addons/irc/data/settings.xml
```diff
@@ -0,0 +1,18 @@
...
+        <server>
+            <serveraddr>irc.twitch.tv</serveraddr>
+            <serverport>6667</serverport>
+        </server>
+        <login>
```
z16, line 8: Imo this should be nested under the server info. you may want to have different login data for different servers, if you wanna add that option in the future. It may require a different approach to the settings file though. Just a thought.

addons/irc/irc.lua
```diff
@@ -0,0 +1,195 @@
...
+_addon.commands = {'i','irc'}
+
+require('chat')
+require('logger')
+require('strings')
+require('maths')
+require('tables')
+
+config = require('config')
+settings = config.load(defaults)
```
z16, line 29: The `defaults` table is not defined anywhere. You should define the default settings table before this line somewhere (usually called `defaults`, as here) and then remove the `settings.xml` file entirely. It will be generated on startup, if it's missing. That way you don't need to maintain an extra file, which will _not_ get updated, so you might break the addon if you add a new settings value in the future.

addons/irc/irc.lua
```diff
@@ -0,0 +1,195 @@
...
+	connecttime = os.time()
+
+	serveraddr = settings.server.serveraddr
+	serverport = settings.server.serverport
+
+	if serveraddr and serverport then
+		-- eventually do something here
+	else
+		log("setting default twitch server and port")
+		serveraddr = 'irc.twitch.tv'
```
z16, line 42: I really think it should abort instead of defaulting to _Twitch_. That will be entirely unexpected to users just wanting an IRC addon and suddenly they'll find themselves in a chatroom they had no intention of joining.

addons/irc/irc.lua
```diff
@@ -0,0 +1,195 @@
...
+		serverport = '6667'
+	end
+	serverport = tonumber(serverport)
+	server = socket.tcp()
+	status = server:connect(serveraddr, serverport)
+	-- twitch irc servers
+	-- irc.twitch.tv:6667 -- use this for your twitch channel
+	-- server:send(perform)
+	server:send("TWITCHCLIENT 2\r\n")
```
z16, line 51: This should also be removed. You could introduce a special case handling mechanism that handles connections to certain services like _Twitch_. If you make this as addon as _Twitch_-oriented as it is now, you should probably incorporate that into the name to not mislead people.

addons/irc/irc.lua
```diff
@@ -0,0 +1,195 @@
...
+	server:send("PASS "..settings.login.password.."\r\n")
```
z16, line 52: If a password is not set, this should not be sent.

addons/irc/socket.lua
```diff
@@ -0,0 +1,133 @@
+-----------------------------------------------------------------------------
```
z16, line 1: Not really sure why you added this file, it's already at `Windower/addons/libs/socket.lua`. You can remove it and things should work the same as they do now.

addons/irc/irc.lua
```diff
@@ -0,0 +1,195 @@
...
+
+	if serveraddr and serverport then
+		-- eventually do something here
+	else
+		log("setting default twitch server and port")
+		serveraddr = 'irc.twitch.tv'
+		serverport = '6667'
+	end
+	serverport = tonumber(serverport)
+	server = socket.tcp()
```
z16, line 46: Changing this line to the following will allow you to use `server:send` without having to add `"\r\n"` at the end of every line:

``` lua
    server = function()
        local s = socket.tcp()
        return setmetatable({
            send = function(t, str)
                s:send(str .. '\r\n')
            end
        }, {__index = s})
    end()
```

Untested, but I think you get the idea, create a proxy for the `server` object returned by `socket.tcp`, so you can modify the send function accordingly. You could even do the following to incorporate string formatting:

``` lua
    server = function()
        local s = socket.tcp()
        return setmetatable({
            send = function(t, fmt, ...)
                s:send((fmt .. '\r\n'):format(...))
            end
        }, {__index = s})
    end()
```

Now you could call it like this:

``` lua
    server:send('JOIN %s', settings.login.channel)
```

addons/irc/irc.lua
```diff
@@ -0,0 +1,195 @@
...
+			end
+			nick,msg = string.match(data, '^:([^!]+).+ PRIVMSG #.+ :(.+)')
+			if nick and msg then
+				windower.add_to_chat(settings.chat.color,"<"..nick.."> "..msg)
+			end
+		end
+	end
+end)
+
+windower.register_event('addon command', function (...)
```
z16, line 133: You may consider something like _FFOChat_ that doesn't use commands but parses outgoing text. It's a bit more work but allows for shorter chat syntax when using multiple channels. I know you said it doesn't support that yet, but if it does in the future (and I'd really love it if it does, so we can deprecate FFOChat), this way would be rather cumbersome to chat in different channels.

z16, line 133: You can also change the signature to this:

``` lua
```

That way the first argument is the actual first argument for the specified command. Would require changing a bit of your code, but imo it would make it clearer what the intentions are.

z16:
The pull request is missing an entry in the `addons.xml` file, which is mandatory for the pull request to be accepted. It's also missing the licensing disclaimer, which is mandatory as well. You can find the template for that [here](http://dev.windower.net/doku.php?id=lua:writing_addons).

Skyrant:
Thanks for all the suggestions, it's my first AddOn and i never worked with lua/windower so i am not so familiar with the standards, tricks and possibilities. I come from a C/C++ world so learning it seems trivial. This was literally done in a day so it is still very rough around the edges. I pulled it this early to get some feedback and see if you would even consider this to be added.

The idea is to have different channels you could switch to with /cm <channelname> just like you would change channels in the game. I don't think i want to support multiple servers because that would require a lot of connection management and make things really complicated for the user input.

I will work on your the suggestions.
Thank you

z16:
I agree on the multiple server part, it would not be easy to add, although it would be convenient for some people.

The advantage of the _FFOChat_ plugin is that you don't have to explicitly switch between channels but can just switch by adding a different channel number. Ideally I'd want something _irssi_-like, where channels are assigned numbers and you can press `Alt+1` with the chat input open and it would prepend `/1` to the current chat line (similar to how the in-game mechanism between switching chat channels works, `Ctrl+L`, `Ctrl+P`, etc.). That is not possible with the current Lua API, but it can be exposed at some point.

And personally I'm very interested in this, we'll be wanting to convert all plugins to addons at some point, but _FFOChat_ was not possible to convert because of the lack of support for _LuaSocket_ until it was added (although even that's almost one year ago now...). So you can count on my support for this. If you come on our IRC channel, I'm sure we can talk more about this than here, but I'd love for this to go live soon to completely replace _FFOChat_ with, as long as it incorporates multiple channels.

Skyrant:
Updated Dynamishelper to make it work again. Been tested in all Dynamis zones. It needs optimized table management for proc mobs, right now there is a lot of redundant information.

I will think about a rewrite for the mob tables some other time.

alphaONE2:
We won't be able to merge this pull request like this, you should split the updates to DynamisHelper into a separate branch and submit a new pull request for it so it can be reviewed and processed separately.

Skyrant:
i don't know how to do that with the github web frontend. Any tipps?

alphaONE2:
You would need to use a git client, but something like this should work.

```
git branch dynamishelper-update
git revert 09fdbb9..c4ff073
git checkout dynamishelper-update
git revert e0e0bfb..09fdbb9
```

Then, publish the new branch and submit a pull request. You'll probably want to make a new branch for each addon you work on in the future.

alphaONE2:
I'm going to close this pull request, as it's been open for nearly 6 months without any resolution.

@Skyrant: If you would still like to submit this Addon (which I'm definitely in support of, and encourage you to do so!), please open another pull request, without the dynamis stuff mixed in.

PR #1108, opened 2014-11-11
Config library

trv6:
Missed one file.set() call.


z16:
I adjusted it a bit differently to make what's going on a bit more transparent, so I'll close this request. Thanks though for bringing this to my attention.

PR #1113, opened 2014-11-14
Menu, file based, etc.

trv6:
At least 70% sure I broke github.


trv6:
I'm not sure how to remove the 'Config library' pull request from the history (it was closed), but it's not a part of this commit.

addons/Rhombus/Rhombus.lua
```diff
@@ -0,0 +1,832 @@
+--[[Copyright � 2014, trv
```
z16: The copyright symbol here is screwed up. Make sure you use UTF-8 for all your files. It seems fine in the other file.

z16:
The config commit is not a problem.

PR #1115, opened 2014-11-18
[GearSwap] enable/disable command always output error message

nekonok:
`gs enable main` output this.

```
GearSwap: enable error, passed an unrecognized slot name (0).
```

disable() and enable() are called with a couple of values which returned from string:gsub() that are replaced string and found index.
So second argument is almost always 0, they never find the slot named "0".

### DIFF
--- a/addons/GearSwap/gearswap.lua
+++ b/addons/GearSwap/gearswap.lua
@@ -215,14 +215,16 @@ windower.register_event('addon command',function (...)
 end)
 
 function disenable(tab,funct,functname,pol)
+    local slot_name = ''
     if tab[2] and tab[2]:lower()=='all' then
         funct('main','sub','range','ammo','head','neck','lear','rear','body','hands','lring','rring','back','waist','legs','feet')
         print('GearSwap: All slots '..functname..'d.')
     elseif tab[2]  then
         for i=2,#tab do
-            if slot_map[tab[i]:gsub('[^%a_%d]',''):lower()] then
-                funct(tab[i]:gsub('[^%a_%d]',''):lower())
-                print('GearSwap: '..tab[i]:gsub('[^%a_%d]',''):lower()..' slot '..functname..'d.')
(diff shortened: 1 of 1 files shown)


z16:
There is a simpler fix available. Change this line:

``` lua
                funct(tab[i]:gsub('[^%a_%d]',''):lower())
```

To this:

``` lua
                funct((tab[i]:gsub('[^%a_%d]','')):lower())
```

This will only return the first argument of the `string.gsub` function. In general, if you have a number of values `...`, putting them in parentheses will restrict it to only the first argument: `(...)`. You don't need to assign it to a variable, nor do you need do explicitly declare an empty variable name `_` for it. This works just fine:

``` lua
local x = 1, 2, 3
```

Now `x` will only contain `1`.

addons/GearSwap/gearswap.lua
```diff
@@ -215,14 +215,16 @@ windower.register_event('addon command',function (...)
...
+    local slot_name = ''
     if tab[2] and tab[2]:lower()=='all' then
         funct('main','sub','range','ammo','head','neck','lear','rear','body','hands','lring','rring','back','waist','legs','feet')
         print('GearSwap: All slots '..functname..'d.')
     elseif tab[2]  then
         for i=2,#tab do
-            if slot_map[tab[i]:gsub('[^%a_%d]',''):lower()] then
-                funct(tab[i]:gsub('[^%a_%d]',''):lower())
-                print('GearSwap: '..tab[i]:gsub('[^%a_%d]',''):lower()..' slot '..functname..'d.')
+            slot_name, _ = tab[i]:gsub('[^%a_%d]',''):lower()
```
z16: Since you didn't declare `_` local, you introduce a new global variable called `_` at this point.

nekonok:
Thank you for your advise!
Removed unnecessary assignment to `_`.

PR #1133, opened 2014-12-09
obiaway update

cantreadcode:
- added new commands: //obiaway [help | notify | get]
- obiaway now pulls obi from other inventories as well as puts them away according to weather, day, and SCH buffs.
- stores all obi when entering the following areas (to prevent gearcollector users from having them misplaced):
  
  Ru'Lude Gardens, Upper Jeuno, Lower Jeuno, Port Jeuno, Port Windurst, Windurst Waters, Windurst Woods, Windurst Walls, Heavens Tower, Port San d'Oria, Northern San d'Oria, Southern San d'Oria, Port Bastok, Bastok Markets, Bastok Mines, Metalworks, Aht Urhgan Whitegate, Tavanazian Safehold, Nashmau, Selbina, Mhaura, Norg, Kazham, Eastern Adoulin, Western Adoulin, Leafallia, Celennia Memorial Library, Mog Garden

### DIFF
--- a/addons/obiaway/Obiaway.lua
+++ b/addons/obiaway/Obiaway.lua
@@ -69,13 +69,13 @@ settings = config.load(default_settings)
 
 -- Accepts msg as a string or a table
 function obi_output(msg)
-        windower.add_to_chat(209, msg)
+	windower.add_to_chat(209, msg)
 end
 
 windower.register_event('addon command', function()
 
 	return function(command, ...)
-	    command = command:lower() or 'help'
+		command = command:lower() or 'help'
         local params = {...}
 
 		if command == 'help' or command == 'h' then
@@ -112,7 +112,7 @@ windower.register_event('addon command', function()
 end())
 
 function get_obis_in_inventory()
-    obis = {}
+	obis = {}
     items = windower.ffxi.get_items()
     inv = items.inventory
     if not inv then return end
--- a/addons/obiaway/Obiaway.lua
+++ b/addons/obiaway/Obiaway.lua
@@ -50,6 +50,10 @@
 -- is called before inventory has loaded and returns nothing.
 --
 -- 3. Obi is not moved when currently equipped.
+--
+-- 4. When using the get all or put all command, obiaway will still
+-- autosort obi if you're not in town. On to-do.
+--
 
 _addon.name = "Obiaway"
 _addon.author = "ReaperX, onetime"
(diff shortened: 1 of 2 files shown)


addons/obiaway/Obiaway.lua
```diff
@@ -1,5 +1,5 @@
 -- 
--- Obiaway v0.1
+-- Obiaway v1.0.2
 -- 
 -- Copyright (c) 2013, ReaperX
```
alphaONE2: Please add yourself to the copyright notice. Your changes legally belong to you, not to ReaperX. Also, while you're doing that, please change the (c) to ©.

addons/obiaway/Obiaway.lua
```diff
@@ -38,43 +37,81 @@
...
--- is called before inventory has loaded and returns nothing. 
+-- is called before inventory has loaded and returns nothing.
 --
 -- 3. Obi is not moved when currently equipped.
 -- 
--- To Do: add function to turn the console_echo's on/off. 
+-- To Do: Use a loop to remove and retrieve obi for shorter code
 
+_addon.name = "Obiaway"
 _addon.author = "ReaperX"
```
alphaONE2: You may also want to add yourself to the author list.

alphaONE2:
I'll leave the rest of the code review to @z16.

addons/obiaway/Obiaway.lua
```diff
@@ -38,43 +37,81 @@
...
-settings = config.load(defaults)
+windower.register_event('addon command', function()
 
-windower.register_event('addon command',function (...)
-    if args[1] == 'location' then
-        if S{'sack','case','satchel'}:contains(args[2]:lower()) then
-            settings.location = args[2]
+    return function(command, ...)
+        command = command:lower() or 'help'
+        local params = {...}
```
z16: Since you use `:lower()` on all values in here (multiple times even) you can speed it up a bit and make it more compact by changing this line to the following:

``` lua
        local params = L{...}:map(string.lower)
```

Then you can drop the `:lower()` on all the uses of `params[i]` later.

addons/obiaway/Obiaway.lua
```diff
@@ -38,43 +37,81 @@
...
+            obi_output("//(obi)away [options]")
+            obi_output("   (h)elp :  Displays this help text.")
+            obi_output("   (g)et :  Command for manually getting and removing obis.")
+            obi_output("   (n)otify [on|off] :  Sets obiaway notifcations on or off.")
+            obi_output("   (l)ocation [sack|satchel|case|wardrobe] :  Sets inventory from which to get and put obis.")
+        elseif command == 'get' or command == 'g' then
+            get_needed_obis()
+            obi_output("Sorting obis...")
+        elseif command == 'notify' or command == 'n' then
+            if S{'on'}:contains(params[1]:lower()) then
```
z16: Can change this to:

``` lua
            if params[1] == 'on' then
```

Same thing below for `'off'`.

addons/obiaway/Obiaway.lua
```diff
@@ -38,43 +37,81 @@
...
+            end
         else
-            print('Location can only be sack, satchel or case.')
+            error("Unrecognized command. See //obiaway help.")
         end
-    else
-        remove_unneeded_obis()
     end
-end)
+end())
```
z16: You defined the function as a closure, but you're not closing over any variables. This isn't technically wrong to do, just unnecessary. Basically these two are equivalent:

``` lua
function() return function(...) --[[ stuff ]] end end() -- Closure over an anonymous function
                                                        -- with no closed over variables
function(...) --[[ stuff ]] end -- Regular anonymous function
```

That means you can change it to the following:

``` lua
    -- Inner function stuff here
end)
```

addons/obiaway/Obiaway.lua
```diff
@@ -98,36 +135,192 @@ function get_obis_in_inventory()
     return obis
 end
 
-function remove_unneeded_obis()
+function get_needed_obis()
     elements = get_all_elements()
     obis = get_obis_in_inventory()
+    
+    areas = {}
+    areas.Cities = S{
```
z16: This set is now created every time the function `get_needed_obis` is called. This won't really be noticeable, but it's unnecessarily wasteful. There are two options to address this issue:
1. Put the `areas` and `arias.Cities` variables in the global space. To do that just define `areas` and `areas.Cities` outside of the function.
2. Use a closure to close over these variables. This is what I mentioned above. The basic idea is this:

``` lua
some_function = function()
    local function_local_variable = 0
    return function()
        --[[ some code that uses function_local_variable ]]
    end
end()
```

What happens is that `function_local_variables` is local only to the inner function, since it's not in the global scope. But at the same time, the inner function doesn't define it, so it's only defined once. This is a useful way to define values local to one function that are not needed for any other function in the code without polluting the global namespace (or the global table in Lua's case). So in this case, it would look like the following:

``` lua
get_needed_obis = function()
    local areas = {
        Cities = S{
            'Foo',
            'Bar',
            'Baz',
        }
    }
    return function()
        elements = get_all_elements()
        obis = get_obis_in_inventory()

        local str = ''
        --[[ Rest of the function ]]
    end
end()
```

addons/obiaway/Obiaway.lua
```diff
@@ -98,36 +135,192 @@ function get_obis_in_inventory()
...
+            if settings.notify == 'on' then
+                obi_output("Putting Korin Obi away in "..settings.location..".")
+            end
+        end
+        if obis["Dark"] then    
+            str = str.."put \"Anrin Obi\" "..settings.location..";wait .5;"
+            if settings.notify == 'on' then
+                obi_output("Putting Anrin Obi away in "..settings.location..".")
+            end
+        end
```
z16, line 206: The preceding 8 if clauses could be condensed into one block if you define an element to name mapping, i.e. this:

``` lua
obi_names = T{
    Light = 'Korin',
    Dark = 'Anrin',
    --[[ And so on... ]]
}
```

Then you can replace the previous 48 lines with this:

``` lua
        for name, element in obi_names:it() do
            if obis[element] then    
                str = str..'put "%s" %s;wait .5;':format(name, settings.location)
                if settings.notify == 'on' then
                    obi_output('Putting %s away into %s.':format(name, settings.location))
                end
            end
        end
```

I didn't test it, but it should be something like this.

addons/obiaway/Obiaway.lua
```diff
@@ -38,43 +37,81 @@
...
 
 require('sets')
+require('logger')
 config = require('config')
 res = require('resources')
 
-defaults = {}
-defaults.location = 'sack'
+default_settings = {}
+default_settings.notify = 'on'
```
z16: This should be a boolean, so you can just do `if settings.notify then`. Makes code a bit clearer and simpler to write.

z16:
The code itself checks out (as far as I can tell), the notes I left were just some suggestions at code improvements, not errors on your part. Otherwise it looks ok and can be pulled once the copyright is adjusted.

alphaONE2:
Oh good, @cantreadcode, you updated. Just FYI, we aren't notified when commits are appended to a pull request, so make sure you comment if you want someone to review again. Did you decide to leave the closure, or would you like us to wait before merging this?

addons/obiaway/Obiaway.lua
```diff
@@ -27,109 +27,136 @@
...
 end
 
-function remove_unneeded_obis()
-    elements = get_all_elements()
-    obis = get_obis_in_inventory()
+-- Pause function. Accepts a number of seconds to pause.
+function wait(seconds)
+  local start = os.time()
+  repeat until os.time() > start + seconds
+end
```
z16, line 112: You should remove this function and use `coroutine.sleep(time)` instead. It will not perform a busy wait like this code does (i.e. max your CPU usage).

addons/obiaway/Obiaway.lua
```diff
@@ -27,109 +27,136 @@
...
-            settings.location = args[2]
-        else
-            print('Location can only be sack, satchel or case.')
-        end
-    else
-        remove_unneeded_obis()
-    end
-end)
+-- lists
+obi_cities = S{
```
z16: If the point of this is to know whether or not certain inventories are available, you can use `windower.ffxi.get_bag_info`. If not, nevermind this comment.

addons/obiaway/Obiaway.lua
```diff
@@ -137,9 +164,146 @@ function inTable(tbl, item)
     return false
 end
 
-function get_all_elements()
+-- converts name of a bag (str) to an id number. returns 0 (inventory) when no argument passed.
```
z16, line 127: Something like this can be done with this line:

``` lua
local inv_ids = res.bags:rekey('english'):key_map(string.lower):map(table.get-{'id'})
```

Then you can use it like this:

``` lua
print(inv_ids.storage)  -- Will output 2
```

One main advantages to this, a) it doesn't perform an expensive function call with many comparisons and b) if the IDs ever change it will automatically get them correct, since it reads them from the resources.

addons/obiaway/Obiaway.lua
```diff
@@ -137,9 +164,146 @@ function inTable(tbl, item)
...
+--              "Lightning" = true
+--              "Water" = false
+--              "Light" = false
+--              "Dark" = true
+--              n = 3
+--          }
+--
+-- function designer: ReaperX
+function get_obi_in_inventory(location)
+    id = inv_str_to_id(location)
```
z16: Is this meant to be `local`?

addons/obiaway/Obiaway.lua
```diff
@@ -150,36 +314,232 @@ function get_all_elements()
...
     elseif inTable(buffs, 185) then
-      elements["Dark"] = elements["Dark"] +1
+      elements["Dark"] = elements["Dark"] + 1
     end
     return elements
 end
 
-windower.register_event('lose buff', remove_unneeded_obis:cond(function(id) return id >= 178 and id <= 185 end))
+----------------------------------------
+---      SORTING functions
```
z16, line 310: Should consider making these return `true` and `false` instead of `1` and `0`.

addons/obiaway/Obiaway.lua
```diff
@@ -150,36 +314,232 @@ function get_all_elements()
...
+
+----------------------------------------
+---            EVENTS
+----------------------------------------
 
+windower.register_event('gain buff', auto_sort_obi:cond(function(id) return id >= 178 and id <= 185 end))
+windower.register_event('lose buff', auto_sort_obi:cond(function(id) return id >= 178 and id <= 185 end))
+windower.register_event('day change', 'weather change', auto_sort_obi)
+windower.register_event('addon command', function(command, ...)
+    local command = command:lower() or 'help'
```
z16: The `local` is redundant, since it's already function-local (all arguments are). Also, this will error if someone tries to call it without any arguments. You want this:

``` lua
```

addons/obiaway/Obiaway.lua
```diff
@@ -137,9 +164,146 @@ function inTable(tbl, item)
...
+    elseif str == 'wardrobe' then
+        return 8
+    end
+end
+
+-- Function which counts how many slots remain in a bag. if no location is passed, checks inventory.
+function free_space(location)
+    local id = inv_str_to_id(location)
+
+    local inv = windower.ffxi.get_items(id)
```
z16: While this works, you should use `windower.ffxi.get_bag_info` here, as it is much more efficient. `windower.ffxi.get_items` will push all items in that particular bag onto the Lua stack and pack them into tables with all their fields (ID, count, slot, extdata, etc.). `windower.ffxi.get_bag_info` is precisely there to prevent these expensive calls if all you want is the bag meta-data and not the contents.

z16:
Sorry that this comes so late, but like Iryoku said, we do not get notified if you make changes to an existing pull request, so we didn't notice it until now. When you get a chance to review my comments, you can leave a short note here or close and reopen the pull request, that way we will get notified.

addons/obiaway/Obiaway.lua
```diff
@@ -27,119 +27,245 @@
...
-            obis["Lightning"] = obis["Lightning"] or (id == 15439)
-            obis["Water"] = obis["Water"] or (id == 15440)
-            obis["Light"] = obis["Light"] or (id == 15441)
-            obis["Dark"] = obis["Dark"] or (id == 15442)
+-- Function which counts how many slots remain in a bag. if no location is passed, checks inventory.
+function free_space(location)
+    local id = inv_str_to_id(location)
+
+    local inv = windower.ffxi.get_bag_info(id)
+    n = inv.max - inv.count
```
z16, line 157: This should probably be `local`.

addons/obiaway/Obiaway.lua
```diff
@@ -150,36 +276,234 @@ function get_all_elements()
...
-      elements["Fire"] = elements["Fire"] +1
+      elements["Fire"] =  elements["Fire"] + 1
     elseif inTable(buffs, 183) then
-      elements["Water"] = elements["Water"] +1
+      elements["Water"] = elements["Water"] + 1
     elseif inTable(buffs, 181) then
-      elements["Earth"] = elements["Earth"] +1
+      elements["Earth"] = elements["Earth"] + 1
     elseif inTable(buffs, 180) then
       elements["Wind"] = elements["Wind"] +1
```
z16, line 295: Not that it matters, but this line is missing a space before the `1` :D

PR #1140, opened 2014-12-13
Update gearswap/user_functions.lua

smd111:
added ability to check state of debug_mode and show_swaps


Byrth:
Sorry, I give an explanation here:
https://github.com/smd111/Lua/commit/1ae9e75b21b7febcb14aae5eb247241ae2284d15

Were you trying to get something like "//gs debugmode check" to work?

PR #1144, opened 2014-12-16
setbgm - new addon

svanheulen:
I figured out how the Orchestrion sets the background music in your Mog House and this lets you set it to any music in the game.

### DIFF
--- a/addons/setbgm/setbgm.lua
+++ b/addons/setbgm/setbgm.lua
@@ -1,5 +1,5 @@
 --[[
-Copyright (c) 2014, Seth VanHeulen
+Copyright © 2014, Seth VanHeulen
 All rights reserved.
 
 Redistribution and use in source and binary forms, with or without
@@ -62,6 +62,7 @@ bgm = {
 }
 
 function setbgm_command(...)
+    local arg = {...}
     local info = windower.ffxi.get_info()
     if not info.mog_house or info.zone == 280 then
         windower.add_to_chat(207, 'Background music can only be set inside you Mog House.')
--- a/addons/setbgm/setbgm.lua
+++ b/addons/setbgm/setbgm.lua
@@ -31,7 +31,7 @@ SOFTWARE, EVEN IF ADVISED OF THE POSSIBILITY OF SUCH DAMAGE.
 --]]
 
 _addon.name = 'setbgm'
-_addon.version = '1.0.0'
+_addon.version = '1.0.1'
 _addon.command = 'setbgm'
 _addon.author = 'Seth VanHeulen (Acacia@Odin)'
 
@@ -62,7 +62,8 @@ bgm = {
 }
 
 function setbgm_command(...)
-    if not windower.ffxi.get_info().mog_house or windower.ffxi.get_info().zone == 280 then
+    local info = windower.ffxi.get_info()
+    if not info.mog_house or info.zone == 280 then
         windower.add_to_chat(207, 'Background music can only be set inside you Mog House.')
     elseif #arg == 1 and arg[1]:lower() == 'list' then
         windower.add_to_chat(207, 'Available music:')
--- a/addons/setbgm/setbgm.lua
+++ b/addons/setbgm/setbgm.lua
@@ -1,0 +1,97 @@
+--[[
+Copyright (c) 2014, Seth VanHeulen
+All rights reserved.
+
+Redistribution and use in source and binary forms, with or without
+modification, are permitted provided that the following conditions are
+met:
+
+1. Redistributions of source code must retain the above copyright
+notice, this list of conditions and the following disclaimer.
+
+2. Redistributions in binary form must reproduce the above copyright
+notice, this list of conditions and the following disclaimer in the
+documentation and/or other materials provided with the distribution.
+
+3. Neither the name of the copyright holder nor the names of its
+contributors may be used to endorse or promote products derived from
+this software without specific prior written permission.
+
+THIS SOFTWARE IS PROVIDED BY THE COPYRIGHT HOLDERS AND CONTRIBUTORS "AS
+IS" AND ANY EXPRESS OR IMPLIED WARRANTIES, INCLUDING, BUT NOT LIMITED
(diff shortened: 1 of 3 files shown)


addons/setbgm/setbgm.lua
```diff
@@ -0,0 +1,97 @@
+--[[
+Copyright (c) 2014, Seth VanHeulen
```
z16: Please change this to `©`.

addons/setbgm/setbgm.lua
```diff
@@ -0,0 +1,97 @@
...
+    [230]='A New Horizon - Tavnazian Archipelago', [231]='Celestial Thunder', [232]='The Ruler of the Skies', [233]="The Celestial Capital - Al'Taieu", [234]='Happily Ever After', [235]='First Ode: Nocturne of the Gods', [236]='Fourth Ode: Clouded Dawn', [237]='Third Ode: Memoria de la Stona', [238]='A New Morning', [239]='Jeuno -Starlight Celebration-',
+    [240]='Second Ode: Distant Promises', [241]='Fifth Ode: A Time for Prayer', [242]='Unity', [243]="Grav'iton", [244]='Revenant Maiden', [245]='The Forgotten City - Tavnazian Safehold', [246]='March of the Allied Forces', [247]='Roar of the Battle Drums', [248]='Young Griffons in Flight', [249]='Run Maggot, Run!',
+    [250]='Under a Clouded Moon', [251]='Autumn Footfalls', [252]='Flowers on the Battlefield', [253]='Echoes of a Zypher', [254]='Griffons Never Die',
+    [900]='Distant Worlds'
+}
+
+function setbgm_command(...)
+    if not windower.ffxi.get_info().mog_house or windower.ffxi.get_info().zone == 280 then
```
z16: `windower.ffxi.get_*` calls are usually pretty expensive. You can cache the result and re-use it later in the function so you don't have to call it twice:

``` lua
    local info = windower.ffxi.get_int()
    if not info.mog_house or info.zone == 280 then
```

addons/setbgm/setbgm.lua
```diff
@@ -0,0 +1,97 @@
...
+        windower.add_to_chat(207, 'Background music can only be set inside you Mog House.')
```
z16, line 68: Are you sure? Sometimes SE forgets those security checks, I wouldn't be surprised if this worked anywhere.

In fact, I know for sure you can manipulate music to a certain degree by adjusting incoming packet 0x00A, so this might be possible.

addons/setbgm/setbgm.lua
```diff
@@ -0,0 +1,97 @@
...
+    elseif #arg == 1 and arg[1]:lower() == 'list' then
```
z16, line 69: `arg` is not defined anywhere. I think you missed a `local arg = {...}` at the top of this function.

addons/setbgm/setbgm.lua
```diff
@@ -0,0 +1,97 @@
...
+            end
+        else
+            windower.add_to_chat(207, 'Command usage: setbgm list|<music id>')
+        end
+    else
+        windower.add_to_chat(207, 'Command usage: setbgm list|<music id>')
+    end
+end
+
+windower.register_event('addon command', setbgm_command)
```
z16, line 99: Since the function isn't used anywhere else, you could insert it in here directly, so you can avoid naming a function that you never use again. It also hides the function from global scope.

addons/setbgm/setbgm.lua
svanheulen, line 68: It's not possible with the 0X5F packet, I tried it.

addons/setbgm/setbgm.lua
svanheulen: I've used '(c)' in my other addons without issue. I prefer my source code to be entirely ASCII.

alphaONE2: (c) is legally meaningless, and while the word "Copyright" suffices, we don't live in the 1980's so © please.

addons/setbgm/setbgm.lua
z16, line 68: Ok, nevermind that then. If you're curious you can try playing with 0x00A for a future version of this addon, I'm not sure of the extent of BGM you can choose from, but I know for a fact _that_ you can set the BGM to a certain degree.

addons/setbgm/setbgm.lua
svanheulen, line 69: The "arg" variable is automatically created in Lua when you have a function with '...' as the parameters. Although you can define it, there's no need.

addons/setbgm/setbgm.lua
z16: We didn't pay it much attention to this in the past, so it's possible we overlooked it, I'm already adjusting it for every addon I commit some changes to.

addons/setbgm/setbgm.lua
alphaONE2, line 69: The arg variable was deprecated in Lua 5.1, and removed in 5.2 as well as LuaJIT. Future versions of Windower will be using LuaJIT, and so this will not work.

svanheulen, line 69: Good to know, I should probably update my other addons too then.

addons/setbgm/setbgm.lua
svanheulen, line 99: I find it more readable this way, but I can change it if you'd like.

z16, line 99: Don't need to, was just a suggestion, since I find the opposite more readable. Then you immediately know that the function is called in the `addon command` event, otherwise you see a function and then have to go check where it's used. But that is just my preference, you can keep it this way, just thought maybe you didn't know it worked this way.

z16:
As a note for further updates, you may wanna use the logger and packets library instead of manually outputting and composing packets, since that's more stable and handles certain error cases better.

Byrth:
You may find this link helpful when filling in your remaining song names:
http://www.ffxiah.com/forum/topic/32347/vanatunes-play-list

PR #1150, opened 2014-12-24
Updated Help List

MilkMansKid:
When treasury was updated none of the help section was expanded.

### DIFF
--- a/addons/Treasury/Treasury.lua
+++ b/addons/Treasury/Treasury.lua
@@ -1,6 +1,6 @@
 _addon.name = 'Treasury'
 _addon.author = 'Ihina'
-_addon.version = '1.2.0.0'
+_addon.version = '1.2.0.1'
 _addon.commands = {'treasury', 'tr'}
 
 res = require('resources')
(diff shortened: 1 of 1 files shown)


addons/Treasury/Treasury.lua
```diff
@@ -309,12 +309,16 @@ windower.register_event('addon command', function(command1, command2, ...)
...
     elseif command1 == 'help' then
         print('%s v%s':format(_addon.name, _addon.version))
-        print('    \\cs(255,255,255)lot|pass add|remove <name>\\cr - Adds are removes all items matching <name> to the specified list')
-        print('    \\cs(255,255,255)lot|pass clear\\cr - Clears the specified list for the current character')
+        print('    \\cs(255,255,255)lot|pass|drop add|remove <name>\\cr - Adds or removes all items matching <name> to the specified list')
+        print('    \\cs(255,255,255)lot|pass|drop clear\\cr - Clears the specified list for the current character')
         print('    \\cs(255,255,255)lot|pass list\\cr - Lists all items on the specified list for the current character')
         print('    \\cs(255,255,255)lotall|passall\\cr - Lots/Passes all items currently in the pool')
-        print('    \\cs(255,255,255)clearall\\cr - Removes lotting/passing settings for this character')
+        print('    \\cs(255,255,255)clearall\\cr - Removes lotting/passing (not dropping) settings for this character')
```
z16: I forgot to add this feature to `clearall`, but it should be there. Can you add that?

addons/Treasury/Treasury.lua
```diff
@@ -309,12 +309,16 @@ windower.register_event('addon command', function(command1, command2, ...)
...
         print('    \\cs(255,255,255)autodrop [on|off]\\cr - Enables/disables (or toggles) the auto-drop setting')
+        --print('    \\cs(255,255,255)verbose [on|off]\\cr - Enables/disables (or toggles) the verbose setting')
+        --print('    \\cs(255,255,255)autostack [on|off]\\cr - Enables/disables (or toggles) the autostack feature')
+        --print('    \\cs(255,255,255)delay #\\cr - Allows you to change the delay of actions Default:0')
```
z16: Why did you comment out these lines? Also, you should write the last line like this: `delay <value>` instead of `delay #`.

z16:
Thanks, looks like I forgot about that :X

addons/Treasury/Treasury.lua
MilkMansKid: Honestly I didn't know if you forgot or didn't want everyone to mess with them. Just like the special predefined lots/passes (seals, crystals etc) aren't known unless you read the code. But know that I know I'll fix the rest.

PR #1162, opened 2015-01-03
Nostrum -- Add indication that party member is out of range, Fields -- 0x00D mismatch

trv6:
My copy of fields is out of date. I'm unsure if only the changes I made will be merged, or if the entire old version will.


trv6:
Removed conflicts.

PR #1171, opened 2015-01-08
Update findAll.lua

mujihina:
small QoL change for those of us who use this often


z16:
I'll push a new version of _FindAll_ tomorrow that uses the `find` command.

Issue #1170, opened 2015-01-06
(bug)windower.text.get_extents


smd111:
windower.text.get_extents returns nil not what its supost to

this is what im using
mx, my = windower.text.get_extents(_G[check_windows[i][2]]._name)
i know that there is more that goes here to get the name

z16:
This works fine for me (and for many other addons, or otherwise dragging and dropping would not be working). Can you show the full code? Here is a short snippet, can you see if this works for you:

``` lua
windower.text.create('foo');
windower.text.set_text('foo', 'Testing');
windower.text.set_font('foo', 'Arial');
windower.text.set_font_size('foo', 14);
print(windower.text.get_extents('foo')); -- > 64    23
```

z16:
On a further note, nil is returned if the name is not found, so maybe the argument you're providing is invalid. Can you print it out and make sure it's correct?

Regardless, you would probably be better off using the `texts` library.

smd111:
this prints correctly
check_windows[i][2]
but when i do this i get nil
_G[check_windows[i][2]]._name
and i do use the text.lua but it does not give extents anyways

z16:
Why do you need the `_G` table then?

smd111:
i use _G to get the table so i can pass the tables name to get_extents with
windower.text.get_extents(_G[check_windows[i][2]].name)

this is my display code
https://github.com/smd111/Gearswap/blob/master/includes/extra_more/Display.lua
this worked till sunday (i did not log in yesterday)

i just tried this and no goo still gives nil
texts.extents(check_windows[i][2])
here is the out put
17:27:51 > button
17:27:51 > GearSwap: Lua runtime error: libs/texts.lua:384:
17:27:51 >   attempt to index field '?' (a nil value)

smd111:
ok i know what the issue is the texts.lua on the live version is bad i copied the dev version ovder and now it working correctly

z16:
Strange, but ok, that will be fixed shortly then.

smd111:
i can only surmise its because the live version have changed
t. to meta[t]. when building the table

z16:
But that was an entirely self-contained change, it shouldn't have changed the behavior of anything outside of `texts.lua`… if it did, it was a mistake. Either way, we're gonna push the changes live shortly, thanks for letting us know.

smd111:
i think it because the info get changed to a local table(or a meta table) here
texts.lua (live)line:171
`t = {}
    local m = {}
[...]

PR #1176, opened 2015-01-10
Merge pending Arcon's approval

Byrth:


smd111:
with texts.lua
did you check to make sure that
var = texts.new(settings)
returns the table because when live had the metatable built this way

```
t = {}
local m = {}
meta[t] = m
```

it would cause an error when you tried to check the name with
var.name and var._name

here is the info
https://github.com/Windower/Lua/issues/1170

z16:
That's not even supposed to work. That particular error should already be fixed on live.

Also, I approve!

smd111:
no its not fix today i had to use an old texts(from dev i have not updated it for a week or more i did not realize this fact) file to make it work but i just tried the latest dev texts and its no good

z16:
Post your issue then, I'll review it before merging this. Every add-on that uses texts works fine, so chances are that it's something very specific to a certain feature of the texts library. That, or you're using it incorrectly.

smd111:
never mind i figured out how to do it
i.e.
var = texts.new(settings)
texts.extents(var)

my brain is farting alot lately

z16:
This should also work:

``` lua
var = texts.new('test')
print(var:extents())
```

smd111:
yes i know its just my brain being stupid
please forgive me

PR #1180, opened 2015-01-12
Latent checker fixed args = l{...} error

smd111:
the error on line 78 referring to a "nil value of variable l" is fixes
i should have run it more before i pushed it


Byrth:
args is never referenced anyway, so I just deleted it instead. I didn't see this fix.

smd111:
i left it there just in-case i decided to add anything to it in the future that might need it

PR #1187, opened 2015-01-16
autolock addon initial release

cantreadcode:
autoLock addon automatically equips a gear set and then uses the lockstyle command when leaving town.
requires gearswap addon and a set called sets.lockstyle in your job.lua file.


addons/autoLock/autoLock.lua
```diff
@@ -0,0 +1,166 @@
...
+_addon.author = "bangerang"
+_addon.version = "1.0"
+_addon.commands = {'al', 'autolock'}
+_addon.language = 'english'
+
+require('luau')
+config = require('config')
+
+defaults = {}
+defaults.zoneDelay = 10
```
z16: This is not mandatory, but we either use `snake_case` or `PascalCase` for settings values, the latter being preferred. Up to you if you wanna change it, but `camelCase` goes against conventions and it makes me sad :'(

Same goes for the addon name.

addons/autoLock/autoLock.lua
```diff
@@ -0,0 +1,166 @@
...
+defaults.keyBind = '^x'
+defaults.autoLock = true
+defaults.ignoreZones = S{
```
z16: I assume you want this to contain all town zones? I believe there is a way to figure those out from the zoning packet (`0x00A`), but I don't know where exactly. Would require some research, just letting you know.

addons/autoLock/autoLock.lua
```diff
@@ -0,0 +1,166 @@
...
+windower.send_command('bind %s al':format(settings.keyBind))
+windower.send_command('input /lockstyle off')
+
+function al_output(msg)
+    prefix = 'autoLock: '
+    windower.add_to_chat(209, prefix..msg)
+end
+
+-- Accepts a boolean value and returns an appropriate string value. i.e. true -> 'on'
+function booltostr(bool)
```
z16: This function can be rewritten to this:

``` lua
booltostr = function(bool)
    return bool and 'on' or 'off'
end
```

addons/autoLock/autoLock.lua
```diff
@@ -0,0 +1,166 @@
...
+    local str = ''
+    if bool then
+        str = 'on'
+        return str
+    elseif not bool then
+        str = 'off'
+        return str
+    else
```
z16: This case can never happen. You can change the second change to just `else`.

addons/autoLock/autoLock.lua
```diff
@@ -0,0 +1,166 @@
...
+        print('Boolean to string: unknown error.')
+    end
+end
+
+function command_lockstyle()
+	if tokens.lockstyle then
```
z16: You've been using 4 spaces for indentation up until this point and here you're switching to tabs. Please switch the entire file to 4 space indentation.

addons/autoLock/autoLock.lua
```diff
@@ -0,0 +1,166 @@
...
+		windower.send_command('input /lockstyle off')
+		coroutine.sleep(1)
+	end
+	windower.send_command('gs equip sets.lockstyle;wait 1.5;input /lockstyle on')
```
z16: You could make this configurable by the user instead of forcing them to make a set called `sets.lockstyle`. Just add `defaults.LockSet = 'sets.lockstyle'` to the default settings then users can overwrite it if they want.

addons/autoLock/autoLock.lua
```diff
@@ -0,0 +1,166 @@
...
+
+settings = config.load(defaults)
+
+tokens = {}
+tokens.lockstyle = false
+
+windower.send_command('bind %s al':format(settings.keyBind))
+windower.send_command('input /lockstyle off')
+
+function al_output(msg)
```
z16: You may consider using the `logger` library and it's `log` function, which basically does the same thing as this. Only it uses color `207` by default, but users can adjust that in its settings file so it outputs to the color they like.

z16:
The addon checks out except for three things:
1. As I mentioned in a comment please switch from tabs to 4 spaces for indentation
2. You're missing an entry in the `addons.xml` file
3. There may be a copy/paste error in the license text. It says _ReaperX_ as the author later on, is that you? If so, you should adjust the author at the top from _bangarang_ to _ReaperX_ or vice versa so the names in the license disclaimer match.

All the other comments I left are just optional changes that you can do if you want, but don't have to. Once these three main issues are fixed I can merge this.

PR #1194, opened 2015-01-17
Setting Commands

MilkMansKid:
Default settings don't line up for me sure others having same issue.


z16:
Thanks for the command handler, but the indentation is all screwy. Can you adjust it to be 4 spaces instead of tabs and add proper block-indentation?

MilkMansKid:
Well thanks for pointing out my notepad++ settings got changed for my tab settings... No problem updating now.

addons/ConsoleBG/ConsoleBG.lua
```diff
@@ -28,19 +29,107 @@ config.register(settings, function(settings)
...
+
+        elseif command2 == 'y' then
+            settings.pos.y = math.floor(command3)
+            log('Position Y set to %s':format(settings.pos.y))
+            config.save(settings)
+            config.reload(settings)
+        end
+
+    elseif command1 == 'bg' then
+    local bgvalue = tonumber(command3)
```
z16: This should be indented :)

addons/ConsoleBG/ConsoleBG.lua
```diff
@@ -28,19 +29,107 @@ config.register(settings, function(settings)
     windower.prim.set_size('ConsoleBG', settings.extents.x, settings.extents.y)
 end)
 
+windower.register_event('addon command', function(command1, command2, command3)
```
z16: It's generally more useful to only specify the root command and leave the others variable (as `...`), then pack them into an `args` table or something. That way it's easier to amend in the future.

addons/ConsoleBG/ConsoleBG.lua
```diff
@@ -28,19 +29,107 @@ config.register(settings, function(settings)
...
+
+    command1 = command1 and command1:lower() or 'help'
+    command2 = command2 and command2:lower()
```
z16: This will be `false` if the user doesn't provide it. Won't matter in this context, but not sure if this is intended. Would also be fixed by making an `args` table instead of explicitly declaring all arguments, especially since `command2` and `command3` aren't accurate terms, as they are not always commands.

z16:
It's common for Windower commands to allow specifying all positions and colors in one command, i.e.:

```
consolebg color 204 0 0 102
```

instead of

```
consolebg color alpha 204
consolebg color red 0
consolebg color green 0
consolebg color blue 102
```

addons/ConsoleBG/ConsoleBG.lua
MilkMansKid: I knew this but I'm new to programming(only background is (x)html/css) >.< and as much as I can edit/hack decently at other addons already trying to write it from starch I;m still working on.  I could of easily looked at another addon and copy and modified the code to suit my needs but was trying to get  some practice in of starting from scratch and I know I don't have it memorized enough to do that.

addons/ConsoleBG/ConsoleBG.lua
MilkMansKid: That was not intended. >.<

MilkMansKid:
Thank you for pointing out my mistakes and giving me ideas.

PR #1202, opened 2015-01-21
Pointwatch -- gold/golden mismatch

trv6:
Changed instances of 'golden' to 'gold' to match the ReadMe.


trv6:
Gold -> Golden instead

addons/pointwatch/statics.lua
```diff
@@ -142,7 +142,7 @@ function initialize()
         setfenv(cur_func,_G)
         dynamis.entry_time = os.clock()
         dynamis.zone = info.zone
-        windower.add_to_chat(123,'Loading PointWatch in Dynamis results in an inaccurate timer. Number of KIs is displayed.')
+        error(123,'Loading PointWatch in Dynamis results in an inaccurate timer. Number of KIs is displayed.')
```
z16: The `123` needs to be removed for this to work right. Was this change intended? This doesn't sound like an error.

trv6:
@z16  Unintended. I may have edited the live file and replaced the dev file with it.

PR #1212, opened 2015-02-07
TargetClaim: initial commit

KenshiDRK:
Initial commit of my first addon TargetClaim. It display who has claim if your target is a mob or what is targetting your target if is a pc.

### DIFF
--- a/addons/TargetClaim/TargetClaim.lua
+++ b/addons/TargetClaim/TargetClaim.lua
@@ -1,0 +1,106 @@
+--[[Copyright © 2014, Kenshi
+All rights reserved.
+
+Redistribution and use in source and binary forms, with or without
+modification, are permitted provided that the following conditions are met:
+
+    * Redistributions of source code must retain the above copyright
+      notice, this list of conditions and the following disclaimer.
+    * Redistributions in binary form must reproduce the above copyright
+      notice, this list of conditions and the following disclaimer in the
+      documentation and/or other materials provided with the distribution.
+    * Neither the name of TargetClaim nor the
+      names of its contributors may be used to endorse or promote products
+      derived from this software without specific prior written permission.
+
+THIS SOFTWARE IS PROVIDED BY THE COPYRIGHT HOLDERS AND CONTRIBUTORS "AS IS" AND
+ANY EXPRESS OR IMPLIED WARRANTIES, INCLUDING, BUT NOT LIMITED TO, THE IMPLIED
+WARRANTIES OF MERCHANTABILITY AND FITNESS FOR A PARTICULAR PURPOSE ARE
+DISCLAIMED. IN NO EVENT SHALL Kenshi BE LIABLE FOR ANY
+DIRECT, INDIRECT, INCIDENTAL, SPECIAL, EXEMPLARY, OR CONSEQUENTIAL DAMAGES
+(INCLUDING, BUT NOT LIMITED TO, PROCUREMENT OF SUBSTITUTE GOODS OR SERVICES;
+LOSS OF USE, DATA, OR PROFITS; OR BUSINESS INTERRUPTION) HOWEVER CAUSED AND
+ON ANY THEORY OF LIABILITY, WHETHER IN CONTRACT, STRICT LIABILITY, OR TORT
+(INCLUDING NEGLIGENCE OR OTHERWISE) ARISING IN ANY WAY OUT OF THE USE OF THIS
+SOFTWARE, EVEN IF ADVISED OF THE POSSIBILITY OF SUCH DAMAGE.]]
+
+_addon.name = 'TargetClaim'
+_addon.author = 'Kenshi'
+_addon.version = '1.0'
+
+require('luau')
+texts = require('texts')
+
+defaults = {}
+defaults.showclaimname = true
+defaults.showclaimid = true
+defaults.display = {}
+defaults.display.pos = {}
+defaults.display.pos.x = 0
+defaults.display.pos.y = 0
+defaults.display.bg = {}
+defaults.display.bg.red = 0
+defaults.display.bg.green = 0
+defaults.display.bg.blue = 0
+defaults.display.bg.alpha = 102
+defaults.display.text = {}
+defaults.display.text.font = 'Consolas'
+defaults.display.text.red = 255
+defaults.display.text.green = 255
+defaults.display.text.blue = 255
+defaults.display.text.alpha = 255
+defaults.display.text.size = 10
+
+settings = config.load(defaults)
+settings:save()
+
+text_box = texts.new(settings.display, settings)
+
+initialize = function(text, settings)
+    local properties = L{}
+	if settings.showclaimname then
+        properties:append('${claim_name}')
+    end
+    if settings.showclaimid then
+        properties:append('${claim_id}')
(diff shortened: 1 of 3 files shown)


z16:
I'm not at home right now and don't have much time to look into it at the moment but this addon looks like it could be merged into TargetInfo rather than be its own addon, as it just displays some additional information about the current target. Or is there a particular reason why you decided to make it separate?

There are also some things you'll need to adjust before we can merge this, but I'll go over that tomorrow night when I'm home again.

KenshiDRK:
The only reason was that adding more information to targetinfo will get on top of my custom timers times, but I can adjust the position of this or change the font size of targetinfo.

And I want to add commands to be able to change the settings while in game.

addons/TargetClaim/TargetClaim.lua
```diff
@@ -0,0 +1,106 @@
+--[[Copyright © 2014, Kenshi
```
z16, line 1: 2014 is gone and never coming back :'(

addons/TargetClaim/TargetClaim.lua
```diff
@@ -0,0 +1,106 @@
...
+
+_addon.name = 'TargetClaim'
+_addon.author = 'Kenshi'
+_addon.version = '1.0'
+
+require('luau')
+texts = require('texts')
+
+defaults = {}
+defaults.showclaimname = true
```
z16, line 35: Please _PascalCase_ this.

addons/TargetClaim/TargetClaim.lua
```diff
@@ -0,0 +1,106 @@
...
+defaults.display.text.size = 10
+
+settings = config.load(defaults)
+settings:save()
+
+text_box = texts.new(settings.display, settings)
+
+initialize = function(text, settings)
+    local properties = L{}
+	if settings.showclaimname then
```
z16, line 61: Use 4-space indentation throughout the entire addon.

addons/TargetClaim/TargetClaim.lua
```diff
@@ -0,0 +1,106 @@
...
+        properties:append('${claim_id}')
+    end
+
+    text:clear()
+    text:append(properties:concat('\n'))
+end
+
+text_box:register_event('reload', initialize)
+
+initialize(text_box, settings)
```
z16, line 74: This line should not be needed because of the preceding line.

addons/TargetClaim/TargetClaim.lua
```diff
@@ -0,0 +1,106 @@
...
+			info.claim_name = mobclaim.name
+			info.claim_id = mob.claim_id
+			text_box:update(info)
+			text_box:show()
+		elseif mob.target_index > 0 then
+			local target = windower.ffxi.get_mob_by_index(mob.target_index)
+			local info = {}
+			info.claim_name = target.name
+			info.claim_id = target.id
+			text_box:update(info)
```
z16: Instead of using `text_box:update(info)`, where `info` is a table containing the named fields `claim_name` and `claim_id` you can just assign to those fields exactly:
```lua
text_box.claim_name = mobclaim.name
text_box.claim_id = mobclaim.id
```

addons/TargetClaim/TargetClaim.lua
```diff
@@ -0,0 +1,106 @@
...
+windower.register_event('prerender', function()
+	local mob = windower.ffxi.get_mob_by_target('st') or windower.ffxi.get_mob_by_target('t')
+	local player = windower.ffxi.get_player()
+	if mob then
+		if mob.claim_id > 0 then
+			local mobclaim = windower.ffxi.get_mob_by_id(mob.claim_id)
+			local info = {}
+			info.claim_name = mobclaim.name
+			info.claim_id = mob.claim_id
+			text_box:update(info)
```
z16, line 85: Instead of using `text_box:update(info)`, where `info` is a table containing the named fields `claim_name` and `claim_id` you can just assign to those fields exactly:

``` lua
text_box.claim_name = mobclaim.name
text_box.claim_id = mobclaim.id
```

addons/TargetClaim/TargetClaim.lua
```diff
@@ -0,0 +1,106 @@
...
+text_box:register_event('reload', initialize)
+
+initialize(text_box, settings)
+
+windower.register_event('prerender', function()
+	local mob = windower.ffxi.get_mob_by_target('st') or windower.ffxi.get_mob_by_target('t')
+	local player = windower.ffxi.get_player()
+	if mob then
+		if mob.claim_id > 0 then
+			local mobclaim = windower.ffxi.get_mob_by_id(mob.claim_id)
```
z16, line 81: This mob may not always exist (or at least the client may not always know about it yet). So you need to wrap the following lines into a check:

``` lua
if mobclaim then
    --[[ following lines here ]]
end
```

addons/TargetClaim/TargetClaim.lua
```diff
@@ -0,0 +1,106 @@
...
+			local info = {}
+			info.claim_name = mobclaim.name
+			info.claim_id = mob.claim_id
+			text_box:update(info)
+			text_box:show()
+		elseif mob.target_index > 0 then
+			local target = windower.ffxi.get_mob_by_index(mob.target_index)
```
z16, line 88: Same notes as before.

addons/TargetClaim/TargetClaim.lua
```diff
@@ -0,0 +1,106 @@
...
+			info.claim_id = target.id
+			text_box:update(info)
+			text_box:show()
+		elseif mob.name == player.name then
+			local info = {}
+			info.claim_name = player.name
+			info.claim_id = player.id
+			text_box:update(info)
+			text_box:show()
+		else
```
z16, line 100: All of the above three cases can be condensed if you rename the text fields from `claim_name` to `name` and from `claim_id` to `id`. Then you can condense the entire `prerender` event function into this:

``` lua
windower.register_event('prerender', function()
    local target = windower.ffxi.get_mob_by_target('st', 't')
    if not target then
        text_box:hide()
        return
    end

    local update = windower.ffxi.get_mob_by_id(mob.claim_id) or windower.ffxi.get_mob_by_index(target.target_index)
    if not update then
        local me = windower.ffxi.get_mob_by_target('me')
        update = me.name == target.name and me or nil
    end

    if update then
        text_box:update(update)
        text_box:show()
    else
        text_box:hide()
    end
end)
```

I haven't tested this code exactly, but the idea is sound. That way you're protected against various issues the code above has and it's much more concise and less repetitive.

KenshiDRK:
so do it as its own addon or merge it into the targetinfo addon? I added the code in the targetinfo addon but it get a few problems:
1. The formatting of the box get to big and messed up due to character names being a max of 15 chars, it get worst with npc that can be even more.
2. I'll like to hide the claim and claim id lines if there is none and the target and target id as well, Im not good still with the text libraries so I dunno how to do it. Though I was able to hide them but in a very weird way that should be avoided.

z16:
You're free to do either, it's up to you. I was just suggesting it because it seems to fit into what TargetInfo already does.

KenshiDRK:
Ok I'll revert those changes and push it to targetinfo. Closing this.

PR #1213, opened 2015-02-09
visiblefavor - new addon

svanheulen:


smd111:
man i have been looking for that indi- bubble for npc's for months
now all thats left is the one for pc's(not player)

also 

```
[0x10] = 0x76, -- Carbuncle
[0x11] = 0x77, -- Fenrir
[0x12] = 0x70, -- Ifrit
[0x13] = 0x73, -- Titan
[0x14] = 0x75, -- Leviathan
[0x15] = 0x72, -- Garuda
[0x16] = 0x71, -- Shiva
[0x17] = 0x74, -- Ramuh
[0x19] = 0x77, -- Diabolos
[0x1c] = 0x76, -- Cait Sith
```

should give you the max size that can be used

svanheulen:
Ah cool, I didn't know there was an even larger bubble size. Although I tested it an those seems to be a 12 yalms radius and the ones I was using are a 9 yalms radius. Since Favor and Wards have a 10 yalm radius, I would rather them be a little too small then too large.

smd111:
thats because its actualy a bit code
1-11-0-110
active-size-ally/enemy-element
or something like that i cant rember

alphaONE2:
I'm not positive, but I'm pretty sure we figured out it's 3' per value. so the bubble can go from 3'-12' in 3' increments. Don't quote me on that though. If I am remembering correctly though, 9' would be the closest, and is probably the best choice.

P.S. This is a _very_ cool idea.

svanheulen:
Very cool! Yeah I just tested it, there's 2 bits for size just as smd111 said.
size = size_bits \* 3 + 3
So you can make it 3/6/9/12 yalms but nothing between.

smd111:
@IryokuChevalier
i agree its a good idea

addons/libs/packets/fields.lua
```diff
@@ -1300,7 +1300,7 @@ fields.incoming[0x00E] = L{
     {ctype='unsigned char',     label='HP %',               fn=percent},        -- 1E
     {ctype='unsigned char',     label='Status',             fn=status},         -- 1F   Status used to be 0x20
     {ctype='unsigned int',      label='_unknown2',          fn=bin+{4}},        -- 20
-    {ctype='unsigned int',      label='_unknown3',          fn=bin+{4}},        -- 24
+    {ctype='unsigned int',      label='_unknown3',          fn=bin+{4}},        -- 24   Byte 0x26 is the Geomancy bubble displaying around the NPC
```
z16: Libs commits should be separate from addon commits.

addons/libs/packets/fields.lua
```diff
@@ -1801,6 +1801,22 @@ fields.incoming[0x036] = L{
 }
 
 enums.indi = {
```
z16, line 1803: This entire enum needs to be removed and replaced with appropriate bit fields. This is not your fault at all, just an observation. You may as well remove it from this commit, I'll try to adjust it tonight. Your addon doesn't seem to be using this data anyway.

addons/visiblefavor/visiblefavor.lua
```diff
@@ -0,0 +1,127 @@
...
+    [0x19] = 0x67, -- Diabolos
+    [0x1c] = 0x66, -- Cait Sith
+}
+
+refresh = true
+pet_index = 0
+favor_buff = false
+
+defaults = {}
+defaults.enabled = true
```
z16: Settings are conventionally _PascalCased_.

addons/visiblefavor/visiblefavor.lua
```diff
@@ -0,0 +1,127 @@
...
+                    effect = effect + 8
+                end
+                return modified:sub(1, 38) .. string.char(effect) .. modified:sub(40)
+            end
+        end
+    end
+end
+
+function visiblefavor_command(...)
+    local arg = {...}
```
z16: If you convert all arguments to lower-case here you don't need to `:lower()` them at every point afterwards.

addons/visiblefavor/visiblefavor.lua
```diff
@@ -0,0 +1,127 @@
...
+    elseif #arg == 2 and arg[1]:lower() == 'effect' and arg[2]:lower() == 'debuff' then
+        settings.effect_mode = 'debuff'
+    else
+        windower.add_to_chat(167, 'Command usage:')
+        windower.add_to_chat(167, '    vf toggle/on/off')
+        windower.add_to_chat(167, '    vf display all/self/favor')
+        windower.add_to_chat(167, '    vf effect buff/debuff')
+        return
+    end
+    windower.add_to_chat(207, 'visiblefavor: enable = %s, display = \31\200%s\30\1, effect = \31\200%s\30\1':format(settings.enabled and '\31\204yes\30\1' or '\31\167no\30\1', settings.display_mode, settings.effect_mode))
```
z16: Please _PascalCase_ the name here :(
Also, using hex literals is safer than decimal literals for various reasons.

addons/libs/packets/fields.lua
svanheulen: ah ok, removed

addons/visiblefavor/visiblefavor.lua
svanheulen: You mean in the string? Yeah that drives me nuts, but I thought that was the only way to represent non-printable characters in a string in Lua?

z16: You can use `\xXX` to indicate hex values:

``` lua
    windower.add_to_chat(207, 'VisibleFavor: enable = %s, display = \x1F\xC8%s\x1E\x01, effect = \x1F\xC8%s\x1E\x01':format(settings.enabled and '\x1F\xCCyes\x1E\x01' or '\x1F\xA7no\x1E\x01', settings.display_mode, settings.effect_mode))
```

svanheulen: Awesome! Thanks, I'll fix that then :)

addons/visiblefavor/visiblefavor.lua
```diff
@@ -0,0 +1,130 @@
...
+HOLDER OR CONTRIBUTORS BE LIABLE FOR ANY DIRECT, INDIRECT, INCIDENTAL,
+SPECIAL, EXEMPLARY, OR CONSEQUENTIAL DAMAGES (INCLUDING, BUT NOT LIMITED
+TO, PROCUREMENT OF SUBSTITUTE GOODS OR SERVICES; LOSS OF USE, DATA, OR
+PROFITS; OR BUSINESS INTERRUPTION) HOWEVER CAUSED AND ON ANY THEORY OF
+LIABILITY, WHETHER IN CONTRACT, STRICT LIABILITY, OR TORT (INCLUDING
+NEGLIGENCE OR OTHERWISE) ARISING IN ANY WAY OUT OF THE USE OF THIS
+SOFTWARE, EVEN IF ADVISED OF THE POSSIBILITY OF SUCH DAMAGE.
+--]]
+
+_addon.name = 'visiblefavor'
```
svanheulen: Should I rename the whole thing to "VisibleFavor"?

z16: Up to you, but I'd really appreciate it. I know not all addons follow that convention because, for a very random reason, it was initially decided against it, but it makes no sense to do that, as it's just harder to read and does nothing else.

addons/visiblefavor/visiblefavor.lua
```diff
@@ -0,0 +1,130 @@
...
+    elseif #arg == 2 and arg[1] == 'effect' and arg[2] == 'debuff' then
+        settings.effect = 'debuff'
+    else
+        windower.add_to_chat(167, 'Command usage:')
+        windower.add_to_chat(167, '    vf toggle/on/off')
+        windower.add_to_chat(167, '    vf display all/self/favor')
+        windower.add_to_chat(167, '    vf effect buff/debuff')
+        return
+    end
+    windower.add_to_chat(207, 'VisibleFavor: enable = %s, display = \031\200%s\030\001, effect = \031\200%s\030\001':format(settings.enabled and '\031\204yes\030\001' or '\031\167no\030\001', settings.display, settings.effect))
```
svanheulen: I tried the hex notation and it didn't work. I looked it up and it seems like that was something added in Lua 5.2, which I assume Windower is still using 5.1? Either way, I switched to the full "\ddd" notation on them instead of the shortened form. Not ideal but better then before.

z16: You're right, I forgot that we manually changed all our libs when we went back to 5.1. This works, or you can use the chat library and use it like this:

``` lua
    windower.add_to_chat(207, ('VisibleFavor: enable = %s, display = ' .. '%s':color(456) .. ', effect = ' .. '%s':color(456)):format(settings.enabled and 'yes':color(460) or 'no':color(423), settings.display, settings.effect))
```

I'm not sure if I got the numbers correctly, but I think I did. If not you can experiment with it a little and add/subtract 1 until you get it right.

addons/visiblefavor/visiblefavor.lua
svanheulen: All renamed... yay, for NTFS being case insensitive >.<

addons/visiblefavor/visiblefavor.lua
svanheulen: Switched to using the chat lib, thanks :)

Byrth:
IIRC there is some minor difference in that bitpacking based on the target of the indi spell (players vs. monsters). The particular packing should be in GearSwap somewhere, if I didn't put it into fields.lua.

PR #1214, opened 2015-02-11
AdherentBuffs

Sechs:
New tentative addon I'd like to share with some friends. Tested it on Windower Live.

### DIFF
--- a/addons/AdherentBuffs/AdherentBuffs.lua
+++ b/addons/AdherentBuffs/AdherentBuffs.lua
@@ -1,0 +1,70 @@
+-- Copyright © 2015, Mafai, Sechs
+-- All rights reserved.
+-- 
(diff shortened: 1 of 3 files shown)


smd111:
did you check to make sure that /ab chatmode say

```
elseif command == 'chatmode' then
 if chatmodes:contains(args[1]) then
```

works because you might need to add

```
require ('strings')
```

or

```
require ('tables')--most likely this
```

but if you only want one require you can use this

```
require ('luau')
```

that has
require('logger')
require('strings')
require('tables')
require('lists')
require('sets')
require('maths')
require('functions')
config = require('config')
res = require('resources')
all in one shot

because you only required config but did not require whats needed for contains

```
 if chatmodes:contains(args[1]) then
```

PR #1217, opened 2015-02-11
AutoLockstyle - Change addon name

cantreadcode:


z16:
Renaming is generally not a good idea, as it will mess with the people who have already downloaded this addon. I'll do it this time since it's still pretty new, but please try to think of a proper name for it from the start next time.

PR #1206, opened 2015-01-30
Initial Commit of Request Addon

Selindrile:
Request allows other players to send you commands using a whitelist and
blacklist for security with optional partial disables.

PR #1218, opened 2015-02-12
AdherentBuffs, new addon

Sechs:
Would love to share this new addon with LS friends and whomever else.
It announces the job and related buff of Adherent mobs in Incursion.


joshk6656:
you dont need settings.xml, it creates with the default automatically.

also whats the point of the mode in the announce function? I don't think that's correct

addons/AdherentBuffs/AdherentBuffs.lua
```diff
@@ -0,0 +1,112 @@
...
+ 
+_addon.name     = 'AdherentBuffs'
+_addon.author   = 'Mafai, Sechs'
+_addon.version  = '1.1'
+_addon.commands = {'adherentbuffs','ab'}
+ 
+defaults = T{}
+ 
+defaults.AnnounceMode = 'party' --this can be say/party/linkshell/linkshell2/shout/s/p/l/l2/sh
+defaults.AutoAnnounce = 'off' --this can be true/t/on/false/f/off
```
z16: This should be `true`/`false` (instead of a string).

addons/AdherentBuffs/AdherentBuffs.lua
```diff
@@ -0,0 +1,112 @@
...
+ 
+settings = config.load(defaults)
+ 
+adherent_maps = {['Steadfast Adherent']="PLD, DEF+", ['Furtive Adherent']="WHM, MDB+", ['Occult Adherent']="WAR, EVA+",
+		['Fleet Adherent']="WAR, Haste+", ['Brawny Adherent']="DRK, ATK+", ['Martial Adherent']="DRK,Regain+",
+		['Honed Adherent']="RDM, Fast Cast+", ['Insidious Adherent']="RDM, MEVA+", ['Hexbreaking Adherent']="BLM, MAB+"}
+chatmodes = S{'say','party','linkshell','linkshell2','shout','s','p','l','l2','sh'}
+automodes = S{'true','t','on','false','f','off'}
```
z16: Might also allow `1` and `0`, as that matches some of Windower's core commands. Should also call this `bool_values` or something, in case you ever add another setting that can be true or false.

addons/AdherentBuffs/AdherentBuffs.lua
```diff
@@ -0,0 +1,112 @@
...
+moblist = S{}
+		
+windower.register_event('addon command', function (command,...)
+	command = command and command:lower()
```
z16: Could change this to the following:

``` lua
```

That way it will default to the help command if no command is specified.

addons/AdherentBuffs/AdherentBuffs.lua
```diff
@@ -0,0 +1,112 @@
...
+	local args = T{...}
+	if command == 'reload' then
+		windower.send_command('lua unload AdherentBuffs; lua load AdherentBuffs')
```
z16: You can use `lua reload AdherentBuffs` :)

addons/AdherentBuffs/AdherentBuffs.lua
```diff
@@ -0,0 +1,112 @@
...
+		if args[1] ~= nil and chatmodes:contains(args[1]) then
+			windower.add_to_chat(038,' ***** Chat Mode changed to "'..args[1]..'" *****')
+			settings.AnnounceMode = args[1]
+			config.save(settings)
+		else
+			windower.add_to_chat(038,' ***** That is not a valid chat mode *****')
+		end
+	elseif command == 'announce' or command == 'a' then
+		announce(1)
+	elseif command == 'autoannounce' or command == 'aa' then
```
z16: This section is a perfect example of why `settings.AutoAnnounce` should be a boolean value. Also you should split up the earlier `automodes` table like so:

``` lua
false_values = S{'false', 'off', 'f', '0'}
true_values = S{'true', 'on', 't', '1'}
```

Then you can shorten the entire section to this:

``` lua
        local value = args[1] and args[1]:lower() or nil
        if not value then
            settings.AutoAnnounce = not settings.AutoAnnounce
        elseif false_values:contains(value) or true_values:contains(value) then
            settings.AutoAnnounce = not false_values:contains(args[1]:lower())
        else
            error('Incorrect syntax: //autoannounce autoannounce [true|false]')
            return
        end

        log('AutoAnnounce setting changed to %s.':format(tostring(settings.AutoAnnounce)))
        config.save(settings)
```

addons/AdherentBuffs/AdherentBuffs.lua
```diff
@@ -0,0 +1,112 @@
...
+--       notice, this list of conditions and the following disclaimer in the
+--       documentation and/or other materials provided with the distribution.
+--     * Neither the name of AdherentBuffs nor the
+--       names of its contributors may be used to endorse or promote products
+--       derived from this software without specific prior written permission.
+-- 
+-- THIS SOFTWARE IS PROVIDED BY THE COPYRIGHT HOLDERS AND CONTRIBUTORS "AS IS" AND
+-- ANY EXPRESS OR IMPLIED WARRANTIES, INCLUDING, BUT NOT LIMITED TO, THE IMPLIED
+-- WARRANTIES OF MERCHANTABILITY AND FITNESS FOR A PARTICULAR PURPOSE ARE
+-- DISCLAIMED. IN NO EVENT SHALL <your name> BE LIABLE FOR ANY
```
z16: Replace this `<your name>` with the addon author(s).

z16:
You're missing an entry in the `Lua/addons/addons.xml` file. That's required before this can be accepted.

z16:
Change all tabs to 4 spaces please.

addons/AdherentBuffs/AdherentBuffs.lua
Sechs: It used to be a boolean, but I changed it to be able to use it inside the following add_to_chat lines, where I announce "Announce Mode changed to: ..settings.AnnounceMode..". Normally it doesn't work with a boolean, and I didn't know about the tostring function. Now that I do I guess I can revert it back to a boolean and put "tostring(settings.AnnounceMode)" in the add_to_chat lines. Will do.

addons/AdherentBuffs/AdherentBuffs.lua
Sechs: Makes sense, will do the changes (once I learn how to use Github for that lol :P)

addons/AdherentBuffs/AdherentBuffs.lua
Sechs: There used to be "or 'help'" in there, I removed it because I didn't want to default to Help but... that would be a strange behaviour compared to other Windower Addons, so I take your point. Will revert it back and add that or once more.

z16: All of these things are up to you by the way. This is just input from my side, you don't have to do it if you don't want to :)

addons/AdherentBuffs/AdherentBuffs.lua
Sechs: I'll need to have a closer look to this, there's a few things you used that I need to learn about.
In general I agree that branch of the "if" wasn't smooth enough and the thing bugged me, so your suggestions are welcome.

addons/AdherentBuffs/AdherentBuffs.lua
z16: You can check out [this post I made](https://github.com/Windower/Lua/pull/1206#discussion_r24493898) regarding the exact same issue for a different addon.

Sechs:
Points taken guys, thanks for the suggestions. You too Mafai with the unncessary adding of the settings.xml, forgot to take it out.

As for your other question, it's a change I made after a few steps.
The manual announce command returns the correct message when you use it on a target that's inside the specified targets table. If you use it on a target that's not part of it, it will return the message "Target is not an Adherent".
So far so good, ok?
I wanted to avoid this second part during autoannounce.
If I move through targets with autoannounce on, I don't want to be spammed with "target is not an adherent" messages.
At the same time though, if I use "announce" manual function DURING autoannounce set to on, I DO WANT the manual command to return the "target is not an adherent" message. Yet the manual command and the autoannounce BOTH use the "announce" function, so I left that unchanged and just added a parameter. The call from the autoannouce will use the 0 parameter (it will never say "Target is not an adherent") whereas the call from the manual command will use the 1 parameter (it will say the "Target is not an adherent" message when required)
This change allowed me to save a few lines of code, simplified some things, and I also thought that adding a parameter to the announce function could've been useful for possible future evolutions of the addon, where I may want to add more messages/situations.

Does it really make no sense at all?

addons/AdherentBuffs/AdherentBuffs.lua
Sechs: I'm not "following orders". I just find your suggestion meaningful, it makes sense and I like it more than what I did, that's why I want to change it and I apreciate your input on it ;)

joshk6656:
I got your announce mode thing now. Good call.

Could you change my name to JoshK6656 where ever you have Mafai?

addons/AnnounceTarget/AnnounceTarget.lua
```diff
@@ -0,0 +1,107 @@
...
+		windower.add_to_chat(038,' *** '.._addon.name..' v'.._addon.version..' - Authors: '.._addon.author..' ***')
+		windower.add_to_chat(038,' help -> Displays this message')
+		windower.add_to_chat(038,' chatmode -> Changes chat output mode. Available settings: say/party/linkshell/linkshell2/shout')
+		windower.add_to_chat(038,' autoannounce -> Turns AutoAnnounce on or off. Available settings: on/true/false/off')
+		windower.add_to_chat(038,' announce -> Manually announces for the current target')
+		windower.add_to_chat(038,' clear -> Clears the list of announced mobs during AutoAnnounce mode on')
+	end
+end)
+
+function announce(mode)
```
trv6: function announce(name)
    if adherent_maps[name] then
        windower.send_command('input /'..settings.AnnounceMode..' '..mob.name..' buff is ==> '..adherent_maps[mob.name]..'')
    else
        windower.add_to_chat(038,' ***** Target is not an Adherent *****')
    end
end

trv6: Possibly:
```Lua
function announce(name)
    if adherent_maps[name] then
        windower.send_command('input /'..settings.AnnounceMode..' '..mob.name..' buff is ==> '..adherent_maps[mob.name]..'')
    else
        windower.add_to_chat(038,' ***** Target is not an Adherent *****')
    end
end
```

addons/AnnounceTarget/AnnounceTarget.lua
```diff
@@ -0,0 +1,107 @@
...
+	local mob = windower.ffxi.get_mob_by_target('t')
+	if mob ~= nil and adherent_maps[mob.name] then 
+		windower.send_command('input /'..settings.AnnounceMode..' '..mob.name..' buff is ==> '..adherent_maps[mob.name]..'')
+	elseif mode == 1 then
+		windower.add_to_chat(038,' ***** Target is not an Adherent *****')
+	end
+end
+ 
+windower.register_event('target change',function(...)
```
trv6: The target change event is pretty slow (possibly based on 0x015 outgoing?), but it does pass the index of the mob as a parameter.
```Lua
windower.register_event('target change',function(index)
    if settings.AutoAnnounce and index ~= 0 then
        local mob = windower.ffxi.get_mob_by_index(index)
        if not moblist:contains(mob.id) then
            announce(mob.name)
            moblist:add(mob.id)
        end
    end
end)
```

addons/AnnounceTarget/AnnounceTarget.lua
```diff
@@ -0,0 +1,107 @@
...
+		windower.send_command('lua unload AnnounceTarget')
+	elseif command == 'chatmode' or command == 'cm' then
+		if args[1] ~= nil and chatmodes:contains(args[1]) then
+			windower.add_to_chat(038,' ***** Chat Mode changed to "'..args[1]..'" *****')
+			settings.AnnounceMode = args[1]
+			config.save(settings)
+		else
+			windower.add_to_chat(038,' ***** That is not a valid chat mode *****')
+		end
+	elseif command == 'announce' or command == 'a' then
```
trv6: Only if you change the announce function:
```Lua
elseif command == 'announce' or command == 'a' then
    local mob = windower.ffxi.get_mob_by_target('t')
    if mob then 
        announce(mob.name)
    end
```

addons/AnnounceTarget/AnnounceTarget.lua
Sechs: Interesting change, but would kinda need to add the "name" parameter to the Announce function, instead of changing it from "mode" to "name". As I explained earlier to Joshk the "mode" parameter is necessary to handle a complicated exception when AutoAnnounce is active.

joshk6656:
Dont go too crazy with changes I have a few ideas when I get home in 15 mins

addons/AnnounceTarget/AnnounceTarget.lua
Sechs: Pardon my ignorance, not getting this 100%. Are you suggesting that using get_mob_by_index with the index parameter is going to be more reliable/efficient than the previous version relying on target? Aside from this change it looks pretty much the same, but if it's a performance issue I'm more than willing to commit this change.

joshk6656: Sechs can you check in your latest version to your branch please? If it isn't already?

addons/AnnounceTarget/AnnounceTarget.lua
trv6: That makes sense. You can still avoid the mode parameter if you want to.
Condensed comments to a single comment since the line markers weren't as useful as I thought they'd be:

``` Lua
function announce(name)
    if adherent_maps[name] then
        windower.send_command('input /'..settings.AnnounceMode..' '..name..' buff is ==> '..adherent_maps[name])
    else
        windower.add_to_chat(038,' ***** Target is not an Adherent *****')
    end
end
```

The target change event is pretty slow (possibly based on 0x015 outgoing?), but it does pass the index of the mob as a parameter.

``` Lua
windower.register_event('target change',function(index)
    if settings.AutoAnnounce and index ~= 0 then
        local mob = windower.ffxi.get_mob_by_index(index)
        if adherent_maps[mob.name] and not moblist:contains(mob.id) then -- this stops announce() from spamming stuff on non-adherent mobs
            moblist:add(mob.id)
            announce(mob.name)
        end
    end
end)
```

Only if you change the announce function:

``` Lua
elseif command == 'announce' or command == 'a' then
    local mob = windower.ffxi.get_mob_by_target('t')
    if mob then 
        announce(mob.name)
    end
```

addons/AnnounceTarget/AnnounceTarget.lua
trv6: It may be more reliable if you're using a subtarget. As it is, get_mob_by_target would continue to see the same mob, while get_mob_by_index using the index parameter would see your subtarget.

Sechs:
Updated the last version on my branch. It's currently work in progress but I stopped.
Tested the clear on zone (works) and was testing out the log function (from logger library). Think I might eventually change all windower.add_to_chat() into log()
Also tested some changes I did this morning under suggestions from Arcon, they all seem to work.

addons/AnnounceTarget/AnnounceTarget.lua
z16: The target change function is indeed very slow, but I will adjust that today to check based on the PreRender event.

Sechs:
Trv I can now see the changes you made to avoid autoannounce mode spamming stuff on non adherent mobs. I like it, thanks man! Think I'm going to implement that change eventually, together with the different announce function, but probably not atm. I want to come out with the addon first, with a working version. For future releases I'm already planning further changes to both autoannounce and announce.
I plan to add a secondary tablemap with mob IDs and a relative chat message to handle placeholders for NMs. This will require some changes in the way I handle some checks and especially it will require me to use dynamic fields within the chat messages, so that they can remain meaningful for all types of target.

The thing about the target_by_index in place of target_by_name is another story and I might change that now, want to ask a couple of things to Arcon first but that's likely to happen now.

Sechs:
Tested in-game, everything seems to be working (had to edit a few minor typos).
Please review this for a possible official release, thanks.

addons/addons.xml
```diff
@@ -17,6 +17,11 @@
     <bugtracker>https://github.com/nitrous24/Lua/issues?page=1</bugtracker>
   </addon>
   <addon>
+    <name>AnnounceTarget</name>
+    <author>JoshK6656, Sechs</author>
+    <description>Creates special manual or automatic chat messages conveying advanced information about some specific targets</description>
```
z16, line 22: Please also add a bugtracker (can use [Windower Lua's issue tracker](https://github.com/Windower/Lua/issues)) and a support site (can use Windower's IRC channel, see some other addons that did that).

addons/AnnounceTarget/AnnounceTarget.lua
```diff
@@ -0,0 +1,112 @@
...
+            settings.AutoAnnounce = not false_values:contains(args[1]:lower())
+        else
+            error(' ***** "'..args[1]..'" is not a valid setting for AutoAnnounce *****')
+            return
+        end
+        log(' ***** AutoAnnounce changed to "',settings.AutoAnnounce,'" *****')
+        config.save(settings)
+    elseif command == 'clear' or command == 'c' then
+        moblist:clear()
+		log(' ***** Previously announced targets table cleared *****')
```
z16: For some reason you forgot to change the tabs to spaces on this line alone :)

addons/AnnounceTarget/AnnounceTarget.lua
```diff
@@ -0,0 +1,112 @@
...
+--       notice, this list of conditions and the following disclaimer in the
+--       documentation and/or other materials provided with the distribution.
+--     * Neither the name of AdherentBuffs nor the
+--       names of its contributors may be used to endorse or promote products
+--       derived from this software without specific prior written permission.
+-- 
+-- THIS SOFTWARE IS PROVIDED BY THE COPYRIGHT HOLDERS AND CONTRIBUTORS "AS IS" AND
+-- ANY EXPRESS OR IMPLIED WARRANTIES, INCLUDING, BUT NOT LIMITED TO, THE IMPLIED
+-- WARRANTIES OF MERCHANTABILITY AND FITNESS FOR A PARTICULAR PURPOSE ARE
+-- DISCLAIMED. IN NO EVENT SHALL JoshK6656, Sechs BE LIABLE FOR ANY
```
z16: `JoshK6656 or Sechs`

addons/AnnounceTarget/AnnounceTarget.lua
```diff
@@ -0,0 +1,112 @@
...
+-- 
+-- Redistribution and use in source and binary forms, with or without
+-- modification, are permitted provided that the following conditions are met:
+-- 
+--     * Redistributions of source code must retain the above copyright
+--       notice, this list of conditions and the following disclaimer.
+--     * Redistributions in binary form must reproduce the above copyright
+--       notice, this list of conditions and the following disclaimer in the
+--       documentation and/or other materials provided with the distribution.
+--     * Neither the name of AdherentBuffs nor the
```
z16: `AnnounceTarget`

addons/AnnounceTarget/AnnounceTarget.lua
```diff
@@ -0,0 +1,112 @@
...
+        log(' chatmode -> Changes chat output mode. Available settings: say/party/linkshell/linkshell2/shout/echo')
+        log(' autoannounce -> Turns AutoAnnounce on or off. Accepted settings: on/true/false/off')
+        log(' announce -> Manually announces for the current target')
+        log(' clear -> Clears the list of announced mobs during AutoAnnounce mode on')
+    end
+end)
+
+function announce(name)
+    if adherents_map[name] then
+        windower.send_command('input /'..settings.AnnounceMode..' '..name..' buff is ==> '..adherents_map[name])
```
z16, line 96: Just a note, I'll be adding `windower.input` to the next version of LuaCore, so you don't need to do `windower.send_command('input ...')` anymore.

Sechs:
Think I fixed everything you told me.
Neat for the windower.input thing :) Will need to remember to change that.

PR #1220, opened 2015-02-14
Adding features to TargetInfo

KenshiDRK:
Added claim id and target index features as well as supporting subtarget.

### DIFF
--- a/addons/targetinfo/targetinfo.lua
+++ b/addons/targetinfo/targetinfo.lua
@@ -13,9 +13,7 @@ defaults.ShowHexID = true
 defaults.ShowFullID = true
 defaults.ShowSpeed = true
 defaults.ShowClaimName = true
-defaults.ShowClaimID = true
 defaults.ShowTargetName = true
-defaults.ShowTargetID = true
 defaults.display = {}
 defaults.display.pos = {}
 defaults.display.pos.x = 0
@@ -38,11 +36,6 @@ settings:save()
 
 text_box = texts.new(settings.display, settings)
 
--- Variables for lines visibility
-
-ClaimIsVisible = true
-TargetIsVisible = true
-
 -- Constructor
 
 initialize = function(text, settings)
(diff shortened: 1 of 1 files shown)


z16:
I really dislike the feature that it hides lines that you don't consider relevant. This should at least be toggleable and default to off. It also messes up the code in a few places, because you call the initialize function (an expensive function which should only be called on login/logout) on every render frame, dozens of times each second. initialize should _never_ be called manually (the text library calls it itself when logging in/out).

If you really want those lines to auto-hide either make an option for it or submit TargetClaim again where it defaults to on. I hope you didn't think that I was telling you not to submit TargetClaim, I was just asking why you didn't try to add it to this, because it seemed like it would go well with this addon (which I still think it does). But this approach makes the on-screen display inconsistent, makes the code slow to execute and even confusing to maintain, because it introduces global state which wasn't there before and is called in a function that the user is not meant to call at all.

KenshiDRK:
Yea thats why I said this way of hiding them should be avoided, but I don't know another way. And it is only hiding the lines when there is no value to display so this way the box only shows 5 lines instead of 7 with 2 lines showing nothing. Another way is to label the lines the same, but this could lead to confusion. Any suggestion?

z16:
You could wrap the entire line in a variable, so instead of `'Claim: ${claim_name||%16s}'` you would do `'${claim_name||\nClaim: %16s}'`. But that should _still_ be optional, making any lines variable is confusing, because the layout of the text box will change all the time when mobs or players target different things. Instead it should default to just not displaying anything (or a `-` or something) when nothing is targeted or no one has it claimed.

KenshiDRK:
I already tried displaying a default value but after targetting something it won't go to the default, it will show the last value, leading to showing someone is targetting something when he is not targetting anything. I'll hop into irc so we can discuss this better.

addons/targetinfo/targetinfo.lua
```diff
@@ -72,15 +82,26 @@ windower.register_event('prerender', function()
...
+        else
+            remove:add('claim_name')
+            remove:add('target_name')
+        end
         text_box:update(info)
         text_box:show()
-	else
-		text_box:hide()
-	end
+        for entry in remove:it() do text_box[entry] = nil end
```
z16: This feature should still be optional and default to off. Also please space it out properly with new lines:

``` lua
        for entry in remove:it() do
            text_box[entry] = nil
        end
```

addons/targetinfo/targetinfo.lua
```diff
@@ -72,15 +82,26 @@ windower.register_event('prerender', function()
...
+    else
+        text_box:hide()
+    end
 end)
 
 --[[
-Copyright (c) 2013-2014, Windower
+Copyright (c) 2013-2015, Windower
```
z16: Since you're adjusting it already please change the `(c)` to a `©`.

PR #1227, opened 2015-02-21
setbgm - add new songs, reorganize code

svanheulen:


addons/setbgm/setbgm.lua
```diff
@@ -61,7 +62,7 @@ bgm = {
     [150]='Ululations from Beyond', [151]='The Federation of Windurst', [152]='The Republic of Bastok', [153]='Prelude', [154]='Metalworks', [155]='Castle Zvahl', [156]="Charteau d'Oraguille", [157]='Fury', [158]='Sauromugue Champaign', [159]='Sorrow',
```
z16, line 62: This is likely a typo, I'm guessing it's supposed to be _Chateau d'Oraguille_.

PR #1238, opened 2015-03-08
Nostrum

trv6:


addons/Nostrum/variables.lua
```diff
@@ -1,4 +1,4 @@
---[[Copyright © 2014-2015, trv
+--[[Copyright © 2014, trv
```
z16: Why did you remove 2015 from here?

trv6: Daylight... savings... time?
Fixed.

alphaONE2: ROFL! Thank you, you just made my day!

z16: Yeah, those can be a bitch :\

PR #1250, opened 2015-03-27
Minor library errors

trv6:


addons/libs/lists.lua
```diff
@@ -312,9 +312,11 @@ function list.clear(l)
 end
 
 function list.copy(l, deep)
-    deep = deep or true
+    if deep == nil then 
+        deep = true
+    end
```
z16: Good catch, this can be `deep = deep ~= false and true`. Can you please also adjust it in `tables.lua` and `sets.lua`?

addons/libs/lists.lua
```diff
@@ -312,9 +312,11 @@ function list.clear(l)
...
     local res = {}
-
+
```
z16: :'(

trv6:
Adjusted sets and tables copy methods as well. Squashed commit (onto the wrong commit!).

addons/libs/lists.lua
trv6: :'(

PR #1251, opened 2015-03-27
Rhombus: empty menus

trv6:


addons/Rhombus/Rhombus.lua
```diff
@@ -558,7 +561,7 @@ function format_response(n,p,bool)
 end
 
 windower.register_event('keyboard', function(dik, down, flags, blocked)
-    if dik == 42 and not (bit.band(flags,32) == 32) then
+    if dik == 42 and flags < 32 then
```
z16, line 564: Are you sure this will work? It will no longer check for just the sixth bit, but it will fail if any of the sixth, seventh or eight bit are set.

trv6, line 564: I think it's just a 6\* bit number.

trv6, line 564: In an old file I have:

```
Shift,1
Alt,2
Ctrl,4
Win,8
App,16
Chatlog is open,32
```

I never found any others.

z16, line 564: While you're right, and it should work, I still wouldn't recommend it, for two reasons. First, it makes the intent less clear and makes it seem like you're working with a number instead of a bit field. Second, it could break if we do add anything more to it. And for the record, it's actually 7 bits long, but LuaCore currently filters the seventh bit out and passes it as its own value.

The entire bit field is built like this:

``` C++
    flags |= ShiftIsDown << 0;
    flags |= AltIsDown << 1;
    flags |= ControlIsDown << 2;
    flags |= WinIsDown << 3;
    flags |= AppsIsDown << 4;
    flags |= ChatOpen() << 5;
    flags |= down << 6;
```

It would probably help if you made a helper function to check for whether or not a bit is set:

``` lua
bit.is_set = function(val, pos)
    return bit.band(val, 2^(pos - 1)) > 0
end
```

That would make the intent very clear (more so than either of the previous solutions):

``` lua
    if dik == 42 and not bit.is_set(flags, 6) then
```

It's up to you, by the way, this is just a suggestion, I'm ready to pull this at any time.

alphaONE2, line 564: I agree that this is brittle, we make no promises about the stability of the unused flag bits.

trv6, line 564: I'll make the changes you suggested. There was just something appealing about the simplicity of a single comparison.

trv6, line 564: Also, I appear to have made a bad merge, which will take a bit to figure out.

PR #1264, opened 2015-04-07
Scoreboard - Enable DPS without engaging

lorand-ffxi:
This is actually my first time ever making a pull request, so I'm not sure what the etiquette is, but here goes...

I always wished that Scoreboard would track damage while not engaged, especially while playing BLM, SCH, etc.  In order to do this, I added a table to store IDs of mobs that have been attacked.  Damage from party members is always recorded rather than only when a party member is engaged.  The DPS clock will now advance if the mob's HP is above 0 and the mob is engaged (rather than when a party member is engaged), otherwise the mob is removed from the list.

### DIFF
--- a/addons/scoreboard/scoreboard.lua
+++ b/addons/scoreboard/scoreboard.lua
@@ -18,6 +18,8 @@ local display
 dps_clock = require('dpsclock'):new() -- global for now
 dps_db    = require('damagedb'):new() -- global for now
 
+local enemies = {}
+
 -------------------------------------------------------
 
 -- Conventional settings layout
@@ -258,6 +260,10 @@ windower.register_event('addon command', function()
     end
 end())
 
+windower.register_event('zone change', function(new_id, old_id)
+	enemies = {}
+end)
+
 local months = {
     'jan', 'feb', 'mar', 'apr',
     'may', 'jun', 'jul', 'aug',
@@ -308,23 +314,20 @@ display = Display:new(settings, dps_db)
 
 -- Keep updates flowing
 local function update_dps_clock()
-    local player = windower.ffxi.get_player()
-    local pet
-    if player ~= nil then
-        local player_mob = windower.ffxi.get_mob_by_id(player.id)
-        if player_mob ~= nil then
-            local pet_index = player_mob.pet_index
-            if pet_index ~= nil then
-                pet = windower.ffxi.get_mob_by_index(pet_index)
-            end
+    local fighting = false
+    for id,_ in pairs(enemies) do
+        local mob = windower.ffxi.get_mob_by_id(id)
+        if (mob ~= nil) and (mob.hpp > 0) then
+            fighting = true
+        else
+            enemies[id] = nil
         end
     end
-    if player and (player.in_combat or (pet ~= nil and pet.status == 1)) then
+    if fighting then
         dps_clock:advance()
     else
         dps_clock:pause()
     end
-
     display:update()
 end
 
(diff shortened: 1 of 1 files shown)


z16:
While I personally like this change, I believe it was done the way it is for a reason (I remember talking about the issue in the past). You should probably make the change optional by adding a setting for it. Maybe @Byrth or @jerryhebert can comment on this.

lorand-ffxi:
Understandable; thanks for the feedback.  I added the original code back in, a new settings value for it, and the command `//sb requireEngaged` that can take an argument of true/false, or will toggle it.

jerryhebert:
The "in_combat" flag isn't the same as engaged. I think you are likely experiencing this issue because of the obscure command in FFXI that lets you toggle whether or not engaged people get battle music.

Like Arcon said, we had tons of discussion around this and a lot of consideration was put into it. The gist was that we wanted the damage to work for melees as well as for BLMs and anyone else really, in a way that you'd like. The problem then was determining how "in combat" should be computed.

There's a signature within FFXI for whether or not you're in combat, but I'm pretty sure that the signature we were using was actually directly targeting whether or not battle music was playing, and IIRC, Iryoku was thinking that there may have been a server-side aspect which was forcing us to do it this way.

So the summary of how things were, at least the time when I wrote it, was Scoreboard tracks damage whenever Lua reports `is_combat` is true (this seems like the correct abstraction). Within Windower, `is_combat` is true whenever the signature is true, which is the same as whenever battle music is playing (this seems like the incorrect implementation but we couldn't come up with a better implementation). Other cases were considered which made things even more complicated, such as how to handle the semantics around this within Legion.

Anyway, the most important point is that you may just not know about the FFXI command that toggles battle music when you aren't engaged. SE added this for BLMs who wanted to hear the music, and then we ended up using it within Windower.

I think basing SB tracking off of whether or not a user is engaged is a mistake for the general case, at least, and if it's going to be added at all, it should be an option for people who don't like the default behavior. I haven't played FFXI in about a year now though so I can't remember the name of that command but I'm willing to bet if you set it so that BLMs can hear music when they aren't engaged, that you'll see SB start to do what you want.

jerryhebert:
Actually, here's the command:
http://wiki.ffxiclopedia.org/wiki/Command/battlebgm

lorand-ffxi:
Thanks; I usually play with sound off, and had no clue about that command.  I turned my sound up, and was able to hear the battle music while nuking a mob, but not engaging.

I actually just tested it though, and regardless of whether or not battle music is playing, if no one in the party is engaged, windower.ffxi.get_player().in_combat is false.

jerryhebert:
Cool. In that case, the signature may have changed (although I think there are asserts about these) or SE changed the semantics around how they use this state.

If we don't have a better solution, adding it as an option as you have seems fine to me. Thanks for pitching in!

lorand-ffxi:
Happy to try to give back after using this and other addons for so long ^^;

z16:
Sorry this has been ignored for so long! I'll get right on checking if `in_combat` works now, and if not, how to fix it. In the meantime, this PR has conflicts that cannot be merged as it is. If you have some free time, do you think you can get it up to date?

Although you may wanna wait until I figure out the `in_combat` issue... up to you.

z16:
I just checked and `in_combat` had a very obvious issue, which I have now fixed.

However, there is another issue that is not quite as obvious. If you lose claim to a mob while it's still attacking you I don't know any way to figure out that you are still in combat. There's also no combat music playing anymore at that point. You could scan incoming packets I suppose to see if the mob is actually attacking or trying to attack, but that would be vague at best. I'm open for suggestions on this.

jerryhebert:
I'm not sure myself. Like you, I just remember we had lots of discussions about this. There was that native SE command to toggle battle music even if you aren't engaged though and that solved some cases. I can't remember what it was called. I also remember some stuff like Legion was tricky.

Honestly, this kind of flag seems pretty nice because it lets the user just specify it rather than trying to do it automatically with unreliable triggers.

There's some rambling for historical purposes.

lorand-ffxi:
No worries!  I hadn't touched anything FFXI-related for almost a year until a couple weeks ago.  When I get some time, I'll look in to resolving the conflicts, and see if I can think of any way to address the situation that you mentioned.

PR #1270, opened 2015-05-01
BarFiller

Morath86:
Addons\BarFiller: initial commit


addons/barfiller/data/settings.xml
```diff
@@ -0,0 +1,29 @@
+<?xml version="1.1" ?>
```
z16: Do not provide a settings file, it will be automatically generated. The data folder should generally stay empty.

addons/barfiller/statics.lua
```diff
@@ -0,0 +1,195 @@
...
+		ON ANY THEORY OF LIABILITY, WHETHER IN CONTRACT, STRICT LIABILITY, OR TORT
+		(INCLUDING NEGLIGENCE OR OTHERWISE) ARISING IN ANY WAY OUT OF THE USE OF THIS
+		SOFTWARE, EVEN IF ADVISED OF THE POSSIBILITY OF SUCH DAMAGE.
+]]
+
+-- Default settings file
+default_settings = {
+	bar_settings = {
+		background = { -- image: Must be the EXACT path to your Windower\Addon folder
+			image = 'C:\\Program Files (x86)\\Windower4\\addons\\barfiller\\data\\bar_bg.png',
```
z16: Do not use hard-coded paths. You can obtain the user's Windower directory with the `windower.windower_path` variable, and the current addon's directory with `windower.addon_path`. Use those instead of hard-coding the absolute path

On a side note, if you do ever need to use user-configurable options, put those in the settings, ideally users should never be asked to edit code.

addons/barfiller/barfiller.lua
```diff
@@ -0,0 +1,118 @@
...
+windower.prim.set_visibility('foregroundBar', true)
+
+-- windower.register_event('login',function(name)
+--     windower.send_command('lua i barfiller clear;')
+-- end)
+
+-- Addon commands
+-- Thanks to Byrth & SnickySnacks' BattleMod addon
+windower.register_event('addon command',function(...)
+	local commands = {...}
```
z16: Use 4 spaces for indentation.

addons/barfiller/barfiller.lua
```diff
@@ -0,0 +1,118 @@
...
+			print('BarFiller successfully reset the experience counter.')
+		elseif first_cmd == 'reload' or first_cmd == 'r' then	-- Reloads BarFiller
+			windower.send_command('lua r barfiller;')
+		elseif first_cmd == 'unload' or first_cmd == 'u' then	-- Unloads BarFiller
+			windower.send_command('lua u barfiller;')
+		elseif first_cmd == 'help' or first_cmd == 'h' then		-- Display helpful information
+			help()
+		end
+	else
+		print('Missing or invalid command, use "bf help" or "bf h" for list of available commands.')
```
z16: May as well call the help function here instead of telling people to call the help function :)

addons/barfiller/statics.lua
```diff
@@ -0,0 +1,195 @@
...
+		(INCLUDING, BUT NOT LIMITED TO, PROCUREMENT OF SUBSTITUTE GOODS OR SERVICES;
+		LOSS OF USE, DATA, OR PROFITS; OR BUSINESS INTERRUPTION) HOWEVER CAUSED AND
+		ON ANY THEORY OF LIABILITY, WHETHER IN CONTRACT, STRICT LIABILITY, OR TORT
+		(INCLUDING NEGLIGENCE OR OTHERWISE) ARISING IN ANY WAY OUT OF THE USE OF THIS
+		SOFTWARE, EVEN IF ADVISED OF THE POSSIBILITY OF SUCH DAMAGE.
+]]
+
+-- Default settings file
+default_settings = {
+	bar_settings = {
```
z16: The naming convention for settings files is _PascalCase_, i.e. capitalize the first letter of each word in a multi-word identifier, without underscores.

addons/barfiller/statics.lua
```diff
@@ -0,0 +1,195 @@
...
+			size = {
+				height = 5,
+				width = 1
+			}
+		}
+	}
+}
+
+-- Current Game Resolution
+windowerResX = windower.get_windower_settings().x_res
```
z16: The naming convention in code is to use _snake_case_, i.e. all lower cased letters with underscores separating single words in an identifier.

addons/barfiller/barfiller.lua
```diff
@@ -0,0 +1,118 @@
...
+        local val = org:unpack('I',0x11)
+        local msg = org:unpack('H',0x19)%1024
+        exp_msg(val,msg)
+	elseif id == 0x61 then
+        xp.current = org:unpack('H',0x11)
+        xp.total = org:unpack('H',0x13)
+        xp.tnl = xp.total - xp.current
+		calcExpBarPerc()
+	end
+end)
```
z16: The world would be a better place if every file ended with a new line :)

addons/barfiller/barfiller.lua
```diff
@@ -0,0 +1,118 @@
...
+		print('Missing or invalid command, use "bf help" or "bf h" for list of available commands.')
+	end
+end)
+
+-- Capture XP Values
+-- Thanks to Byrth's PointWatch addon
+windower.register_event('incoming chunk',function(id,org,modi,is_injected,is_blocked)
+	if is_injected then return end
+	if id == 0x2D then
+        local val = org:unpack('I',0x11)
```
z16: You may consider using the `packets` library here. The advantage is that if packets ever change, the library will be the first to be adjusted, and all addons using it will automatically work again, without having to be touched.

(And it's also prettier to look at than manually extracted bits and bytes.)

addons/barfiller/barfiller.lua
Morath86: Done.

addons/barfiller/barfiller.lua
Morath86: Done.

addons/barfiller/statics.lua
Morath86: Done.

addons/barfiller/statics.lua
Morath86: Done.

addons/barfiller/barfiller.lua
Morath86: Done.

addons/barfiller/barfiller.lua
Morath86: Will investigate in future update.

addons/barfiller/statics.lua
Morath86: Done.

addons/barfiller/barfiller.lua
smd111: this is what you should need 

```
windower.register_event('incoming chunk',function(id,org,modi,is_injected,is_blocked)
    if is_injected then return end
    if id == 0x2D then
        local packet = packets.parse('incoming', data)
        exp_msg(packet['Param 1'],packet['Message'])
    elseif id == 0x61 then
        local packet = packets.parse('incoming', data)
        xp.current = packet['Current EXP']
        xp.total = packet['Required EXP']
        xp.tnl = xp.total - xp.current
        calcExpBarPerc()
    end
end) 
```

just so you know i have not tested this yet

Morath86: That actually worked, thank you for that!

Morath86:
Changed console output to chat window output

addons/barfiller/barfiller.lua
```diff
@@ -0,0 +1,88 @@
...
+file = require('files')
+packets = require('packets')
+math = require('math')
+
+--BarFiller Libs
+require('statics')
+
+-- Generate Settings Files
+-- Thanks to Byrth & SnickySnacks' BattleMod addon
+settings_file = file.new('data\\settings.xml',true)
```
z16: You never use this variable. If you just want to create a file, without using the variable, you don't have to assign it, you can just do:

``` lua
file.new('<filename>', true)
```

That said, in this case you don't need it at all. The config library will automatically create both the data folder and the settings file, so you can safely remove this line.

addons/barfiller/barfiller.lua
```diff
@@ -0,0 +1,88 @@
...
+settings_table = config.load('data\\settings.xml',DefaultSettings)
```
z16: Since `data/settings.xml` is the default location you don't need to specify it. In fact, you probably _shouldn't_ specify it, unless you want it to be something other than the default. That way the config library is free to make its own assumptions about what and where it should be and is one less thing the addon developer needs to consider.

addons/barfiller/statics.lua
```diff
@@ -0,0 +1,232 @@
...
+        DIRECT, INDIRECT, INCIDENTAL, SPECIAL, EXEMPLARY, OR CONSEQUENTIAL DAMAGES
+        (INCLUDING, BUT NOT LIMITED TO, PROCUREMENT OF SUBSTITUTE GOODS OR SERVICES;
+        LOSS OF USE, DATA, OR PROFITS; OR BUSINESS INTERRUPTION) HOWEVER CAUSED AND
+        ON ANY THEORY OF LIABILITY, WHETHER IN CONTRACT, STRICT LIABILITY, OR TORT
+        (INCLUDING NEGLIGENCE OR OTHERWISE) ARISING IN ANY WAY OUT OF THE USE OF THIS
+        SOFTWARE, EVEN IF ADVISED OF THE POSSIBILITY OF SUCH DAMAGE.
+]]
+
+-- Default settings file
+DefaultSettings = {
```
z16: Earlier when I said the settings should be _PascalCase_ I didn't mean the settings table itself, but the settings values. I.e.:

``` lua
defaults = {
    Bars = {
        Background = {
            Image = ...,
            Pos = {
                X = 164,
                Y = 6,
            },
        },
    },
}
```

addons/barfiller/statics.lua
```diff
@@ -0,0 +1,232 @@
...
+    bars = {
+        background = {
+            image = windower.windower_path..'addons\\barfiller\\data\\bar_bg.png',
```
z16: While this works now I'm still not sure if it makes much sense to put this into the settings. Do you want users to be able to adjust it? I don't think many of them will actually change the images themselves.

If that actually is what you intended, I think there's a better way by just making a `Windower/addons/BarFiller/data/images/` folder which users can use to put their own files in to override the original files which should reside outside of the data folder. In general the addon should be completely operational when users delete their data folder, and currently it wouldn't be, because it would delete the images.

addons/barfiller/statics.lua
```diff
@@ -0,0 +1,232 @@
...
+            pos = {
```
z16: Is this necessary? I took from your forum post that it automatically centers the addon, so using a positional value doesn't seem to make much sense. Unless you add another setting `Center` and default it to `true`. When it's `true`, it can automatically center it, otherwise it can use what the user specifies under `Pos`.

addons/barfiller/statics.lua
```diff
@@ -0,0 +1,232 @@
...
+                x = 164,
+                y = 6
+            },
+            size = {
```
z16: Isn't the size given by the image? Do you need to specify it here?

Ideally you could do for the background the same thing you did for the foreground, i.e. make just a one pixel wide image and stretch it to the size the user specifies here. Then you'd need to put the left and right brackets into their own images, which would be a bit more work, but it would make it more flexible. Could consider that for a future update.

addons/barfiller/barfiller.lua
```diff
@@ -0,0 +1,88 @@
...
+config.save(settings_table)
+
+initialize()
+
+-- Addon commands
+-- Thanks to Byrth & SnickySnacks' BattleMod addon
+windower.register_event('addon command',function(...)
+    local commands = {...}
+    local first_cmd = table.remove(commands,1):lower()
```
z16: Instead of doing this, there's a simpler way:

``` lua
    command = (command or 'help'):lower()
    local args = {...}

    -- Rest here
end)
```

This will automatically make the first value the command and set it to `'help'` if it doesn't exist (your approach would currently result in a Lua error if the first command was missing).

z16:
All of the current suggestions are just that, suggestions. The code itself looks fine and all the essentials are implemented. Let me know if you still want some time to implement those suggestions or if you want me to merge it as it is now.

addons/barfiller/barfiller.lua
Morath86: Thank you for the clear explanations, I have removed that line at this time.

Morath86:
These are all great suggestions, and are helping me to better understand Lua. This is my first time working with it, and I appreciate all the criticism. I suspect I'll have another push ready in an hour or so.

addons/barfiller/statics.lua
Morath86: I'll get around to that when I get into changing the colours, and allowing for changing sizes more easily.

addons/barfiller/statics.lua
Morath86: I thought to myself it was a good idea to put the smallest values for 800x600 resolution just as a fail safe in case the math somehow broke.

When I get around to allowing moving the users to change the `Pos` in the commands I'll fix this up.

addons/barfiller/barfiller.lua
Morath86: It was indeed, that has now been amended. Thanks.

addons/barfiller/statics.lua
Morath86: Figured I misinterpreted that. Thanks again.

Morath86:
Added EXP/hr counter, fewer calculation updates, and now compatible at character selection screen. Ready for public usage at this point.

addons/barfiller/statics.lua
```diff
@@ -0,0 +1,433 @@
...
+        ON ANY THEORY OF LIABILITY, WHETHER IN CONTRACT, STRICT LIABILITY, OR TORT
+        (INCLUDING NEGLIGENCE OR OTHERWISE) ARISING IN ANY WAY OUT OF THE USE OF THIS
+        SOFTWARE, EVEN IF ADVISED OF THE POSSIBILITY OF SUCH DAMAGE.
+]]
+
+-- Default settings
+defaults = {
+    Bars = {
+        Background = {
+            Image = windower.windower_path..'addons\\barfiller\\data\\bar_bg.png',
```
z16: I still advise very much against both specifying any paths in the settings file and having the icons in the `data` sub folder.

addons/barfiller/barfiller.lua
```diff
@@ -0,0 +1,116 @@
...
+-- Thanks to Byrth & SnickySnacks' BattleMod addon
+settings_table = config.load(defaults)
+config.save(settings_table)
+
+ready = false
+chunk_update = false
+
+-- Make sure character is logged in, and loaded before initializing
+windower.register_event('load',function()
+    if windower.ffxi.get_info().logged_in and windower.ffxi.get_player() then
```
z16: The second check is unnecessary, if `windower.ffxi.get_info().logged_in` is true then `windower.ffxi.get_player()` will also exist.

addons/barfiller/barfiller.lua
```diff
@@ -0,0 +1,116 @@
...
+        initialize() -- Populate character details
+    end
+end)
+
+-- Delay initialize() for 10 seconds to allow game to download chunks
+windower.register_event('login',function()
```
z16, line 64: This function should probably also call `initialize`. Currently `initialize` is only called when loading the addon, not when relogging.

Edit: Nevermind, I just saw that that's what the next function does. In which case there's a better way to write it:

``` lua
    initialize:schedule(10)
```

But why the 10 sec delay in the first place? Are you waiting for a specific packet? In which case you may wanna check for it with the `incoming chunk` event.

addons/barfiller/barfiller.lua
```diff
@@ -0,0 +1,116 @@
...
+    windower.send_command('wait 10;barfiller clear;')
+end)
+
+-- If you're switching characters this will clear the addon from memory
+windower.register_event('logout',function()
+    windower.send_command('lua r barfiller')
```
z16: Addons should be made to be able to handle relogging. There is not a lot that you need to consider for it. Reloading just unnecessarily spams the console and possibly removes previous unsaved settings.

z16:
If you want I can merge this, but what I wrote about settings needs to be done now, or else it may screw up peoples' settings files in the future. Do you want to move that out of the settings? If so I'll wait with the merge, otherwise I'll do it.

addons/barfiller/statics.lua
alphaONE2: I second this. I highly suggest this be changed before the addon is merged. Windower _will not_ overwrite or delete files in an addon's data directory. The directory doesn't belong to the addon, it belongs to the users. This was an intentional design decision so that addons could provide a safe location for user files that should not be overwritten during updates.

If you place images here you can't change them in the future. Things that an addon needs to function should not be in the data directory. Users should be able to delete the data directory and the addon should still be fully functional.

For the initial release I suggest using hard-coded paths (don't include them in settings) and moving the images into a separate directory, outside of the data directory. We can discuss the options you have for allowing customization for the next release with you on IRC if you want.

TL;DR: As this is you have _no_ upgrade path, you really cant change or update your images in the future if you leave it like this. For instance, you won't be able to make the bar change color dynamically, because the color is baked into the image. And, you wont be able to update that image, because it's in the data directory.

Morath86: I'll try and get on IRC tomorrow evening to discuss this one.

EDIT: I moved all 3 images to the root folder and changes the lines to
`Image = windower.addon_path..'file_names.png'`

addons/barfiller/barfiller.lua
Morath86: Got it, thank you!

addons/barfiller/barfiller.lua
Morath86, line 64: I haven't figured out why, but without delaying the `initialize()` until you're fully logged in (on average 5-10 seconds on my slow 3Mbps connection) the player data would not populate. Exp would still calculate, but the job, and level details wouldn't load.

This is the current problem I'm looking into, trying to find out which chunk may be the relevant one to wait for.

addons/barfiller/barfiller.lua
Morath86: That was a (very poor) cheat on my part, I've changed it to hide all the `prim` and `text` objects in my next push.

addons/barfiller/statics.lua
alphaONE2: OK, that at least gives you an upgrade path that wont force users to reinstall the addon.

Morath86:
Added new Image library, updated prims to new use Objects, additional getters and setters added.

Morath86:
Smooth animation logic added to `prerender` event

Morath86:
Added err checking for when calc_exp_bar() returns nil

Morath86:
No more delay on startup/login, and re-added `frame_count` to prevent text flicker

alphaONE2:
@Morath86, @z16: Are we still waiting on anything here or can this get merged?

Morath86:
@IryokuChevalier @z16 Unless you see any other critical issues, I believe it's ready. I haven't encountered any problems tonight, and just fixed my last known issue.

addons/libs/images.lua
```diff
@@ -0,0 +1,344 @@
...
+                return
+            end
+        end
+    end
+end
+
+return images
+
+--[[
+Copyright © 2013-2015, Windower
```
z16: This should be `2015` only, since the library didn't exist before.

z16:
@IryokuChevalier @Morath86 I was really busy the last few days, and since this is adding a new library I believe it deserves some more scrutiny than just an addon commit. I'm going over it right now.

addons/barfiller/statics.lua
```diff
@@ -0,0 +1,844 @@
...
+
+function set_fg_fit(new_fit)
+    fg_fit = new_fit
+    foreground_bar:fit(new_fit)
+end
+
+function set_fg_width(new_width)
+    fg_width = new_width
+    foreground_bar:size(new_width, get_fg_height())
+    if frame_count%15 == 0 then update_strings() end
```
z16: I believe this should be done with a timer instead of a frame counter. If someone is playing with uncapped frame rate, this may still update seven times a second. If someone is lagging it may only update once a second or only every few seconds. If you did this with a timer (i.e. update if half a second has passed since the last time you did it) it would be more consistent:

``` lua
do
    local last_update = 0

    function set_fg_width(new_width)
        fg_width = new_width
        foreground_bar:size(new_width, get_fg_height())

        local now = os.clock()
        if now - last_update > 0.5 then
            update_strings()
            last_update = now
        end
    end
end
```

Edit:
I wrote this before the comment above. If you remove all of these lines I would put this directly into the `prerender` event. As a matter of fact, I would put it there regardless.

addons/barfiller/statics.lua
```diff
@@ -0,0 +1,844 @@
...
+    rested_bonus:hide()
+    box:clear()
+    ready = false
+end
+
+function mog_house()
+    local mh = windower.ffxi.get_info().mog_house
+    if mh then rested_bonus:show() else rested_bonus:hide() end
+end
+
```
z16: Do you need any of the following functions? The benefit of the library was so you could get rid off all of these. The entire rest of the file could be removed, if I understand this correctly.

addons/libs/images.lua
```diff
@@ -0,0 +1,344 @@
+--[[
+    A library to facilitate image primitive creation and manipulation.
+]]
+
+local images = {}
+local meta = {}
+
+windower.prim.saved_images = {}
+local dragged
```
z16, line 9: This variable isn't currently used. The texts library uses it to expose mouse events on the text boxes. I think it's worth having in this library as well.

addons/libs/images.lua
```diff
@@ -0,0 +1,344 @@
...
+
+function images.pos(t, x, y)
+    local m = meta[t]
+    if not x then
+        return m.settings.Pos.X, m.settings.Pos.Y
+    end
+
+    local settings = windower.get_windower_settings()
+    windower.prim.set_position(m.name, x, y)
+    m.settings.Pos.X = x
```
z16: I'm almost certain this won't work, because Windower settings are not the same as addon settings, and they are all _snake_cased_.

addons/barfiller/statics.lua
Morath86: Thanks, I'll jump on this tonight.

addons/barfiller/statics.lua
Morath86: As soon as I update the other functions that utilized them, I will remove the getters and setters.

addons/libs/images.lua
Morath86, line 9: Okay I'll look into putting it back. I personally hated always missclicking the screen, and dragging the Text fields on the pointwatch addon, but it is a lot better easier than positioning with POS commands.

addons/libs/images.lua
Morath86: Woops, that would be a copy and paste fumble. They do work as it is, but I'll fix it for follow proper syntax.

addons/libs/images.lua
z16, line 9: You can disable dragging on a per-addon basis. There is a `draggable` attribute that can be passed to it as `true` or `false`, which defaults to `true`:

``` lua
text = texts.new('Some text', { draggable = false })
```

The TParty addon does this, for example, because it explicitly positions the text boxes. If you copied the API from the `texts.lua`, this functionality would be included in the `images.lua` library, so you could make _BarFiller_ use the custom X and Y coordinates and disable it, but other addons could still use it.

addons/libs/images.lua
Morath86: Done.

Morath86:
Fixed memory leak when calculating new width, renamed functions to be read easier, strings are now
updated by a factor of time instead of frames for a more consistent experience at any frame rate.

PR #1281, opened 2015-05-16
FindAll mog safe 2 support.

xurion:
Fix to allow findall to find items in mog safe 2.


xurion:
Mistake. Ignore request.

z16:
I was already working on fixing FindAll and refactor it a bit in the process, since the way it currently works will just break again the next time something like this changes. You don't need to submit a pull request for it right now.

xurion:
Ahh ok. I did do another request, #42. Feel free to remove it if there is a larger refactor job underway. Thanks for the speedy response.

PR #1284, opened 2015-05-17
Bump up verbose and properly respect bag_priority

kelhor:
bag priority wasn't being respected. No one ever noticed because the settings used to always get blown away.

Also, this makes verbose a little more useful.


kelhor:
Thanks!

PR #1287, opened 2015-05-17
Enforce the pull/push lists at the core move function, and add ignores.

kelhor:


Byrth:
Did you mean to leave the Warp Ring in the default settings.ignore table?

holophrastic:
Yeah, since it's an item everyone has, and shows an example.

z16:
Just fyi, I don't think this is a good idea. Many people get tripped up by it and wonder why it's doing something they didn't tell them to do. We get such complaints all the time about other addons that include default settings the users don't know about and/or agree with. This is something that belongs either in a separate example file or in the addon's ReadMe.

PR #1298, opened 2015-05-19
Add organizer_items, and fix a typo.

kelhor:
Byrth: Up to you, if this is functionality you want to include, since it's still your module.

tl,dr; allows someone to define something like:

organizer_extra = {
   echos="Echo Drops",
   shihei="Shihei",
   orb="Macrocosmic Orb"
}

So they can include things for organizer without making BS sets or polluting gs validate.

### DIFF
--- a/addons/GearSwap/libs/organizer-lib.lua
+++ b/addons/GearSwap/libs/organizer-lib.lua
@@ -22,7 +22,7 @@ function org.export_set()
         return
     end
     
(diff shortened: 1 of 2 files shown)


smd111:
yes please add this
also im wondering if it possible to add a way to put the items in default bags

Byrth:
Document it somewhere, change the name to organizer_items, and you have yourself a deal!

Thanks!

kelhor:
Renamed.

PR #1311, opened 2015-05-30
[gearswap]added extdata to  player.<bag> info

smd111:
i added extdata to the player.<bag> item info as this data is useful in many ways even from inside gearswap user files

example(to get the time(server aka GMT) when you can use an item again):

```
function get_item_next_use(name)--returns time remaining till items next use
    if player.wardrobe[name] then
        return player.wardrobe[name].extdata.next_use_time
    elseif player.inventory[name] then
        return player.inventory[name].extdata.next_use_time
    end
end
```

im hoping that this is accepted and pushed quickly(both live and dev)

### DIFF
--- a/addons/GearSwap/refresh.lua
+++ b/addons/GearSwap/refresh.lua
@@ -584,7 +584,7 @@ function refresh_item_list(itemlist)
             -- If we don't already have the primary item name in the table, add it.
             if res.items[v.id] and res.items[v.id][language] and not retarr[res.items[v.id][language]] then
                 -- We add the entry as a sub-table containing the id and count
-                retarr[res.items[v.id][language]] = {id=v.id, count=v.count, shortname=res.items[v.id][language]:lower()}
+                retarr[res.items[v.id][language]] = {id=v.id, count=v.count, extdata = extdata.decode(v), shortname=res.items[v.id][language]:lower()}
                 -- If a long version of the name exists, and is different from the short version,
                 -- add the long name to the info table and point the long name's key at that table.
                 if res.items[v.id][language..'_log'] and res.items[v.id][language..'_log']:lower() ~= res.items[v.id][language]:lower() then
--- a/addons/GearSwap/refresh.lua
+++ b/addons/GearSwap/refresh.lua
@@ -126,7 +126,7 @@ function load_user_files(job_id,user_file)
         loadstring=loadstring,assert=assert,error=error,pcall=pcall,io=io,dofile=dofile,
         
         debug=debug,coroutine=coroutine,setmetatable=setmetatable,getmetatable=getmetatable,
-        rawset=rawset,rawget=rawget,require=include_user,
+        rawset=rawset,rawget=rawget,require=include_user,extdata=extdata,
         
         -- Player environment things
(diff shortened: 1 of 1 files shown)


z16:
It's up to @Byrth whether or not to merge this, but I think (if I understand this correctly) that this will make it very expensive, because it will call the `extdata.decode` function for every item you own, every time the item list gets refreshed, which happens a lot. It would be better if it was evaluated delayed, i.e. don't pass a table there, but set a meta-index to call the function when the user tries to access it:

``` lua
                retarr[res.items[v.id][language]] = setmetatable({id=v.id, count=v.count, shortname=res.items[v.id][language]:lower()}, {
                    __index = function(t, k)
                        if k == 'extdata' then
                            return extdata.decode(v)
                        end
                    end
                })
```

Alternatively this could even be done for `shortname` and other parts of the table that need to be computed for every item. The drawback is that the key `extdata` will not show up when iterating the table.

smd111:
the extdata lib is already called for gear export and as far as i have seen res is more resource hungry no to mention the only way to get the extdata to be able to put in extdata.decode(v) is with packets or windower.ffxi.get_items() 
the latter is way to much of a resource hog and if done in user files causes ffxi to stall for 2-3 second and if you dont want to filly redo all the item info again then this would be the only way

be that as it may after i modified my gs i did not see any slow downs but i also did not test moving items around but i can do that tomorrow

z16:
It doesn't cost much resources at all, it costs CPU time. Being called on export is very cheap, because that is done rarely and it only has to use it on at most 16 items, but item updates are sent constantly, especially when zoning (since all your item bags are sent to you at once) and it may need to be used on up to 720 items.

It may not be very noticeable to most people, but it might be to some, and there's really no reason to do it on every item if you can only do it on demand.

smd111:
....useless

smd111:
after testing i found that what z16 said was true beyond a shadow of a doubt (using extdata.decode() on every item is to costly) so i exposed extdata and just put the un-decoded extdata in the item data

this will allow us to get the extdata with this code

```
function get_item_next_use(name)--returns time of items next use
    for _,v in ipairs({"inventory","sack","satchel","case","wardrobe"}) do
        local item = player.[v][name]
        if item then
            return extdata.decode(item)
        end
    end
end
```

with the above code you just call get_item_next_use(name) to get the tim you can use the item next (server time) and to see what the current time is use os.time(os.date("!*t", os.time()))
example:

```
if get_item_next_use("Trizek Ring").next_use_time <= os.time(os.date("!*t", os.time()))  then-- this will be true if you can use  the Trizek Ring again
```

smd111:
@Byrth 
any thought on this

Byrth:
We talked about this last night, and the issue is that the exposed player.inventory (etc.) resources are inherently ambiguous because they're keyed to a string. As such, we can't provide data specific to an individual item (basically just extdata and inventory slot ID). Going to close it due to those issues.

smd111:
you said i should try gearswap.items to cut out the lag from windower.ffxi.get_items() but there is no data in it as you can see here https://github.com/Windower/Lua/issues/1330

PR #1320, opened 2015-06-10
fix for windows xp

retroman:
With these changes Gearswap appears to function correctly on windows XP SP3 32-bit.  I know XP isn't supported anymore, but with my configuration windows7 (32 and 64 bit) has substantially lower framerates when in intense action compared to XP.

I'm not sure why these changes are needed on windows XP only and not win7, but I needed to add the to_windower_compact function directly into gearswap.lua to get the addon to load, and I had to add the nil check or a runtime error would occur.

### DIFF
--- a/addons/GearSwap/gearswap.lua
+++ b/addons/GearSwap/gearswap.lua
@@ -255,6 +255,11 @@ function disenable(tab,funct,functname,pol)
     end
 end
 
+
+function to_windower_compact(str)
+    return __raw.lower(str:gsub(' ',''))
+end
+
 windower.register_event('incoming chunk',function(id,data,modified,injected,blocked)
     windower.debug('incoming chunk '..id)
     
@@ -356,7 +361,7 @@ windower.register_event('incoming chunk',function(id,data,modified,injected,bloc
             end
             encumbrance_table[slot_id] = tf
         end
-        if table.length(tab) > 0 then
+        if tab and table.length(tab) > 0 then
             refresh_globals()
             equip_sets('equip_command',nil,tab)
         end


z16:
This is something that should definitely not happen, and I would really like to avoid forcing it in there. Can you try to figure out _why_ exactly it isn't working? Especially the latter makes no sense whatsoever. The former could be easily explained by you not having an updated addon directory. Can you check if your `Windower/addons/GearSwap/helper_functions.lua` file has the above function? If not, your files are out of date and you need to replace them.

retroman:
I did a fresh install of windower before anything else, and I did check the helper_functions to verify the function exists.  I agree that it makes no sense, but unfortunately I'm not very good with lua so I have no idea how to troubleshoot this.  For the first problem maybe tab isn't getting anything added to it?  I guess I'll just add some print statements to see what's going on.

For the second problem,  maybe it could be due to how __raw is defined in gearswap.lua but not in helper_functions.lua, so if the loader decides to start parsing the files in parallel in win7 it will be able to properly construct the _compact function.  If loading is serialized on winxp, the required includes may be loaded before __raw is defined, which would make _compact be invaild.  This seems to be a very obscure possibility (wild-assed guess).
Maybe there is some kind of type incompatibility that can't be resolved across different units? Is there some sort of instrumented loader that can be used?  Could even be some kind of windows bug causing it.

z16:
Sorry, this has been ignored for some time, but like I said, the fixes don't seem like they relate to XP at all. Is it still not working right on XP with the stock GS? If so, I don't mind adjusting the code so that it works, as long as it doesn't impair GS in other ways, because XP is not officially supported by Windower, and hasn't been for a while. I know it's been a while, but would be cool if you could drop a line letting us know if the problems still exist.

z16:
I'm going to close this in the meantime, since the PR is not compatible with the current state of GS anyway. Feel free to reopen it with changes if you find out what needs to be done.

PR #1323, opened 2015-06-24
Added setting for combining pets from same player

geno3302:
I added a setting (combinepets) to combine all the pets from one person.

Example:
Pre-change

Shiva (Gen.) 1000
Ifrit (Gen.) 1000
Shiva(Soa.) 1000

After change

Pets (Gen.) 2000
Pets (Soa.) 1000


z16:
Sorry this has been open for so long, I never went through our PR backlog and it seems to have slipped my mind :(

Issue #1332, opened 2015-07-06
Scoreboard Request: Add Adventuring Fellow


finnigantime:
Adventuring Fellows do not currently appear in scoreboard. I want to see my fellow's DPS.

PR #1333, opened 2015-07-06
Scoreboard - add adventuring fellow

finnigantime:
Fix issue #1332 

Note the windower API only seems to allow getting the ID for the current player's fellow, so I only added the player's fellow to scoreboard. Not sure if there's a way to get the fellow ID of party members / allies.

### DIFF
--- a/addons/scoreboard/readme.txt
+++ b/addons/scoreboard/readme.txt
@@ -98,6 +98,7 @@ additional configuration options:
 * sbcolor - Color of scoreboard's chat log output
 * showallidps - Set to true to display the alliance DPS, false otherwise.
 * resetfilters - Set to true if you want filters reset when you "//sb reset", false otherwise.
+* showfellow - Set to true to display your adventuring fellow's DPS, false otherwise.
  
 Caveats:
 * DPS is an approximation, although I tested it manually and found it to
--- a/addons/scoreboard/scoreboard.lua
+++ b/addons/scoreboard/scoreboard.lua
@@ -27,6 +27,7 @@ default_settings.sbcolor = 204
 default_settings.showallidps = true
 default_settings.resetfilters = true
 default_settings.visible = true
+default_settings.showfellow = true
 default_settings.UpdateFrequency = 0.5
 
 default_settings.display = {}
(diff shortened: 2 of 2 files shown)


alphaONE2:
@finnigantime: It is possible to find any player's fellow but it requires iterating through portions of the mob array. I'm also not sure if we expose the necessary fields of the mob struct, but if not it can certainly be added.

P.S.: You don't _have_ to open an issue before submitting a pull request. I know some projects require it, but we're not nearly organized enough over here. :)

finnigantime:
Thanks! Good to know about finding other player's fellows. Up to you if this can be merged in to add the current player's fellow to scoreboard of if we should wait for a PR that includes allies' fellows also. I may try that on a second account but it would take some time before I am set up to test it.

alphaONE2:
I just checked, and it looks like we don't expose the required information to find another player's fellow at this time. If you would like it I suggest you open a feature request for it to be added to LuaCore (the main Windower/Issues tracker is the right place for this). It's been a while since I worked on this, but I _believe_ the field you need is owner_id.

As for this PR, it's fine as is. Though I suggest this be made optional. I'm sure someone out there will get upset about the change. Because change.

Other than that, everything seems fine. If you don't have time to make it optional now I can merge it as is.

finnigantime:
I am happy to make it optional. What's the right way to do that?

Filed issue https://github.com/Windower/Issues/issues/748

alphaONE2:
Add a new value to the default_settings table (created on line 24). Then check if settings['new_setting_name'] is set before adding fellows to the allies table. And finally, add a new command to change the setting around line 140.

@z16: Am I forgetting anything important?

z16:
Don't think so, sounds all fine. But for the record, we do expose a person's fellow via index. So this should work:

``` lua
print(windower.ffxi.get_mob_by_index(windower.ffxi.get_mob_by_target('t').fellow_index).name)
```

alphaONE2:
Ah good to know... I don't think that's documented right now, though.

joshk6656:
I'm pretty sure you can pass the following to ffxi.get_mob_by_target, although i do not know what some of them do. 
me
t
st
lastst
r
ft
bt
ht
pet
scan
p# (# being the party number of person)
a## (# being the party number of person)

This should allow you to get to party / alliance fellows.

z16:
They all do the same thing `<x>` does when entered in an in-game command, with the exception of `ht`. `ht` will also retrieve a CfH'd target, but it may not necessarily be the same as the in-game one.

Although I think Iryoku meant that `fellow_index` is not documented. Someone should probably change that. Not it!

finnigantime:
Setting added.

Docs updated too: http://dev.windower.net/doku.php?id=lua:api:functions:ffxi:start#windowerffxiget_mob_by_id_id
(assuming this is the only place we need to update)

So is settings.xml loaded by Windower? How does it play with `settings = config.load(default_settings)` (who trumps)?

z16:
`config.load` is precisely what loads the XML. It gets passed a default settings table in case the XML a) doesn't exist or b) is missing some values, in both cases it repopulates the XML. The user should only ever touch the settings XML, not the code itself.

PR #1334, opened 2015-07-09
stopwatch addon

finnigantime:
Proposing to add a new stopwatch addon I created.

### DIFF
--- a/addons/stopwatch/stopwatch.lua
+++ b/addons/stopwatch/stopwatch.lua
@@ -6,6 +6,7 @@ _addon.commands = {'sw', 'stopwatch'}
 require('logger')
 config = require('config')
 texts = require('texts')
+timeit = require('timeit')
  
 defaults = {}
 defaults.pos = {}
@@ -17,10 +18,9 @@ defaults.text.size = 12
  
 settings = config.load(defaults)
 times = texts.new(settings)
-startTime = os.time()
-isActive = false
-cumulativeSeconds = 0
-
+is_active = false
+cumulative_time = 0
+timer = timeit.new()
 
 windower.register_event('prerender', function()
     local info = windower.ffxi.get_info()
@@ -29,28 +29,30 @@ windower.register_event('prerender', function()
         return
     end
 
-    seconds = cumulativeSeconds
-    if isActive == true then
-        seconds = seconds + os.time() - startTime
+    total_time = cumulative_time
+    if is_active == true then
+        total_time = total_time + timer:check()
     end
-    times:text(os.date('!%H:%M:%S', seconds))
+    times:text(string.format('%0.2d:%0.2d:%0.2d', math.floor(total_time / 3600) % 100, math.floor(total_time / 60) % 60, math.floor(total_time) % 60))
     times:visible(true)
 end)
 
-windower.register_event('addon command', function(...)
-    local param = L{...}
-    local command = param[1]
-    command = command:lower() or 'help'
+windower.register_event('addon command', function(command)
+    command = command:lower()
 
     if command == 'start' then
-        isActive = true
-        startTime = os.time()
+        if is_active == false then
+            is_active = true
+            timer:start()
+        end
     elseif command == 'stop' then
-        isActive = false
-        cumulativeSeconds = cumulativeSeconds + os.time() - startTime
+        if is_active == true then
+            is_active = false
+            cumulative_time = cumulative_time + timer:stop()
+        end
     elseif command == 'reset' then
-        cumulativeSeconds = 0
-        startTime = os.time()
+        cumulative_time = 0
+        timer:next()
     else
         log("'sw start' to start the stopwatch" )
         log("'sw stop' to stop the stopwatch")
@@ -59,7 +61,7 @@ windower.register_event('addon command', function(...)
 end)
 
 --[[
-Copyright (c) 2015, Patrick Finnigan
+Copyright © 2015, Patrick Finnigan
 All rights reserved.
 
 Redistribution and use in source and binary forms, with or without
@@ -70,14 +72,14 @@ modification, are permitted provided that the following conditions are met:
     * Redistributions in binary form must reproduce the above copyright
       notice, this list of conditions and the following disclaimer in the
       documentation and/or other materials provided with the distribution.
-    * Neither the name of zonetimer nor the
+    * Neither the name of stopwatch nor the
       names of its contributors may be used to endorse or promote products
       derived from this software without specific prior written permission.
 
 THIS SOFTWARE IS PROVIDED BY THE COPYRIGHT HOLDERS AND CONTRIBUTORS "AS IS" AND
 ANY EXPRESS OR IMPLIED WARRANTIES, INCLUDING, BUT NOT LIMITED TO, THE IMPLIED
 WARRANTIES OF MERCHANTABILITY AND FITNESS FOR A PARTICULAR PURPOSE ARE
-DISCLAIMED. IN NO EVENT SHALL IHINA BE LIABLE FOR ANY
+DISCLAIMED. IN NO EVENT SHALL Patrick Finnigan BE LIABLE FOR ANY
 DIRECT, INDIRECT, INCIDENTAL, SPECIAL, EXEMPLARY, OR CONSEQUENTIAL DAMAGES
 (INCLUDING, BUT NOT LIMITED TO, PROCUREMENT OF SUBSTITUTE GOODS OR SERVICES;
 LOSS OF USE, DATA, OR PROFITS; OR BUSINESS INTERRUPTION) HOWEVER CAUSED AND
--- a/addons/stopwatch/stopwatch.lua
+++ b/addons/stopwatch/stopwatch.lua
@@ -1,0 +1,87 @@
+_addon.name = 'stopwatch'
+_addon.author = 'Patrick Finnigan (Puhfyn@Ragnarok)'
+_addon.version = '1.0'
+_addon.commands = {'sw', 'stopwatch'}
+
+require('logger')
+config = require('config')
+texts = require('texts')
+ 
+defaults = {}
+defaults.pos = {}
+defaults.pos.x = 450
+defaults.pos.y = 0
+defaults.text = {}
+defaults.text.font = 'Arial'
+defaults.text.size = 12
+ 
+settings = config.load(defaults)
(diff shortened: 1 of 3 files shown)


addons/stopwatch/stopwatch.lua
```diff
@@ -0,0 +1,87 @@
...
+windower.register_event('prerender', function()
+    local info = windower.ffxi.get_info()
+    if not info.logged_in then
+        times:hide()
+        return
+    end
+
+    seconds = cumulativeSeconds
+    if isActive == true then
+        seconds = seconds + os.time() - startTime
```
alphaONE2: &lt;pedant&gt;
This should be `seconds = seconds + os.difftime(os.time(), startTime)`

addons/stopwatch/stopwatch.lua
```diff
@@ -0,0 +1,87 @@
...
+    local param = L{...}
+    local command = param[1]
+    command = command:lower() or 'help'
+
+    if command == 'start' then
+        isActive = true
+        startTime = os.time()
+    elseif command == 'stop' then
+        isActive = false
+        cumulativeSeconds = cumulativeSeconds + os.time() - startTime
```
alphaONE2: And ditto here, this should be `cumulativeSeconds = cumulativeSeconds + os.difftime(os.time(), startTime)`
&lt;/pedant&gt;

addons/stopwatch/stopwatch.lua
```diff
@@ -0,0 +1,87 @@
...
+    if not info.logged_in then
+        times:hide()
+        return
+    end
+
+    seconds = cumulativeSeconds
+    if isActive == true then
+        seconds = seconds + os.time() - startTime
+    end
+    times:text(os.date('!%H:%M:%S', seconds))
```
alphaONE2: `os.date()` has some funny corner cases, and you should probably avoid using it here, since it's not really intended to be used for this purpose. The most obvious one in this case is that the displayed time will roll over to 00:00:00 if the stopwatch is ever left running for more than 24 hours, but there are some other weird and subtle things that can happen on Windows. Instead, I suggest you just do the calculation manually, as it's not too difficult.

``` lua
times:text(string.format('%0.2d:%0.2d:%0.2d', math.floor(seconds / 3600),
    math.floor(seconds / 60) % 60, math.floor(seconds) % 60))
```

This is also probably a little bit faster, though that's of minimal importance here.

alphaONE2:
One other thing you should be aware of is that os.time() is _not_ guaranteed to be monotonic or stable; that is, the value can jump forward or backward unexpectedly. On Windows this happens when Daylight Saving Time/Summer Time starts or ends, the value will jump one hour. Another thing that can cause unexpected jumps is when the clock is changed, either manually or automatically when Windows synchronizes with an NTP server online.

I'm honestly not sure if there's any simple way to deal with this in vanilla Lua, or with the Windower Addon API, but I wanted you to be aware that this problem exists.

trv6:
Is using the prerender event preferable to coroutine.sleep?

```
while true do
    coroutine.sleep(1)
    seconds = seconds + 1
end
```

z16:
That is not very accurate either. `coroutine.sleep` will usually sleep a bit more than the specified time (in the milliseconds range). Not very noticeable usually, but not fit for a timer. You should use the pre-render event and check the current time instead.

addons/stopwatch/stopwatch.lua
```diff
@@ -0,0 +1,87 @@
...
+    times:visible(true)
+end)
+
+windower.register_event('addon command', function(...)
+    local param = L{...}
+    local command = param[1]
```
z16: You should probably change the function signature to `function(command, ...)`, then you can avoid this assignment and it's clearly defined what the command is what the rest of the parameter.

addons/stopwatch/stopwatch.lua
```diff
@@ -0,0 +1,87 @@
...
+        startTime = os.time()
+    else
+        log("'sw start' to start the stopwatch" )
+        log("'sw stop' to stop the stopwatch")
+        log("'sw reset' to reset the stopwatch cumulative time to 0")
+    end
+end)
+
+--[[
+Copyright (c) 2015, Patrick Finnigan
```
z16: Change the `(c)`to `©` please.

addons/stopwatch/stopwatch.lua
```diff
@@ -0,0 +1,87 @@
...
+
+Redistribution and use in source and binary forms, with or without
+modification, are permitted provided that the following conditions are met:
+
+    * Redistributions of source code must retain the above copyright
+      notice, this list of conditions and the following disclaimer.
+    * Redistributions in binary form must reproduce the above copyright
+      notice, this list of conditions and the following disclaimer in the
+      documentation and/or other materials provided with the distribution.
+    * Neither the name of zonetimer nor the
```
z16: This is not `zonetimer`:)

addons/stopwatch/stopwatch.lua
```diff
@@ -0,0 +1,87 @@
...
+      names of its contributors may be used to endorse or promote products
+      derived from this software without specific prior written permission.
+
+THIS SOFTWARE IS PROVIDED BY THE COPYRIGHT HOLDERS AND CONTRIBUTORS "AS IS" AND
+ANY EXPRESS OR IMPLIED WARRANTIES, INCLUDING, BUT NOT LIMITED TO, THE IMPLIED
+WARRANTIES OF MERCHANTABILITY AND FITNESS FOR A PARTICULAR PURPOSE ARE
+DISCLAIMED. IN NO EVENT SHALL IHINA BE LIABLE FOR ANY
```
z16: I can't be entirely sure, but I suspect you're not Ihina either :)

addons/stopwatch/stopwatch.lua
```diff
@@ -0,0 +1,87 @@
...
+defaults.pos = {}
+defaults.pos.x = 450
+defaults.pos.y = 0
+defaults.text = {}
+defaults.text.font = 'Arial'
+defaults.text.size = 12
+ 
+settings = config.load(defaults)
+times = texts.new(settings)
+startTime = os.time()
```
z16: While not mandatory, it's a convention to use `snake_case`for Windower addon code. This is up to you though, just letting you know.

addons/stopwatch/stopwatch.lua
alphaONE2: And this is why @z16 does code reviews. :)

smd111:
couldn't you just use the timeit lib??
something like this

```
...
timeit = require('timeit')
...
    times:text(os.date('!%H:%M:%S', timeit.check(ctime)))
    times:show()
...
    if command == 'start' then
        isActive = true
        timeit.start(ctime)
    elseif command == 'stop' then
        isActive = false
        timeit.stop(ctime)
    elseif command == 'reset' and isActive then
        timeit.next(ctime)
...
```

this is just a thought

addons/stopwatch/stopwatch.lua
finnigantime: Respect

addons/stopwatch/stopwatch.lua
finnigantime: Nope ;)

finnigantime:
Thanks for the comments. I'm new to Lua and this project so I appreciate the thorough CR. I'll try to patch this up tomorrow.

@smd111 I missed the timeit lib functions. Looks like it uses `os.clock()` which counts system uptime and isn't susceptible to the vagaries of the Roman calendar like `os.time()`. Maybe zonetimer should be switched over as well (probably as a separate PR).

alphaONE2:
`os.clock()` is probably the best solution you have right now. However, you should be aware of a very important caveat regarding its use, as it's not really appropriate either. It is _supposed_ to return the elapsed CPU time for the current process (that is, time when the process is actually running, not idle or preempted by another process). It is implemented as a call to the C standard function `clock()`. The version of the Microsoft's Visual C++ Runtime we link against incorrectly implements this function to return elapsed wall time instead of CPU time. We can guarantee that it will continue to do so in all current and future versions of Windower 4, but this guarantee may not hold in Windower 5 if Microsoft ever fixes their implementation of `clock()`.

`clock()` also overflows every 24 days, but I don't suspect that will be much of a problem.

TL;DR: Go ahead and use `os.clock()`, but this addon may unexpectedly break in Windower 5, at which time another solution will need to be devised.

addons/stopwatch/stopwatch.lua
finnigantime: Good to know. Is this because os.difftime is more cross platform? Looks like subtraction is equivalent on Windows: http://www.lua.org/manual/5.1/manual.html#pdf-os.difftime

addons/stopwatch/stopwatch.lua
finnigantime: Good point about rollover, but this is still susceptible to that problem e.g. when hours surpasses 99. I'll add % 100 to the hours term and use the approach you suggest for now. I'm thinking about increasing precision and perhaps allowing rollover past 24 hours.

finnigantime:
Comments should be addressed. Thanks for the heads up on clock. Since I've changed the code to use the timeit lib, now only the central lib will need to be updated if Microsoft revs the CRT and fixes the bug. Lua doesn't have an `os.walltime()` method it seems? Too bad.

addons/stopwatch/stopwatch.lua
finnigantime: I avoided using `os.difftime()` here since I'm now using `os.clock()` which returns seconds.

addons/stopwatch/stopwatch.lua
alphaONE2: Nope when it goes over 99 hours it will just start displaying 100, and so on. %0.2d just means that it must be _at least_ 2 digits long, it can be longer.

addons/stopwatch/stopwatch.lua
alphaONE2: Works for me.

P.S.: Yes it is equivalent on Windows, yes it is more cross platform, and no it's not really completely necessary (hence the &lt;pedant&gt;...&lt;/pedant&gt;), but you might as well do things right if the language provides such a mechanism.

addons/stopwatch/stopwatch.lua
finnigantime: Ah, thank you for correcting me. Fixed.

PR #1344, opened 2015-08-17
Display party buffs

trv6:
This would add 18 extracted status icons. Is that okay?


addons/Nostrum/prims.lua
```diff
@@ -0,0 +1,433 @@
...
+        for i = 1, meta[t].events[event].n do
+            if meta[t].events[event][i] == n then
+                meta[t].events[event][i] = nil
+                return
+            end
+        end
+    end
+end
+
+return prims
```
z16, line 433: Any reason why you copied this file instead of using the one in the libs? Are they different?

trv6, line 433: Oh, I see. I wasn't aware of the images library. Prims.lua was a leftover from an abandoned, half-finished widgets library that I'd started. I'll probably swap the file out.

z16, line 433: Oh I didn't realize they were different, they looked very similar superficially. I'll merge this now, can swap it out (and update the existing library if you feel like it) at another point.

PR #1346, opened 2015-08-28
enemybar

mjmckee:
enemybar is an add-on that displays the current target in the top center
of the screen for easy viewing.
This is the initial version.


addons/enemybar/enemybar.lua
```diff
@@ -0,0 +1,107 @@
...
+THIS SOFTWARE IS PROVIDED BY THE COPYRIGHT HOLDERS AND CONTRIBUTORS "AS IS" AND
+ANY EXPRESS OR IMPLIED WARRANTIES, INCLUDING, BUT NOT LIMITED TO, THE IMPLIED
+WARRANTIES OF MERCHANTABILITY AND FITNESS FOR A PARTICULAR PURPOSE ARE
+DISCLAIMED. IN NO EVENT SHALL Mike McKee BE LIABLE FOR ANY
+DIRECT, INDIRECT, INCIDENTAL, SPECIAL, EXEMPLARY, OR CONSEQUENTIAL DAMAGES
+(INCLUDING, BUT NOT LIMITED TO, PROCUREMENT OF SUBSTITUTE GOODS OR SERVICES;
+LOSS OF USE, DATA, OR PROFITS; OR BUSINESS INTERRUPTION) HOWEVER CAUSED AND
+ON ANY THEORY OF LIABILITY, WHETHER IN CONTRACT, STRICT LIABILITY, OR TORT
+(INCLUDING NEGLIGENCE OR OTHERWISE) ARISING IN ANY WAY OUT OF THE USE OF THIS
+SOFTWARE, EVEN IF ADVISED OF THE POSSIBILITY OF SUCH DAMAGE.
```
z16, line 23: This part needs to be commented out, or the addon will be unable to load.

addons/enemybar/enemybar.lua
```diff
@@ -0,0 +1,107 @@
...
+_addon.name = 'enemybar'
+_addon.author = 'mmckee'
+_addon.version = '0.2'
+_addon.language = 'English'
+
+
+require('gui_settings')
+
+windower.register_event('load', function()
+	if windower.ffxi.get_info().logged_in then
```
z16: Change the file to indent one level with 4 spaces, instead of 1 tab.

alphaONE2:
Also need to add an entry to addons.xml

addons/enemybar/gui_settings.lua
```diff
@@ -0,0 +1,82 @@
...
+SOFTWARE, EVEN IF ADVISED OF THE POSSIBILITY OF SUCH DAMAGE.
+
+config = require('config')
+file = require('files')
+packets = require('packets')
+images = require('images')
+texts = require('texts')
+
+defaults = {}
+defaults.global = {}
```
z16: You should not add a `global` sub group to settings. This will be generated automatically.

addons/enemybar/gui_settings.lua
```diff
@@ -0,0 +1,82 @@
...
+THIS SOFTWARE IS PROVIDED BY THE COPYRIGHT HOLDERS AND CONTRIBUTORS "AS IS" AND
+ANY EXPRESS OR IMPLIED WARRANTIES, INCLUDING, BUT NOT LIMITED TO, THE IMPLIED
+WARRANTIES OF MERCHANTABILITY AND FITNESS FOR A PARTICULAR PURPOSE ARE
+DISCLAIMED. IN NO EVENT SHALL Mike McKee BE LIABLE FOR ANY
+DIRECT, INDIRECT, INCIDENTAL, SPECIAL, EXEMPLARY, OR CONSEQUENTIAL DAMAGES
+(INCLUDING, BUT NOT LIMITED TO, PROCUREMENT OF SUBSTITUTE GOODS OR SERVICES;
+LOSS OF USE, DATA, OR PROFITS; OR BUSINESS INTERRUPTION) HOWEVER CAUSED AND
+ON ANY THEORY OF LIABILITY, WHETHER IN CONTRACT, STRICT LIABILITY, OR TORT
+(INCLUDING NEGLIGENCE OR OTHERWISE) ARISING IN ANY WAY OUT OF THE USE OF THIS
+SOFTWARE, EVEN IF ADVISED OF THE POSSIBILITY OF SUCH DAMAGE.
```
z16, line 23: Same as in the other file, this needs to be commented out.

addons/enemybar/gui_settings.lua
```diff
@@ -0,0 +1,82 @@
...
+defaults.global.Height = 10
+defaults.global.X = windower.get_windower_settings().x_res / 2 - defaults.global.Width / 2
+defaults.global.Y = 50
+defaults.global.BGPath = windower.addon_path.. 'bar_bg.png'
+defaults.global.FGPath = windower.addon_path.. 'bar_fg.png'
+defaults.global.visible = false
+defaults.global.font = 'Arial'
+defaults.global.txtSize = 14
+defaults.global.strkSize = 1
+defaults.global.style = 0
```
z16: While this is merely a stylistic suggestion, you should be consistent with your casing. Settings commonly use _PascalCase_. Also, there's no reason to abbreviate _stroke_ to _strk_, this will only confuse some people. `StrokeSize` is a perfectly fine name for this setting.

addons/enemybar/gui_settings.lua
```diff
@@ -0,0 +1,82 @@
...
+	txtMain:bold(true)
+	txtMain:text('Enemy Name')
+	
+	txtMain:bg_visible(false)
+	txtMain:color(255, 255, 255)
+	txtMain:alpha(255)
+	
+	txtMain:stroke_width(settings.global.strkSize)
+	txtMain:stroke_color(50, 50, 50)
+	txtMain:stroke_transparency(127)
```
z16: The `texts` library has an option to automatically generate settings for you, you can look into that. You can check out other addons like _Distance_ to see how they do it. It's really easy to set up.

addons/enemybar/gui_settings.lua
```diff
@@ -0,0 +1,82 @@
...
+	bg_image:size(settings.global.Height, settings.global.Width)
+	bg_image:repeat_xy(1, 1)
+	
+	fg_image:pos(400 + 2, 50 + 1)
+	fg_image:path(settings.global.FGPath)
+	fg_image:fit(true)
+	fg_image:size(settings.global.Height, settings.global.Width)
+	fg_image:repeat_xy(1, 1)
+	
+	txtMain:pos(400, 50)
```
z16: Another stylistic choice, but we commonly use _snake_case_ for all our code. But as above, it's up to you if you want to adhere to that, just a suggestion.

mjmckee:
I will totally fix all of that and more. I've pushed a new version anyway that fixes some things like the comments on license.

addons/enemybar/targetBar.lua
```diff
@@ -0,0 +1,143 @@
...
+			if player.in_combat == true then
+				t_text:color(255, 80, 80)
+			else
+				if target.is_npc == false then
+					t_text:color(255, 255, 255)
+				else
+					if target.claim_id == 0 then
+						t_text:color(230, 230, 138)
+					else 
+						if target.hpp == 0 then
```
z16: No need to nest these, Lua supports the `elseif` keyword.

addons/enemybar/gui_settings.lua
```diff
@@ -0,0 +1,72 @@
...
+defaults.global.subtargetBarHeight = 12
+defaults.global.subtargetBarWidth = 198
+defaults.global.X = windower.get_windower_settings().x_res / 2 - defaults.global.targetBarWidth / 2
+defaults.global.Y = 50
+
+defaults.global.bg_cap_path = windower.addon_path.. 'bg_cap.png'
+defaults.global.tbg_body_path = windower.addon_path.. 'tbg_body.png'
+defaults.global.tfg_body_path = windower.addon_path.. 'tfg_body.png'
+defaults.global.stbg_body_path = windower.addon_path.. 'stbg_body.png'
+defaults.global.stfg_body_path = windower.addon_path.. 'stfg_body.png'
```
z16: I would not define paths in the settings unless you actually want users to change them. This feels more like part of the addon than something configurable, so I would make it static.

mjmckee: I've fixed this on my local version as well as the globals deal. I'm not sure what the actual structure of settings.xml would be so I just guessed based on what one looked like.

addons/addons.xml
```diff
@@ -185,6 +185,12 @@
     <support>http://www.ffxiah.com/player/Bahamut/Krizz</support>
   </addon>
   <addon>
+    <name>Enemybar</name>
+    <description>This is an addon for Windower4 for FFXI. It creates a big health bar for the target to make it easy to see.</description>
+    <bugtracker>https://github.com/mjmckee/Lua/issues</bugtracker>
+    <support>http://www.ffxiah.com/player/Ragnarok/Pesche</support>
+  </addon>
```
alphaONE2, line 193: Missing an `<author>` tag here.

addons/enemybar/targetBar.lua
mjmckee: This code didn't work correctly anyway. I have just fixed this to not only work correctly but use elseif method to look cleaner.

addons/addons.xml
mjmckee, line 193: Okay, I'll fix this next and then push all my changes so far.

alphaONE2:
It looks like you've fixed pretty much everything here, but your indentation is all over the place. Make sure you've configured your text editor to use 4 spaces instead of tabs. Any good text editor can do this, if you're using Microsoft Notepad, stop and go get [Sublime Text](http://www.sublimetext.com/), [Atom](https://atom.io/), [Notepad++](https://notepad-plus-plus.org/), or literally [anything else](https://en.wikipedia.org/wiki/Comparison_of_text_editors). Then untabify everything; that is convert all the tabs to spaces. Most good editors can do this automatically as well.

mjmckee:
I fixed my editor (Microsoft Code) but I think it the change isn't retroactive. To be honest I'm not sure how to change it retroactively. I don't know what everyone else sees but it looks fine to me. I'll get notepad++ even though I'm not a fan of changing my tools. Anyway, I just pushed my final version (1.0)   before I make a new pull request (I'm not entirely sure if I need to, or how that works) I'll go through it all and make sure it's 4 spaces and not a full tab stop.

_Okay, after further investigation, Code DID in fact replace tab stop to 4 space BUT it was not retroactive so there were some places in old code that stayed tab. Fixed and now will do another pull request_

addons/enemybar/targetBar.lua
```diff
@@ -0,0 +1,102 @@
...
+(INCLUDING NEGLIGENCE OR OTHERWISE) ARISING IN ANY WAY OUT OF THE USE OF THIS
+SOFTWARE, EVEN IF ADVISED OF THE POSSIBILITY OF SUCH DAMAGE.
+--]]
+
+tbg_cap_l = images.new(tbg_cap_settings)
+tbg_cap_r = images.new(tbg_cap_settings)
+tbg_body = images.new(tbg_body_settings)
+tfgg_body = images.new(tfgg_body_settings)
+tfg_body = images.new(tfg_body_settings)
+t_text = texts.new('  ${name|(Name)} - HP: ${hpp|(100)}% ${debug|}', settings.textSettings, settings)
```
z16: Is that `debug` meant to remain in there?

addons/enemybar/gui_settings.lua
```diff
@@ -0,0 +1,251 @@
...
+require('subtargetBar')
+
+init_images = function()
+    init_target_images()
+    init_subtarget_images()		
+end
+
+check_claim = function(claim_id)
+    local player = windower.ffxi.get_player()
+    local party = windower.ffxi.get_party()
```
z16: You don't seem to need this variable in your code since you're using `windower.ffxi.get_mob_by_target` instead, so you might wanna remove it. `windower.ffxi.get_party` is a very expensive function call, so you should avoid it where possible. Currently you call it twice on each rendered frame.

addons/enemybar/subtargetBar.lua
```diff
@@ -0,0 +1,80 @@
...
+    stbg_body:pos(settings.pos.x + 400, settings.pos.y + 15)
+    stfg_body:pos(settings.pos.x + 400, settings.pos.y + 15)
+    st_text:pos(settings.pos.x + 400, settings.pos.y + 15)
+end
+
+render_subtarget_bar = function(...)
+    if visible == true then
+        local target = windower.ffxi.get_mob_by_target('t')
+        local subtarget = windower.ffxi.get_mob_by_target('st')
+        local player = windower.ffxi.get_player()
```
z16: While `windower.ffxi.get_player` isn't as bad as `windower.ffxi.get_party` it's still rather expensive. You might wanna place it inside the following `if` clause, because it's not needed outside of it.

What some people do, to avoid calling `windower.ffxi.get_player` on each rendered frame is to globalize it and only update it on load and login. This works if you only need things like the player name or ID, since those don't change while logged in, and in your case you only need the ID, so you avoid needlessly calling it many times.

In fact, as a future improvement you may consider caching the target and subtarget IDs as well and switch to using the `target change` event instead of `prerender` to determine those. But that would be a minor improvement and I'd only recommend it if you're out of other things to do, since you'd still need to call them on `prerender` to determine their HP percentage, but you'd only need to call one of them.

addons/enemybar/targetBar.lua
```diff
@@ -0,0 +1,102 @@
...
+        tbg_cap_l:show()
+        tbg_cap_r:show()
+        tbg_body:show()
+        tfg_body:show()
+        tfgg_body:show()
+        t_text:show()
+        
+        local target = windower.ffxi.get_mob_by_target('t')
+        local player = windower.ffxi.get_player()
+        local party = windower.ffxi.get_party()
```
z16: Same as above, `windower.ffxi.get_party` does not seem to be needed at all, so you should remove it from here.

z16:
As for the tab thing, you can use a simple text replace mechanism to replace all tabs to spaces, all editors (including VSCode and even Notepad) support. You shouldn't have to switch editors just for this, VSCode is more than enough for this.

But I wanted to write a bit on settings and how you use them. Most things inside the code are fine, and I'd be ready to pull it (the things I mentioned just now were mostly just efficiency improvements, nothing critical), but there are some things in the settings handling that I feel should be improved before we merge this.

Take this line:

``` lua
defaults.pos.x = windower.get_windower_settings().x_res / 2 - defaults.targetBarWidth / 2
```

Default values should be constants and not calculated when first run. These will be saved into the settings, and then reloaded and overwritten on the next load. Now if someone changes resolutions and logs in again, the settings will be off, and they will overwrite the default values, so the bar will be misplaced.

The key to this is to not write things that vary like that to the settings at all. The same goes for settings that depend on other settings, for the same reason, because it can result in a settings mismatch. Like these:

``` lua
stbg_cap_settings.pos.x = settings.pos.x
stbg_cap_settings.pos.y = settings.pos.y
```

In this case someone could have different settings for their characters and when they relog into the other character, the settings for `stbg_cap_settings` will not update anymore.

Only the most basic settings that all other settings depend on should be explicitly specified. Anything that is inferred on load should _only_ be inferred on load and relog of characters. The `config` library has a mechanism for that in its `register` function, and it works like this (example taken from the _Treasury_ addon):

``` lua
s = S{'pass', 'lot', 'drop'}
code = {}
code.pass = S{}
code.lot = S{}
code.drop = S{}

config.register(settings, function(settings_table)
    code.pass = settings_table.Pass:map(table.get+{ids} .. string.lower)
    code.lot = settings_table.Lot:map(table.get+{ids} .. string.lower)
    code.drop = settings_table.Drop:map(table.get+{ids} .. string.lower)
end)
```

The `code` table contains IDs which the addon should pass, lot or drop respectively. The problem is that the addon settings store the items in string form (their english names). But internally it needs the item IDs.

This snippet registers a function with the `config` library. It will be executed every time the settings are reloaded (on the `load`, `login` and `logout` events). This function sets up values based on the settings. In this case, it reads the new settings (from the new character), which is the sole argument to the function. It parses the strings from the settings file and converts them to IDs, which the addon uses internally. That way it's ensured that every time the settings are reloaded the internal state is valid.

You should adapt your code to use the same function. Make one function that initializes all text/prim settings based on the settings that are stored in the file. And settings that should not be stored (like screen resolution dependent settings, because they can vary) can be determined at this point as well instead of being written to the defaults.

Let me know if you're unclear on any of this, I'd be happy to explain in more detail. You can also hit us up on IRC, since it's easier to talk there.

mjmckee:
Okay. I'll work on all this tomorrow. Now that it is basically and functionally stable making these little improvements should be a lot easier. I really appreciate the notes they have helped a lot.

_All issues have been addressed and fixed._

addons/enemybar/targetBar.lua
mjmckee: Yes. I figured I'd leave it in for the future.

addons/enemybar/gui_settings.lua
mjmckee: Fixed. I'll remember this for the future.

addons/enemybar/subtargetBar.lua
mjmckee: I fixed the get_player issue. I moved it inside the if, and cached the id which only really benefits sub-target since I don't use the animation which relies on player.in_combat)

I think as for the second part, I'll just worry about this later.

addons/enemybar/targetBar.lua
mjmckee: Fixed.

mjmckee:
I fixed a couple bugs and would like to request a new pull, but this one is STILL? open?

addons/enemybar/enemybar.lua
```diff
@@ -0,0 +1,54 @@
...
+    end
+end)
+
+windower.register_event('prerender', render_target_bar)
+windower.register_event('prerender', render_subtarget_bar)
+windower.register_event('target change', target_change)
+
+windower.register_event('logout', function(...)
+    -- This is a super cheap fix, but it works. 
+    windower.send_command("input //lua r enemybar");
```
z16, line 53: Fix for what? Also:

``` lua
    windower.send_command('lua r enemybar')
```

z16:
Sorry, I completely forgot about this. Totally my bad, everything looks fine now (except for the thing I posted but that's not really an issue, just a little ugly).

addons/enemybar/enemybar.lua
mjmckee, line 53: Well if you logout while something is targeted (I.E. In the mog house and type /logout to switch characters.) the screen elements will persist. It's "ugly" because I didn't want to spend any more time on this addon than I already have, but still wanted to fix this issue.

PR #1355, opened 2015-09-23
Retain

MilkMansKid:
This just seemed like it belonged as default option over using custom items for it. Since we have retain slip gear already.


alphaONE2:
Going to need @kelhor to review this.

kelhor:
Solid.

PR #1358, opened 2015-09-30
boxdestroyer: updated message IDs (thanks to semiju)

svanheulen:


z16:
For the record, this could be rewritten so it's a lot easier to maintain, since the offsets between entries in a single zone are usually the same. Not sure if it's worth it, considering the final update will come soon, but just something to consider.

svanheulen:
I have a script that automatically generates the `messages.lua` from the DAT files. I just don't have the game installed any more, because I haven't played since February, so I needed someone else to run it, haha.

PR #1366, opened 2015-10-20
How to use

JovemDark:
Hello, I wonder how I make use of this language. I became interested enough, after checking the project. Sorry for my English, it is very bad rs.


nitrous24:
This is not really the place to be asking about it. However i will answer your question and close this anyway. Here is the best place to learn about the Lua API and how to use it. http://dev.windower.net/doku.php?id=lua:start   

There's some helpful links in that too. Good luck.

nitrous24:
also, for future reference, Opening a pull request is not generally a good idea. There is an Issues tracker for reporting issues. Pull requests are just for updating addons and such.

PR #1371, opened 2015-11-15
New Addon - findTrust

DamienDennehy:
New addon to simply list what trusts you're missing.

### DIFF
--- a/addons/findTrust/findTrust.lua
+++ b/addons/findTrust/findTrust.lua
@@ -9,14 +9,14 @@
 --    * Redistributions in binary form must reproduce the above copyright
 --      notice, this list of conditions and the following disclaimer in the
 --      documentation and/or other materials provided with the distribution.
---    * Neither the name of <addon name> nor the
+--    * Neither the name of FindTrust nor the
 --      names of its contributors may be used to endorse or promote products
 --      derived from this software without specific prior written permission.
 
 --THIS SOFTWARE IS PROVIDED BY THE COPYRIGHT HOLDERS AND CONTRIBUTORS "AS IS" AND
 --ANY EXPRESS OR IMPLIED WARRANTIES, INCLUDING, BUT NOT LIMITED TO, THE IMPLIED
 --WARRANTIES OF MERCHANTABILITY AND FITNESS FOR A PARTICULAR PURPOSE ARE
---DISCLAIMED. IN NO EVENT SHALL <your name> BE LIABLE FOR ANY
+--DISCLAIMED. IN NO EVENT SHALL DAMIEN DENNEHY BE LIABLE FOR ANY
 --DIRECT, INDIRECT, INCIDENTAL, SPECIAL, EXEMPLARY, OR CONSEQUENTIAL DAMAGES
 --(INCLUDING, BUT NOT LIMITED TO, PROCUREMENT OF SUBSTITUTE GOODS OR SERVICES;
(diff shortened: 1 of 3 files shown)


addons/findTrust/findTrust.lua
```diff
@@ -0,0 +1,78 @@
...
+
+--Redistribution and use in source and binary forms, with or without
+--modification, are permitted provided that the following conditions are met:
+
+--    * Redistributions of source code must retain the above copyright
+--      notice, this list of conditions and the following disclaimer.
+--    * Redistributions in binary form must reproduce the above copyright
+--      notice, this list of conditions and the following disclaimer in the
+--      documentation and/or other materials provided with the distribution.
+--    * Neither the name of <addon name> nor the
```
alphaONE2: Replace `<addon name>` with FindTrust.

addons/findTrust/findTrust.lua
```diff
@@ -0,0 +1,78 @@
...
+--      names of its contributors may be used to endorse or promote products
+--      derived from this software without specific prior written permission.
+
+--THIS SOFTWARE IS PROVIDED BY THE COPYRIGHT HOLDERS AND CONTRIBUTORS "AS IS" AND
+--ANY EXPRESS OR IMPLIED WARRANTIES, INCLUDING, BUT NOT LIMITED TO, THE IMPLIED
+--WARRANTIES OF MERCHANTABILITY AND FITNESS FOR A PARTICULAR PURPOSE ARE
+--DISCLAIMED. IN NO EVENT SHALL <your name> BE LIABLE FOR ANY
```
alphaONE2: Put your name here instead of `<your name>`.

DamienDennehy:
Done, sorry for missing it.

alphaONE2:
You also need to add an entry to [addons.xml](https://github.com/Windower/Lua/blob/dev/addons/addons.xml). Without this Launcher won't be able to find the addon. Please fill out all of the tags.

alphaONE2:
OK, everything looks good to me. I'll let @z16 do his usual code review stuff, though. He's much better at this than I am.

DamienDennehy:
Thanks, appreciate the feedback - my first addon :)

z16:
The code checks out, although quite a few things could be done simpler. You mentioned that you used the code from [this thread](http://forums.windower.net/index.php?/topic/938-blu-spell-checklist-script/) to make it. If you scroll down further in that thread you'll find a better version of the code and I would recommend you use that as your base. But that is up to you.

Since we're on it, I would actually suggest you close this pull request in turn it into something more generic by combining the two features and adapting it to check for arbitrary spell types. For example, you could use this command to find missing trusts:

```
//addonname trust
```

And this would list all missing blue magic spells:

```
//addonname bluemagic
```

And so on. It would not be too hard to adjust to that and then we wouldn't need one addon for each spell category. And people could easily check for BRD songs, Ninjutsu, etc. as well.

But this is still just a suggestion on my side, if you don't want to do that let me know, your addon as it is is ready to be pulled.

DamienDennehy:
You have a point actually. Maybe an addon called findSpell would be more generic.
Okay I'll start on that and use some of the code you mentioned in the thread.

Feel free to reject the pull request.

alphaONE2:
OK then, open a new pull request when you're ready.

z16:
If you do that, may I suggest _SpellCheck_ as the name, if only to confuse people? :D

Also please don't lowercase the first letter, it makes baby Jesus cry :< Also, me.

DamienDennehy:
lol. PascalCase it is then. SpellCheck sounds good.

PR #1378, opened 2015-11-21
SpellCheck - filtered out spells with no levels & added exception list

DamienDennehy:
Spells with no levels are now filtered out from the list of spells you have yet to learn.

Also added a manual exception list as there are several spells with levels (Hastega etc) that only an avatar or mob can use.

Tidied up and simplified some of the code.

### DIFF
--- a/addons/SpellCheck/SpellCheck.lua
+++ b/addons/SpellCheck/SpellCheck.lua
@@ -85,7 +85,7 @@ function display_spell_count(command)
         --Trust and spells must be processed separately
         if command == 'tru' then
             --Only include non Unity trusts
-            if not string.match(res.spells[spell].name, '(UC)') then
+            if not res.spells[spell].name:endswith('(UC)') then
                 missing_spells_len = missing_spells_len + 1
                 table.insert(missing_spell_names, res.spells[spell].name)
             end
--- a/addons/SpellCheck/SpellCheck.lua
+++ b/addons/SpellCheck/SpellCheck.lua
@@ -26,12 +26,14 @@
 
 _addon.name    = 'SpellCheck'
 _addon.author  = 'Zubis'
-_addon.version = '1.0.0'
+_addon.version = '1.0.1'
 _addon.command = 'SpellCheck'
 
 require('sets')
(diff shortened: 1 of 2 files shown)


addons/SpellCheck/SpellCheck.lua
```diff
@@ -67,33 +69,33 @@ end
...
+        
     --Add missing spells to table for sorting
     for spell in missing_spells:it() do
-        --if trusts are being searched, exclude Unity trusts
-        if command == "tru" then
-            if not string.match(res.spells[spell].name, "(UC)") then
+        --Trust and spells must be processed separately
+        if command == 'tru' then
+            --Only include non Unity trusts
+            if not string.match(res.spells[spell].name, '(UC)') then
```
z16: This works just fine, but personally I prefer being more specific when possible. As such I'd do:

``` lua
            if not res.spells[spell].name:endswith('(UC)') then
                -- Stuff
            end
```

z16:
As before, your solution works and there's nothing major to object to, there are just a few things I think could be done clearer. Let me know if you're fine with it as it is, then I'll merge it.

First you're relying on IDs being different for all search term. Which happens to be the case for spells, but not for other resources, if you ever want to expand the addon to include other things. For that to work you should refactor the exceptions table to be searchable by category.

Furthermore, this is again a good opportunity to use sets instead of regular tables. You want to know whether or not an element (an ID) is contained in a collection, so all you need is membership testing, which is exactly what sets are made for.

The final table would ideally look something like this:

``` lua
exceptions = {
    blu = S{
        123,
        234,
        356,
    },
    smn = S{},
    nin = S{
        456,
        567,
    },
    -- And so on
}
```

Then you'd check it like this:

``` lua
               if not table.empty(res.spells[spell].levels) and not exceptions[command]:contains(res.spells[spell].id) then 
```

DamienDennehy:
I've updated the string search, but will probably hold on splitting the exceptions for a future release, so feel free to merge when you get a chance. 

You're right though, it is the best way to do it if I was searching for anything else. Nice and neat.

PR #1391, opened 2015-12-15
Libs/Packets/Fields: Fixed 0x068

aukon:
The HP/MP fields of 0x068 are currently reversed:

example


z16:
Good catch, thanks for that.

PR #1395, opened 2015-12-15
SpellCheck - Added exception for Sleepga Blood Pacts.

DamienDennehy:
Added Sleepga I/II spell exception as the spells were listed in spells.lua twice. 
I'm guessing the second Sleepga I/II were Blood Pacts.


z16:
Actually, I may have merged it prematurely... the SMN one is a Blood Pact, hence it's listed under `job_abilities.lua`:

``` lua
-- Automatically generated file: Job Abilities

return {
    --[[ stuff ]]
    [611] = {id=611,en="Sleepga",ja="スリプガ",element=1,icon_id=346,mp_cost=54,prefix="/pet",range=4,recast_id=174,targets=32,tp_cost=0,type="BloodPactWard"},
    --[[ more stuff ]]
}
```

DamienDennehy:
Maybe the duplicate spells aren't SMN spells at all then. 

The issue was that in spells.lua Sleepga and Sleepga II are listed twice.
The spells 273, 274 seem to be the usable ones, but spells 363, 364 are the ones I excluded.

I used a mule to check it and it seems to be working fine at least.
It's no longer reporting Sleepga and Sleepga II as being unlearned spells if you already have them.

z16:
Whatever, the pull request still seems fine. They are probably mob-only variants of Sleepga. Either way, they're not player usable, as such it makes sense to exclude them.

PR #1398, opened 2015-12-25
New addon RAWR

geno3302:
RAWR plays a sound effect when one of the unity leaders mentions one of
the dragons in chat.


addons/RAWR/RAWR.lua
```diff
@@ -0,0 +1,43 @@
+_addon.name = 'RAWR'
+_addon.author = 'Genoxd'
+_addon.version = '1.0.0.0'
+_addon.commands = {'rawr'}
```
z16: You don't need to set a command if the addon does not support any commands.

addons/RAWR/RAWR.lua
```diff
@@ -0,0 +1,43 @@
...
+
+require 'lists'
```
z16: You should change this to `require('tables')`, since you don't actually use any lists in here, only the Windower-specific tables.

z16:
1. The addon is missing the license disclaimer
2. You need an entry in the `addons.xml` file for it to show up in the Windower launcher
3. The sound file should be in a different folder, the `data` sub-folder is for user-specific items only

addons/RAWR/RAWR.lua
```diff
@@ -0,0 +1,68 @@
+--Copyright (c) 2016, geno3302
```
z16: Please change that to `©`.

addons/RAWR/RAWR.lua
```diff
@@ -0,0 +1,68 @@
...
+
+--Redistribution and use in source and binary forms, with or without
+--modification, are permitted provided that the following conditions are met:
+
+--    * Redistributions of source code must retain the above copyright
+--      notice, this list of conditions and the following disclaimer.
+--    * Redistributions in binary form must reproduce the above copyright
+--      notice, this list of conditions and the following disclaimer in the
+--      documentation and/or other materials provided with the distribution.
+--    * Neither the name of <addon name> nor the
```
z16: Enter your addon name here :)

PR #1401, opened 2016-01-04
organizer now prefers putting gear in wardrobe over inventory

zapatoverde:
@kelhor 
@Byrth 

If an item can be stored in the wardrobe, that is, it is equippable (it
does not matter whether the item was mentioned in the sets or the
organizer_items variable), it is now added to the wardrobe. All
non-equippable items are added to the inventory directly. If items
stored in the wardrobe pass the wardrobe max limit, the surplus items
are moved into the inventory.  If the inventory max limit is passed
after surplus items from wardrobe has been moved into inventory, an
error message is given. The reason for all of this logic is to use the
wardrobe in a more effective manner, while leaving more space for loot
in the inventory. In order to facilitate this a new tidy function (tidy_wardrobe) has been introduced that moves "unwanted" gear from wardrobe into the inventory. Using the existing tidy function (renamed to tidy_inventory) will now move "unwanted" items from inventory (whether they orginally came from tidying the wardrobe or not) into the dump bags.


kelhor:
This is a sizable change in behavior from what's been the standard all along. My inclination is that it should be an option in the settings (even if it's the default).

zapatoverde:
I see your point, only I don't think it's all about new features here. Half of this is about bugfixes.

Consider the gearswap part for instance. There are what I believe to be 3 bugs in the section I've wiped clean:
- wardrobe max mistakingly hardcoded to 80 (see the slice on line 85)
- surpluss inventory items are copied to wardrobe, but never removed from the inventory (also line 85)
- surpluss inventory items may not be equippable, but is still moved into the wardrobe (also line 85)

Also, similar to what is now, the gearswap part did in fact distribute gear across inventory and wardrobe earlier, but as you no doubt have pointed out, it preferred inventory over wardrobe earlier (I do agree we prolly should make a setting for this, allthough I don't see why people would like that priority (since the inventory is where all the loot drops and where it would benefit the most to have free space)).

Now, let's consider the organizer part. Since the gearswap part distributes gear over inventory and wardrobe it makes sense that they have to be tidied in order to fit gear into them. However, how does the organizer work now? It only tidies the inventory, when it needs to tidy the wardrobe aswell. So I would argue that the wardrobe tidying is more of a fix than a feature.

It might be that I don't see the bigger picture here (which would not surprise me lol). If that's the case, we should probably, just as you have pointed out, make an enable_wardrobe_tidying setting.

Would be sweet if you could elaborate what you would like settings for bud :) Also, if I'm not mistaking, it is as of now possible to set wardrobe and inventory as drop bags. Is this really a smart idea considering that the gearswap part tags these spaces as where the wanted items should go? If you agree this is dangerous, should I fix this now?

nitrous24:
Would this move items out of wardrobe to storage/banks/locker just because
i don't have them set to be in my inventory at all times? (i.e. my warp
ring and other charge based items that i keep in my wardrobe at all times
as well as things like proc weapons for if i decide to go to abyssea/vw)

zapatoverde:
Yes this would move items out of wardrobe UNLESS you have them in the organizer_items set like I do, which is the point of the organizer_items set in the first place. This is the set you use when you wish to for instance have a warp ring with you at all times (independent of job if you wish), or shihei if you're on ninja, or pet jugs if you're on bst. Abyssea weapons are a waste of space if you're not going to Abyssea (why not have the organizer clear these out for you if you're not going there?). And if you want Abyssea weapons in your wardrobe at all all times whether you go to abyssea or not, whether going as nin or whm, just put them into each jobs gearswap organizer_items set (or even make a organizer_items set you reuse in all job gearswap files). If, on the other hand, you'd really like to not waste space and be lazy at the same time, you could also choose to dynamically control organizer_items set through gearswap commands (similar to switching/toggle idle/tp sets) before organizing if you're into that thing ^^ Hope I understood your question correct! I have a habit of misunderstanding ; ;

kelhor:
This is more or less exactly what I was talking about when I expressed concern about changing the default behavior. It's been standard practice for a while to treat wardrobe as more or less exempt from organizer's behavior.

I've been out of town and will look over your response in more detail after work.

smd111:
What we really need is a way to define default locations for our items
example:
a single file that lets orginizer know where the user wants each items default storage to be

```
Safe = {"Outrider Mask","Outrider Mail","Outrider Mittens","Outrider Hose","Outrider Greaves",etc.}
Storage = {"Eminent Axe","Eminent Scimitar","Eminent Dagger","Eminent Voulge",,"Eminent Sword","Em. Baghnakhs","Eminent Sickle",etc.}
Case = {usable items}
with a setting that not items but whats listed can go there (true/false)
```

zapatoverde:
@kelhor: If that's the case why would the gearswap part choose to put surplus items into the wardrobe, when the organizer part have no way of cleaning it before doing so? I haven't tried it myself, but I am quite confident this would be an issue for people needing more than 80 items on a job. Sometimes, if the wardrobe was cleaned out it would work. Other times, if the wardrobe was packed it would fail.

@smd111: Yeah I agree something along those lines would be a nice feature. I also believe the gearswap/organizer should support its current behaviour (since one of the sweet things imo of the gearswap/organizer is that it works very nice from the getgo without too much configuration). I suggest you open a feature request for this, as that is not what I'm trying to do here. I'm just trying to create a better default behaviour.

Byrth:
@smd111 that already exists in the form of default inventory files. People just don't use the functionality.

I'm opposed to this change for all the reasons outlined by everyone else.

smd111:
@Byrth if you talking about the organizer command to freeze the only way to use it is to make one for each job and a default and its cant be updated on the fly as you have to rebuild every file when you add even one new piece of gear to your inventory (i actually did try using this part of organizer)

alphaONE2:
I think this feature makes sense, and it's something a few people have asked for; it just needs to be a toggle-able option. If it is defaulted on, an announcement needs to be made detailing the change before it goes live.

Byrth:
@smd111 No, you can manually make default organizer files.

smd111:
@Byrth How ??
and how do you force them to trigger when using gs's org include??

Byrth:
You make them in a text editor. They trigger automatically if there's nothing that matches the file name for that inventory.

Byrth:
Also, I didn't mean that this merge was a terrible idea that shouldn't be implemented. I just meant that it needs to be optional.

zapatoverde:
Ok! Happy to hear that. I'll look into it when I get around to it :)

smd111:
@Byrth do you have an example of how thoes files should look?? what there file names should be and the folder layout

http://www.filedropper.com/data_8
this is the only way i can think to do it but as i said its a pain in the ass to update the files because you need re make ever one.

but what i would like to see is have it place every it in the default location if its no needed byu the current job when using the gs organizer-lib 

and i did try to just have the default files but it was not good it would still put gear in the the baggs based on the bag order

what id like to see is this

```
    for _,v in pairs(default_item_locations) do
        --remove bag bag use list
        if item_needed.contains(item name) and inventory.contains(itemname) then
            -- leave item in inventory
        elseif v.contains(itemname) then
            --put item in bag
            --remove item from move list
        end
    end
    for _,v in pairs(bag_use) do
        if not v then
            --put item in bag
        end
    end
```

the above probably is not worded right

alphaONE2:
I've been thinking about this a little bit more. The biggest issue is that in Windower 4 there's no (simple) way to communicate between two addons. So even if a "use_legacy_org_mode" or some such is added to organizer, there's no easy way to let the gearswap lib no about it. What this tells me is that some of the logic needs to be moved to the organizer side.

What I'm thinking is this. On the gearswap side, all it should do is compile a list of items to be collected, and leave the rest of the logic up to organizer. On the organizer side, it can check the setting and sort the requested items into bags as is done now or as this change would do depending on the user's settings, and then everything can proceed as before.

It shouldn't be too hard to make this change, mostly just copy/pasting some functions and minor refactoring. Or Is there's something I'm missing? an easier way to do this? Probably need @Byrth or @kelhor to chime in.

kelhor:
That's actually more or less what it already does - the GS org piece just writes out config files to be passed in to organizer as arguments. This PR is really two things; change organizer's behavior wrt the wardrobe, and let the organizer-lib in GS generate different config files. Changing that to just handing a list of desired items to organizer (and letting it sort out inv vs wardrobe/write out config files) is actually a much bigger change.

I meant to get to this this weekend, and just couldn't, sorry. I still think it has real value, both with the bug fixes and the behavior change, so long as the behavior change is not the default (as I'm the one who gets support questions about it ;)) I'll see if I can find some time in the next day or so if no one else does.

alphaONE2:
I'm not sure what the status of this is currently, but the PR is no longer mergeable, so I'm going to close it.

PR #1402, opened 2016-01-29
Update MobCompass.lua

strevorrow:
Addresses a bug where if the PC is on top of target, nil error is thrown ad infinitum.

Adds a simple added if statement checking if the positional difference is nil, which occurs most often when a PC uses a home point crystal to port to another location and targets another player right under them.

### DIFF
--- a/addons/MobCompass/MobCompass.lua
+++ b/addons/MobCompass/MobCompass.lua
@@ -223,11 +223,11 @@ do
             end
             local next45angle = math.ceil((angle+pi/8)/(pi/4))
             if next45angle ~= last45angle then
-            	if next45angle ~= nil and last45angle ~=nil then
-                arrow_map[last45angle]:color(255,255,255)
-                arrow_map[next45angle]:color(255,0,0)
-                last45angle = next45angle
-            end
+                if next45angle ~= nil and last45angle ~=nil then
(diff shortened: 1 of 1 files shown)


addons/MobCompass/MobCompass.lua
```diff
@@ -223,10 +223,12 @@ do
             end
             local next45angle = math.ceil((angle+pi/8)/(pi/4))
             if next45angle ~= last45angle then
+			   if next45angle ~= nil and last45angle ~=nil then
                 arrow_map[last45angle]:color(255,255,255)
                 arrow_map[next45angle]:color(255,0,0)
                 last45angle = next45angle
             end
```
alphaONE2: Need to fix the indentation here from lines 226 through 230. Please use 4 spaces to indent and no tabs.

addons/MobCompass/MobCompass.lua
```diff
@@ -223,10 +223,12 @@ do
...
+            end
```
alphaONE2, line 231: Need to fix the indentation here from lines 226 through 230. Please use 4 spaces to indent and no tabs.

alphaONE2:
Well that was fast... All looks good now.

PR #1404, opened 2016-02-14
DressUp: Make "party" and "others" in blinking work properly.

nekonok:
"party" and "others" in blink config table had been ignored because
blink_logic() function recieves only "follow" or "self" in 1st argument.

### DIFF
--- a/addons/DressUp/DressUp.lua
+++ b/addons/DressUp/DressUp.lua
@@ -166,7 +166,7 @@ windower.register_event('incoming chunk',function (id, data)
     end
     
     -- Model ID 0xFFFF in ranged slot signifies a monster. This prevents undesired results.
-    modified,packet = modify_gear(packet, name or _char, blink_logic(player.autorun and "follow" or "self", character, player), models)
+    modified,packet = modify_gear(packet, name or _char, blink_logic(blink_type, character, player), models)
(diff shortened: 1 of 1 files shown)


z16:
Sorry for the late reply, but I'm not sure if this is the correct solution. There's obviously something missing, but this ignores the auto-run/target lock combination, which most people use instead of the regular `/follow`. This should be accounted for in the previous `if` block. I can look into it after work tonight, if you don't do it yourself by then.

nekonok:
For simplicity, "auto-run + target lock" should be treated by "target" rule.
Actually, `/follow` and "auto-run + target lock" have different behaviors like tracing route and cancel message.
If we treat these as same, when you cancel auto-run, the followed character blinks with no system message and then the target is blurred. Don't you think this is strange and annoying?

BTW, this PR is aimed to just make "party" and "others" rule work properly.
Existing code treat `player.autorun` = `/follow`, not "auto-run + target lock".

Byrth:
He's right. I don't know how DressUp was working as it appears to generate the blink_type variable and then ignores it. Perhaps I messed it up in a previous edit.

Minor adjustment, but the above "if character then" on line 147 is unnecessary, as the function would already have returned on line 144 if character didn't exist.

Byrth:
I will change that, as well as adding some more code that I've had living in my DressUp folder for the last week. Thanks for the commit!

Issue #1410, opened 2016-03-17
GearSwap: `//gs export` should escape single quotations in augment properties


nekonok:
If a gear has augment property that contains single quote, exported lua file is not valid because single quote is not escaped.
e.g.)

``` lua
sets.exported={
    feet={ name="Plun. Poulaines +1", augments={'Enhances "Assassin's Charge" effect',}},
}
```

this should be:

``` lua
sets.exported={
    feet={ name="Plun. Poulaines +1", augments={'Enhances "Assassin\'s Charge" effect',}},
}
```

Byrth:
Thanks. I'll look into this this weekend. It's honestly kind of amazing this hasn't been reported before.

Byrth:
Fixed it, I think.

PR #1411, opened 2016-03-18
#1410 GearSwap: escape single quotes in augments when exporting lua file

nekonok:


nekonok:
fix #1410

Byrth:
There were actually 4 places this needed to be changed. I've made the adjustments and will upload it today.

PR #1414, opened 2016-03-24
Changed set spell function to do in-place set instead of clear first.

AkadenTK:
Clear-then-set would unequip all traits which would unequip the sub weapon, causing a loss in TP or just unexpected results. This won't put spells in the correct order (but that doesn't matter anyway). Tested for about a week of normal play on BLU.

### DIFF
--- a/addons/azureSets/azuresets.lua
+++ b/addons/azureSets/azuresets.lua
@@ -75,41 +75,94 @@ function set_spells(spellset)
         error(spellset..' was already equipped.')
         return
     end
-    windower.ffxi.reset_blue_magic_spells()
+    --[[windower.ffxi.reset_blue_magic_spells() ]]--
     log('Starting to set '..spellset..'.')
-    set_spells_from_spellset(spellset,1)
+    set_spells_from_spellset(spellset)
     return
 end
 
-function set_spells_from_spellset(spellset,slot)
-    local islot
-    if tonumber(slot) < 10 then 
-        islot = '0'..slot
-    else islot = slot end
-    local tempname = settings.spellsets[spellset]['slot'..islot]
-    if tempname ~= nil then
-        for spell in spells:it() do
-                if spell['english']:lower() == tempname:lower() then
-                    windower.ffxi.set_blue_magic_spell(spell['id'], tonumber(slot))
-                    break
-                end
+function set_spells_from_spellset(spellset)
+    local setToSet = settings.spellsets[spellset]
+    local currentSet = get_current_spellset()
+    local setSlot = 'none'
+    local spellToSet = 'none'
+    local slotToSetTo = 0
+    for k,v in pairs(currentSet) do
+      if not set_contains_value(setToSet, v) then
+        setSlot = k
+        slotToSetTo = tonumber(k:sub(5, k:len()))
+        --log('Found spell to remove: '..v..','..slotToSetTo)
+        break
+      end
+    end
+    if slotToSetTo == 0 then
+      --log('No spell to remove. Look for empty')
+      -- Find an empty slot
+      for i = 1, 20 do
+        local t = ''
+        if i < 10 then t = '0' end
+        local slotName = 'slot'..t..i
+        if currentSet[slotName] == nil then
+          slotToSetTo = i
+          --log('Found empty at: '..slotToSetTo)
+          break
         end
+      end
     end
-    if tonumber(slot) < 20 then
-        windower.send_command('@wait .5;lua i azuresets set_spells_from_spellset '..spellset..' '..slot+1)
+    for k,v in pairs(setToSet) do
+      if not set_contains_value(currentSet, v) then
+        spellToSet = k
+        --log('Found spell to set: '..v)
+        break
+      end
+    end
+    if setSlot ~= 'none' then
+      windower.ffxi.remove_blue_magic_spell(tonumber(slotToSetTo))
+      --log('Removed spell at: '..slotToSetTo)
+    end
+    if spellToSet ~= 'none' then
+      local spellName = setToSet[spellToSet]
+      if spellName ~= nil then
+        local spellID = find_spell_id_by_name(spellName)
+        if spellID ~= -1 then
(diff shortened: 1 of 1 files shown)


nitrous24:
The reason i first did clear then set was because of the chance of running out of points in the swap. I am not sure how big of an issue this would be though as i don't play blu, it was just something i was weary of.

AkadenTK:
Ah I see. I wonder why I haven't hit that problem yet... lol. I do see how it could be a problem, though. I'll change it tonight to first unset all spells not in the new set, then set the spells that aren't already set. That should never overflow points.

z16:
Cool beans, then I won't have any qualms about merging it :)

AkadenTK:
Okay, I've updated to unset all spells not in the spell list before moving on to adding the spells not currently set. 
There actually was a problem with the get_current_set function where if there were empty slots, the set would be truncated at the empty slot. For example, if you have 2 spells set at slot01 and slot02, an empty at slot03, and a spell set at slot04, get_current_set would only return the spells in slot01 and slot02. I managed to fix this so that it correctly returns slot01, slot02, and slot04. 
I also bumped up the delay because 0.5s was not enough sometimes.

addons/azureSets/azuresets.lua
```diff
@@ -75,41 +75,90 @@ function set_spells(spellset)
         error(spellset..' was already equipped.')
         return
     end
-    windower.ffxi.reset_blue_magic_spells()
+    --[[windower.ffxi.reset_blue_magic_spells() ]]--
```
z16: Should just remove this instead of commenting it out.

addons/azureSets/azuresets.lua
```diff
@@ -133,15 +182,16 @@ function get_current_spellset()
...
     if windower.ffxi.get_player()['main_job_id'] == 16 then
-        local tmpTable = T(windower.ffxi.get_mjob_data()['spells'])
-        local i,id
-        for i = 1, #tmpTable do
-            local t = ''
-            if tonumber(tmpTable[i]) ~= 512 then
+        local tmp = windower.ffxi.get_mjob_data()['spells']
+        for k,v in pairs(tmp) do
+            if tonumber(v) ~= 512 then
                 for spell in spells:it() do
```
z16: While not technically wrong, I don't know why it iterates the spells to find the one with the right ID... in fact, this entire function does a lot of unnecessary boilerplate to get its result. It can be replaced entirely with this:

``` lua
function get_current_spellset()
    if windower.ffxi.get_player().main_job_id ~= 16 then return nil end
    return T(windower.ffxi.get_mjob_data().spells)
        -- Returns all values but 512
        :filter(function(id) return id ~= 512 end)
        -- Transforms them from IDs to lowercase English names
        :map(function(id) return spells[id].english:lower() end)
        -- Transform the keys from numeric x or xx to string 'slot0x' or 'slotxx'
        :keymap(function(slot) return '%02u':format(slot) end)
end
```

Learn functional programming people, will make your lives easier :)

_Edit: The comments in there can be removed, just added them for educational purposes in here._

addons/azureSets/azuresets.lua
```diff
@@ -75,41 +75,90 @@ function set_spells(spellset)
...
+end
+
+function find_spell_id_by_name(spellname)
+  for spell in spells:it() do
+    if spell['english']:lower() == spellname:lower() then
+        return spell['id']
     end
-    
+  end
+  return -1
```
z16: `nil` should be returned here. This is (thankfully) not C, so we don't need to do weird things when nothing is found :)

addons/azureSets/azuresets.lua
```diff
@@ -75,41 +75,90 @@ function set_spells(spellset)
...
+          --log('Removed spell: '..v..' at #'..slotToRemove)
+
+          windower.send_command('@wait .65;lua i azuresets set_spells_from_spellset '..spellset..' remove')
+          return
+        end
+      end
+    end
+    -- Did not find spell to remove. Start set phase
+    -- Find empty slot:
+    local slotToSetTo = -1
```
z16: Don't need to initialize this (it will be `nil` by default), then later check with `if not slotToSetTo then`.

addons/azureSets/azuresets.lua
```diff
@@ -75,41 +75,90 @@ function set_spells(spellset)
...
+function set_contains_value(set, value)
+  for k,v in pairs(set) do
+    if v:lower() == value:lower() then
+      return true
+    end
+  end
+  return false
+end
+
+function find_spell_id_by_name(spellname)
```
z16, line 136: This function can be removed. instead you can replace any call to this with:

``` lua
spells:with('english', spellname:lower())
```

This will return the spell object or `nil`, if no object is found, then just use `.id` to get its ID.

AkadenTK:
Lua is _far_ from my favorite language. I really miss .NET's Linq, honestly which is a similar concept to the map/filter functions. 
I tested your get_current_spellset, but it errors with "attempt to call method 'keymap' a nil value"
image

addons/azureSets/azuresets.lua
```diff
@@ -75,41 +75,90 @@ function set_spells(spellset)
...
-    else
-        log(spellset..' has been equipped.')
-        windower.send_command('@timers c "Blue Magic Cooldown" 60 up')
+
+    -- Unable to find any spells to set. Must be complete.
+    log(spellset..' has been equipped.')
+    windower.send_command('@timers c "Blue Magic Cooldown" 60 up')
+end
+
+function set_contains_value(set, value)
```
z16: This function can be removed. Instead you can replace any call to this with:

``` lua
spells:contains(value:lower())
```

z16:
@baharms Same here, which is why I added the map/filter/reduce functions :)

Also, it should be `key_map`, not `keymap`. I didn't write any Windower Lua code in a long time, sorry @.@

AkadenTK:
My life is forever changed by maps, filters and key_maps. Also, I really did miss my format()

z16:
Also, one more thing... try to use 4 spaces for indentation to match the current style.

PR #1423, opened 2016-04-07
04/05 Update - Fix

MilkMansKid:


alphaONE2:
Really, you don't need to close/reopen a PR for every change, GitHub is smart enough to group up new commits into a PR after opening one.

Also, it's usually better to just get all of the changes checked in and then open a single PR for all of the changes. PRs can handle multiple commits just fine.

MilkMansKid:
Then I did it wrong because it didn't auto group it like I thought it would. Hence why I closed the last one. Sorry about that haven't used git in like a year so I must of did it wrong.

alphaONE2:
OK, I see what happened, you (possibly not intentionally) created 4 different branches for each commit. GitHub only groups commits together if they're made on the same branch.

PR #1424, opened 2016-04-10
[GearSwap] feature: add `all` option to export

nekonok:
We have to know actual augment property to identify the item.
But `gs export inv` is not useful for this purpose because we have many bags and augmented items now.

### DIFF
--- a/addons/GearSwap/export.lua
+++ b/addons/GearSwap/export.lua
@@ -25,12 +25,14 @@
 --SOFTWARE, EVEN IF ADVISED OF THE POSSIBILITY OF SUCH DAMAGE.
 
 function export_set(options)
-    local item_list = {}
+    local item_list = T{}
     local targinv,xml,all_sets,use_job_in_filename,use_subjob_in_filename,overwrite_existing
     if #options > 0 then
         for _,v in ipairs(options) do
             if S{'inventory','inv','i'}:contains(v:lower()) then
                 targinv = true
+            elseif v:lower() == 'all' then
+                all_items = true
             elseif S{'xml'}:contains(v:lower()) then
                 xml = true
             elseif S{'sets','set','s'}:contains(v:lower()) then
@@ -50,7 +52,9 @@ function export_set(options)
     end
     
     local buildmsg = 'Exporting '
-    if targinv then
+    if all_items then
+        buildmsg = buildmsg..'your all items'
+    elseif targinv then
         buildmsg = buildmsg..'your current inventory'
     elseif all_sets then
         buildmsg = buildmsg..'your current sets table'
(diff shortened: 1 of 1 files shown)


smd111:
if you equip the gear and then use the command "gs export" it will give you all the augments for all the gear that you have equipped

nekonok:
@smd111 Yes. But I think a role of `gs export sets` is building a lua script that suitable for use in sets.lua.
I just want to know augment property. For this purpose, I think executing many times `gs export inv` and `gs export sets` is painful.

And I marked this PR `feature`. I added parameter `all`, this does not change current behavior.

smd111:
you want to get all the correct augments for all your items in every bag??

nekonok:
@smd111 Yes. I always build my sets by my ruby build system.
I think `gs export inv` has similar purpose to this.

nekonok:
@Byrth  What do you think about this PR?

z16:
For the record, @Byrth is very busy with non-FFXI stuff right now, so he might take some time to reply.

nekonok:
@z16 I see. Could you review?

z16:
I'll try to review this in the coming days, but the decision rests with @Byrth. The team only directly merges changes for addons whose developers are MIA.

Byrth:
I'm fine with this as long as it works.

addons/GearSwap/export.lua
```diff
@@ -80,33 +84,13 @@ function export_set(options)
     end
     
     local inv = items.inventory
```
z16: This variable can be moved below into the `elseif` clause. It can even be removed and its usage just replaced with `items.inventory`.

addons/GearSwap/export.lua
```diff
@@ -80,33 +84,13 @@ function export_set(options)
...
-                        end
-                        if string.len(aug_str) > 0 then
-                            item_list[#item_list].augments = aug_str
-                        end
-                    end
-                else
-                    msg.addon_msg(123,'You possess an item that is not in the resources yet.')
-                end
-            end
+    local bag_list = {'inventory', 'safe', 'storage', 'locker', 'satchel', 'sack', 'case', 'wardrobe', 'safe2', 'wardrobe2'}
```
z16: This, similarily to the above, can be moved inside the `if` clause.

addons/GearSwap/export.lua
```diff
@@ -80,33 +84,13 @@ function export_set(options)
...
+    if all_items then
+        for _, bag in ipairs(bag_list) do
```
z16: Instead of using a custom `bag_list` which you define you should probably iterate the `items` table itself. You may need to make sure there's nothing else in there for it to work. The downside to a custom list like this is that it will need to be updated any time it changes.

Alternatively, if iterating the `items` table is not an option, you should define the bag list like this:

``` lua
local bag_list = res.bags:map(function(bag) return bag.english:gsub(' ', '') end)
```

This uses the bags as they are defined in the resources, and thus will get updated automatically with every update. But I'd still prefer the above approach, it's faster and less code.

z16:
I gave my loose change, personally I feel this is ready to be pulled once the `bag_list` thing is adjusted.

nekonok:
Sorry, please wait merging.
`res.bags[0]` is inventory, but `ipairs` don't refer index `0`.

alphaONE2:
Yeah, lua, and therefore ipairs, assumes that table indexes start at 1, but Windower uses the internal IDs used by FFXI wherever possible. These two ideologies are frequently in conflict.

alphaONE2:
Leave a comment here when you think this is ready to be merged, we only get notified when a comment is left, not when you add a commit.

nekonok:
I fixed resource and variable problems. I think this is ready to be merged.
Thank you for advises!

addons/GearSwap/export.lua
```diff
@@ -79,34 +83,12 @@ function export_set(options)
...
-                            item_list[#item_list].augments = aug_str
-                        end
-                    end
-                else
-                    msg.addon_msg(123,'You possess an item that is not in the resources yet.')
-                end
-            end
+    if all_items then
+        for i = 0, #res.bags do
+            item_list:extend(get_item_list(items[res.bags[i].en:gsub(' ', ''):lower()]))
```
z16: Please use `english` instead of `en`. We need to find a better way to document that, I agree, but `en` is just an implementation detail, `english` is the proper name to use.

nekonok:
Fixed :)

PR #1425, opened 2016-04-10
Addon autogm

True-ba-door:
Created the addon "autogm," which will generate alerts and logs when "GM," is detected in chat.


z16:
Do GMs even exist anymore? Anyway, not really any use for this other to support those nasty, dirty cheaters and RMT, so no thanks :)

alphaONE2:
If I'm reading your readme (which really should be written in markdown instead of a docx file) correctly, this is intended for admins of private servers? This isn't really the right place for such an addon.

True-ba-door:
Didn't even think about the whole private server/retail segregation. (People still play retail?) Sorry to waste your time.

Issue #1389, opened 2015-12-14
Scoreboard - Fails to handle "no argument" commands


chitterchitters:
When attempting to turn scoreboard on in game with //sb , this error is displayed in console:

Scoreboard: Lua runtime error: scoreboard/scoreboard.lua:79: attempt to index local 'command' (a nil value)

(If I've posted this in the wrong place, please let me know)

Byrth:
Well, you did find an error (it should handle a blank command like that), but the actual command to turn scoreboard on is //sb visible

chitterchitters:
haha, right you are, I suppose I'll leave it up for the existing error then. Thank you ~

Byrth:
Okay, changed the title to reflect this. It'll probably be fixed shortly.

jerryhebert:
oops

MilkMansKid:
This should be closed. Fix was merged/pushed over 2 months ago.

alphaONE2:
Indeed, we forgot to close the issue.

@MilkMansKid: In the future, if you include the text "Closes #1389" (obviously, replacing the issue number) in your pull request, GitHub will automatically close the referenced issue when the PR is merged.

PR #1432, opened 2016-04-28
"no argument" commands - Fix

MilkMansKid:
This is a fix for Scoreboard - Fails to handle "no argument" commands
#1389 issue.

### DIFF
--- a/addons/scoreboard/scoreboard.lua
+++ b/addons/scoreboard/scoreboard.lua
@@ -75,13 +75,8 @@ windower.register_event('addon command', function()
             assert(loadstring(table.concat({...}, ' ')))()
             return
         end
-        
-        if command == nil then
-	    error('No command given. See //sb help')
-            return
-        end
 
-        command = command:lower() or 'help'
+        command = (command or 'help'):lower()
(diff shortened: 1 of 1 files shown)


addons/scoreboard/scoreboard.lua
```diff
@@ -75,6 +75,11 @@ windower.register_event('addon command', function()
             assert(loadstring(table.concat({...}, ' ')))()
             return
         end
+        
+        if command == nil then
+	    error('No command given. See //sb help')
+            return
+        end
 
         command = command:lower() or 'help'
```
alphaONE2: Should just display the help text instead of telling the user to use `//sb help`. Simplest way to do that is to change this line to this:

``` lua
command = (command or 'help'):lower()
```

Then you don't need the nil check at all.

z16: Also, the indentation is off.

MilkMansKid:
Thanks for the recommendation, I swear you guys teach me something new every-time I try  to help out(good thing) or make me think different then I do and sorry about the indentation.

PR #1434, opened 2016-05-03
Allow users to choose how to set spells.

AkadenTK:
Added new setting and function argument to allow users to choose how to set spells. 
'clearfirst' clears all spells first, then adds all spells of the set back. Faster for large changes, slower for small changes, and preserves order.
'preservetraits' removes spells one at a time and adds new ones in. Preserves traits across sets and faster for small changes.

Also updated the method for checking if the spells have already been set. Not the most elegant solution, but I'm not a native Lua person. Anybody want to comment on how to do it better?


addons/azureSets/azuresets.lua
```diff
@@ -247,8 +268,11 @@ windower.register_event('addon command', function(...)
         elseif comm == 'help' then
             local helptext = [[AzureSets - Command List:')
             1. removeall - Unsets all spells.
-            2. spellset <setname> -- Set (setname)'s spells.
-            3. set <setname> -- Same as spellset
+            2. spellset <setname> (clearfirst|preservetraits) -- Set (setname)'s spells,
+                             optional parameter: clearfirst or preservetraits: overrides
+                             setting to clear spells first or remove individually,
+                             preserving traits where possible.
+            3. set <setname> (clearfirst|preservetraits) -- Same as spellset
```
alphaONE2, line 272: Minor nitpick here: [square brackets] are usually used to indicate optional arguments (&lt;angle brackets&gt; for required arguments). I would also prefer if the help text said exactly what the default was.

addons/azureSets/azuresets.lua
```diff
@@ -234,7 +251,11 @@ windower.register_event('addon command', function(...)
 
         elseif comm == 'spellset' or comm == 'set' then
             if args[1] ~= nil then
-                set_spells(args[1])
+                if args[2] ~= nil then
+                    set_spells(args[1], args[2])
+                else
+                    set_spells(args[1])
+                end
```
alphaONE2: `set_spells(args[1])` is the same as `set_spells(args[1], nil)`, so the nil check is kind of redundant here.

AkadenTK: Good call. I did have some other logic here before, but it has since become simpler.

addons/azureSets/azuresets.lua
AkadenTK, line 272: I've seen it many ways, but I'll use [square brackets]. I also added what the default was: use settings file or if settings aren't configured, use 'preservetraits'.

alphaONE2:
Looks good to me, gonna wait for @z16 to review this before merging though.

addons/azureSets/azuresets.lua
```diff
@@ -68,19 +68,36 @@ windower.register_event('login', initialize)
 
 windower.register_event('job change', initialize:cond(function(job) return job == 16 end))
 
-function set_spells(spellset)
+function set_spells(spellset, setmode)
     if windower.ffxi.get_player()['main_job_id'] ~= 16 --[[and windower.ffxi.get_player()['sub_job_id'] ~= 16]] then return nil end
     if settings.spellsets[spellset] == nil then return end
-    if settings.spellsets[spellset]:equals(get_current_spellset()) then
+    if is_spellset_equipped(settings.spellsets[spellset]) then
         error(spellset..' was already equipped.')
```
z16: I know this isn't your code, but I wouldn't error on this, just print a warning and abort. If you're already messing around in this area, may as well adjust this :) But up to you.

addons/azureSets/azuresets.lua
```diff
@@ -68,19 +68,36 @@ windower.register_event('login', initialize)
...
         return
     end
 
     log('Starting to set '..spellset..'.')
-    set_spells_from_spellset(spellset, 'remove')
+    if setmode=='clearfirst' or (setmode == nil and settings.setmode == 'clearfirst') then
+        remove_all_spells(nil)
```
z16: The `nil` can be removed, will (mostly) have the same effect.

addons/azureSets/azuresets.lua
```diff
@@ -68,19 +68,36 @@ windower.register_event('login', initialize)
...
+        windower.send_command('@wait .65;lua i azuresets set_spells_from_spellset '..spellset..' add')
```
z16: Coroutine magic!

``` lua
set_spells_from_spellset:schedule(0.65, 'add')
```

addons/azureSets/azuresets.lua
```diff
@@ -68,19 +68,36 @@ windower.register_event('login', initialize)
...
+    elseif setmode == 'preservetraits' or (setmode == nil and settings.setmode == 'preservetraits') or settings.setmode == nil then
+        set_spells_from_spellset(spellset, 'remove')
+    end
     return
```
z16: This return statement is unnecessary, since the function is done already.

addons/azureSets/azuresets.lua
```diff
@@ -68,19 +68,36 @@ windower.register_event('login', initialize)
 
 windower.register_event('job change', initialize:cond(function(job) return job == 16 end))
 
-function set_spells(spellset)
+function set_spells(spellset, setmode)
     if windower.ffxi.get_player()['main_job_id'] ~= 16 --[[and windower.ffxi.get_player()['sub_job_id'] ~= 16]] then return nil end
```
z16: The `return nil` can just be a `return`.

addons/azureSets/azuresets.lua
```diff
@@ -68,19 +68,36 @@ windower.register_event('login', initialize)
...
     end
 
     log('Starting to set '..spellset..'.')
-    set_spells_from_spellset(spellset, 'remove')
+    if setmode=='clearfirst' or (setmode == nil and settings.setmode == 'clearfirst') then
+        remove_all_spells(nil)
+        windower.send_command('@wait .65;lua i azuresets set_spells_from_spellset '..spellset..' add')
+    elseif setmode == 'preservetraits' or (setmode == nil and settings.setmode == 'preservetraits') or settings.setmode == nil then
+        set_spells_from_spellset(spellset, 'remove')
+    end
```
z16, line 95: This if-else block is a bit confusing and I think it can be done better. I assume you want to take the user-provided value, and if that is not present, use the `settings.setmode` value? If so, I would change the call-site (in the command handler) to this:

``` lua
set_spells(args[1], args[2] or settings.setmode)
```

This will call the function with a second parameter, if provided, otherwise use the one from the settings table. And, judging by your code, you want it to default to 'preservertraits' if it is nil? Ideally, such a thing should never happen - settings should not be able to be `nil`, so you probably should not consider the case. The defaults can be adjusted so that they hold a value that will be used in case it's missing from the `settings.xml`. I'll add a comment to the relevant position.

Anyway, then you can rewrite the if/else-structure to this:

``` lua
if setmode == 'clearfirst' then
    ...
elseif setmode == 'preservetraits' then
    ....
end
```

It will have the same effect as what you do right now and it's easier to read.

addons/azureSets/azuresets.lua
```diff
@@ -68,19 +68,36 @@ windower.register_event('login', initialize)
...
     return
 end
 
+function is_spellset_equipped(spellset)
+    if spellset:equals(get_current_spellset()) then return true end
+    local currentSet = get_current_spellset()
```
z16: This can be defined one line earlier, then used in the first `if`. Removes one `get_current_spellset` call. However, it may not be relevant, see next comment.

addons/azureSets/azuresets.lua
```diff
@@ -68,19 +68,36 @@ windower.register_event('login', initialize)
...
+    for k,v in pairs(spellset) do
+        if not currentSet:contains(v:lower()) then return false end
+    end
+    for k,v in pairs(currentSet) do
+        if not spellset:contains(v:lower()) then return false end
+    end
+    return true
+end
```
z16, line 100: I'm not entirely sure of the point of this function. `spellset:equals(get_current_spellset())` should already check for memberwise table equality. The only thing it does not consider is case. Is that why you added it? If so, I would probably adjust the tables to validate case and auto-correct themselves on load instead. But that's also up to you.

addons/azureSets/data/settings.xml
```diff
@@ -3,6 +3,8 @@
 	<global>
```
z16: This is also not your fault, but this file should be deleted. It will be auto-generated (and auto-corrected) on load. It won't help adjusting anything in here anyway, as the `data` folder is exempt from updates. The reason for that is that the `data` folder contains only user-added and user-modified files. In this case we would overwrite all user-settings if we updated this XML.

Instead of adjusting the XML, instead adjust the `defaults` table, which is constructed on line 43, and subsequently passed to the config loader on line 58. Add an entry for `setmode` there:

``` lua
defaults.setmode = 'preservetraits'
```

You should delete this file and commit the deletion itself, that way it will be removed from the repository.

z16:
Added my comments, and one more (minor) thing... we use _snake_case_ in Lua files, but normally _PascalCase_ in user-files (such as settings). You should consider changing the settings options to `ClearFirst` and `PreserveTraits`.

You should also adjust the _ReadMe_ file to reflect the new settings options.

addons/azureSets/azuresets.lua
AkadenTK: I like this. I might rewrite the function itself to do this instead, mostly because I hate the send_command anyway. I'll test it out at least.

addons/azureSets/azuresets.lua
AkadenTK, line 95: That's way cleaner. I like it. I'll update it tonight.

addons/azureSets/azuresets.lua
AkadenTK, line 100: It's there because I want to compare values only. The sets may not be in order if they were set with 'preservetraits'.

addons/azureSets/azuresets.lua
AkadenTK: I noticed this too, during testing, but I forgot about it. Will update.

addons/azureSets/azuresets.lua
z16, line 100: I see. In that case you can do:

``` lua
function is_spellset_equipped(spellset)
    return S(spellset) == S(get_current_spellset())
end
```

`S(t)` takes a table and creates a set (i.e. an unordered collection) from its values. So `S({1,2,3}) == S({3,1,2})` would return `true`.

AkadenTK, line 100: I think I should also use `spellset:map(function(s) s:lower())`. (is that the right syntax?) `S(get_current_spellset())` should always be lowercase, but `spellset` can be edited by users and may not be lowercase.

z16, line 100: That works. Like I said above, I'm more of a fan of table validation. To be honest, these tables should be using IDs, not names, but that's too much of a refactoring for this pull request. Anyway, the mapping works, but it can be simplified:

``` lua
function is_spellset_equipped(spellset)
    return S(spellset):map(string.lower) == S(get_current_spellset())
end
```

AkadenTK:
I've updated to include changes specified in comments above.

addons/azureSets/azuresets.lua
```diff
@@ -68,17 +70,25 @@ windower.register_event('login', initialize)
...
         return
     end
 
     log('Starting to set '..spellset..'.')
-    set_spells_from_spellset(spellset, 'remove')
-    return
+    if setmode:lower()=='clearfirst' then
+        remove_all_spells()
+        set_spells_from_spellset:schedule(settings.setspeed, spellset, 'add')
+    else -- if setmode:lower() == 'preservetraits' then
```
z16: Is this meant to be commented out?

addons/azureSets/azuresets.lua
```diff
@@ -94,8 +104,7 @@ function set_spells_from_spellset(spellset, setPhase)
 
                 windower.ffxi.remove_blue_magic_spell(slotToRemove)
                 --log('Removed spell: '..v..' at #'..slotToRemove)
```
z16, line 114: This looks like a previous debug statement and could also be removed, if you feel like it.

addons/azureSets/azuresets.lua
AkadenTK: I originally had that if there but then realized that misspelling either would then do nothing. Looking back, I think it should warn the user if something unexpected happens and do nothing.

addons/azureSets/azuresets.lua
AkadenTK, line 114: I left the debug messages in, but commented for future use. I considered adding a "debug" mode to display these log messages to the user if they choose, but I didn't end up implementing it because it wouldn't be nearly as useful as GearSwap's debug mode...

PR #1442, opened 2016-05-15
Scoreboard - Add ability to report on weapon skill accuracy

gregawoods:
Hello,

This change adds the ability to see your overall weapon skill accuracy via `sb reportstat wsacc`. This is accomplished by incrementing a count for every missed ws and comparing to the total number of data points in the ws damage table.

I arrived at `message_id` 188 for missed skills through trial and error. If there is a better way to track these instances just let me know and I can modify it.

Thanks!

ffxi_2016 05 15_11 25 36

### DIFF
--- a/addons/scoreboard/mergedplayer.lua
+++ b/addons/scoreboard/mergedplayer.lua
@@ -234,8 +234,21 @@ function MergedPlayer:wsavg()
     end
 end
 
-
-
+function MergedPlayer:wsacc()
+    local hits, misses = 0, 0, 0
+    
+    for _, p in ipairs(self.players) do
+        for _ in pairs(p.ws) do hits = hits + 1 end
+        misses = misses + p.ws_misses
+    end
+    
+    local total = hits + misses
+    if total > 0 then
+        return {hits / total, total}
+    else
+        return {0, 0}
+    end
+end
 
 -- Unused atm
 function MergedPlayer:merge(other)
--- a/addons/scoreboard/mergedplayer.lua
+++ b/addons/scoreboard/mergedplayer.lua
@@ -238,7 +238,7 @@ function MergedPlayer:wsacc()
     local hits, misses = 0, 0
     
     for _, p in ipairs(self.players) do
-        for _ in pairs(p.ws) do hits = hits + 1 end
+        hits = hits + table.length(p.ws)
         misses = misses + p.ws_misses
     end
     
--- a/addons/scoreboard/mergedplayer.lua
+++ b/addons/scoreboard/mergedplayer.lua
@@ -235,7 +235,7 @@ function MergedPlayer:wsavg()
 end
 
 function MergedPlayer:wsacc()
-    local hits, misses = 0, 0, 0
+    local hits, misses = 0, 0
     
     for _, p in ipairs(self.players) do
         for _ in pairs(p.ws) do hits = hits + 1 end
--- a/addons/scoreboard/damagedb.lua
+++ b/addons/scoreboard/damagedb.lua
@@ -18,7 +18,7 @@ DamageDB.player_stat_fields = T{
     'mavg', 'mrange', 'critavg', 'critrange',
     'ravg', 'rrange', 'rcritavg', 'rcritrange',
     'acc', 'racc', 'crit', 'rcrit',
-    'wsavg'
+    'wsavg', 'wsacc'
 }
 
 function DamageDB:new (o)
@@ -142,6 +142,7 @@ function DamageDB:add_r_hit(m, p, d)         self:_get_player(m, p):add_r_hit(d)
 function DamageDB:add_r_crit(m, p, d)        self:_get_player(m, p):add_r_crit(d)        end
 function DamageDB:incr_misses(m, p)          self:_get_player(m, p):incr_m_misses()      end
 function DamageDB:incr_r_misses(m, p)        self:_get_player(m, p):incr_r_misses()      end
+function DamageDB:incr_ws_misses(m, p)       self:_get_player(m, p):incr_ws_misses()     end
 function DamageDB:add_damage(m, p, d)        self:_get_player(m, p):add_damage(d)        end
 function DamageDB:add_ws_damage(m, p, d, id) self:_get_player(m, p):add_ws_damage(id, d) end
 
--- a/addons/scoreboard/display.lua
+++ b/addons/scoreboard/display.lua
@@ -364,6 +364,7 @@ Display.stat_summaries._all_stats = T{
     ['crit']       = {percent=true,  category="average", name='Melee Crit. Rate'},
     ['rcrit']      = {percent=true,  category="average", name='Ranged Crit. Rate'},
     ['wsavg']      = {percent=false, category="average", name='WS Average'}, 
+    ['wsacc']      = {percent=true,  category="average", name='WS Accuracy'}, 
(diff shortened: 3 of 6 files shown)


addons/scoreboard/mergedplayer.lua
```diff
@@ -234,8 +234,21 @@ function MergedPlayer:wsavg()
     end
 end
 
-
-
+function MergedPlayer:wsacc()
+    local hits, misses = 0, 0
+    
+    for _, p in ipairs(self.players) do
+        for _ in pairs(p.ws) do hits = hits + 1 end
```
z16: This line is the same as:

``` lua
        hits = hits + table.length(p.ws)
```

Only it avoids the explicit loop. (You probably took the code from the above function, which could also be simplified a bit.)

addons/scoreboard/mergedplayer.lua
```diff
@@ -234,8 +234,21 @@ function MergedPlayer:wsavg()
     end
 end
 
-
-
+function MergedPlayer:wsacc()
+    local hits, misses = 0, 0
+    
+    for _, p in ipairs(self.players) do
```
z16, line 240: I'm not very familiar with _Scoreboard's_ code, but this line looks weird. Wouldn't this sum up the stats from all players? Or what info does the `self.players` variable hold?

I can see that the above function also does this, so I won't go as far as call this wrong, but it's definitely not intuitive... any ideas?

z16:
You should probably log packets to find the right value. You may want to stop by on IRC if you need help on how to do that.

addons/scoreboard/mergedplayer.lua
gregawoods: Ah thanks, that is much cleaner. I had googled this before in search of a general purpose `.length` accessor on the object itself and it seemed like all the answers pointed to doing a loop.

addons/scoreboard/mergedplayer.lua
gregawoods, line 240: This seemed strange to me as well, but it turns out that scoreboard stores one player instance in the database for each type of mob fought (see comments [here](https://github.com/Windower/Lua/blob/dev/addons/scoreboard/player.lua#L4)). So if you fought crabs and bats you would have two player instances per actual player.

z16, line 240: I see, thanks for clearing that up.

addons/scoreboard/mergedplayer.lua
z16: `table.length` is a non-standard function that is available through our libraries, which does exactly that :) Everyone gets tired of writing loops just to count the length of an array. Just part of Lua's weirdness.

From `Windower/addons/libs/tables.lua`:

``` lua
-- Returns the number of elements in a table.
function table.length(t)
    local count = 0
    for _ in pairs(t) do
        count = count + 1
    end

    return count
end
```

If this ever gets refactored, it should probably be changed to use [`lists`](https://github.com/Windower/Lua/blob/dev/addons/libs/lists.lua), which keep a running count and can get the length in constant time. But that's not your job, just mentioning it for possible future maintainers.

z16:
Another question, for a possible future improvement... how does reporting work? Does it write it to the `party` chat? I assume it's configurable where it writes it to? If so it could be expanded to write to the clipboard (with [`windower.copy_to_clipboard`](http://dev.windower.net/doku.php?id=lua:api:functions:start#windowercopy_to_clipboard_str)), then users could have better control over where they wanna post it.

gregawoods:
The default output appears to be reporting in `/say`, which terrifies me personally as I'd rather not broadcast in the middle of a busy area that I'm using an addon. But you can pass options `p` for party chat or `t player` for tell. 

I was a big fan of KParser back when that still worked reliably so I've thought about going in and improving the overall reporting here - for example I'd love to see accuracy stats appearing in the gui overlay. But I haven't had time to really learn the code or Lua just yet.

jerryhebert:
Ahh, I was just about ready to unsubscribe from these since there are so many crickets around this addon.

A few things:

First, I don't remember why I used tables instead of lists. Maybe lists didn't exist yet. :) A lot of that stuff was still being developed at the time that this was written, even the action packet stuff didn't exist yet on the first iteration.

As for the MergedPlayer class, it's a bit weird but players wanted a way to show only the damage done on particular bosses. This allowed for predefined filter sets (which was never implemented, but the idea was something like `//sb set filterset morimar` or whatever). At the time of writing, linkshells were doing delve and one typically wanted to see the filters for only the bosses in the instance or just particular bosses. So I needed some way to filter aggregated damage by mob (I think I had some reason why I chose to store it by player rather than mob but it's been a few years). Storing all of the damage instances so that they could later be queried either way on the fly was considered too memory-intensive for a Lua addon so this approach allowed players to see most of the data with some preemptive aggregations.

If you don't pass an arg to report, this just uses whatever your current chatmode is, which is a feature. It's annoying to have to explicitly tell it to send to party if you're in `/cm p`. I guess the code assumes that an add-on user would not run around in `/cm s` since at least folks usually have `/cm l` enabled. It would be a nice little enhancement to check the player's current chatmode and warn if it's `s`.

As for reporting, the easiest display to get up and running was just some tediously formatted text output. Because of the volume of potential data that would be queried (think: KParser), it didn't seem reasonable to do it in any sort of overlays. The overlay data would have to be limited one way or another, even in full screen, and resolutions vary, so that didn't seem like a good idea.. I got a job, quit FFXI and lost interest in maintaining this before I finished it, but I was working on a way to just dump the full report to a file. It just never got pushed.

Additionally, and I'm sure Arcon remembers some of this, I had some more grandiose ideas which will perhaps never get written considering FFXI's ostensibly limited lifespan at this point but the timeline was something like this:
- Get luasocket capability added to Windower (I was working on this a bit but I think it's complete now?)
- Use sockets to get some sort of local DB access from Lua addons so that bigger data sets could be stored. I never did this but from searching around it looked like CouchDB was going to be a good candidate. I saw a PR for another addon a little while ago where someone wanted to connect to msql. Maybe they did it, but either nosql or sql would be fine.
- Write a local web client that you can use to analyze the state of the DB. Then you can have all sorts of fancy graphs. I was going off inspiration from Recount from WoW, which has some really sexy graphs. In the browser, you get a lot of that sort of thing for free these days, so it seemed like a powerful path to take. Just `//sb report open` or something and it'd fire off a browser that you could query however you like.

Of course, the DB structure (if you can call it a "DB") would have to be rewritten but it would have actually been somewhat legitimized as a true window into FFXI stats with that kind of data collection and reporting.

alphaONE2:
LuaSocket is supported, and I think @z16 added Sqlite too (it has a Lua binding that comes with it). It's just a shame you left before you could finish it.

z16:
Also, you should totally come back and do all that :)

gregawoods:
Wow @jerryhebert, that all sounds amazing! It hadn't occurred to me you could tie into windower data from the browser. I guess we'll all just have to imagine the awesomeness that could have been. :)

And while you're here, thanks for writing scoreboard! It was fun to get to contribute something after being a Windower user for years.

jerryhebert:
@gregawoods, you got this! :)

And @z16, that would require playing FFXI again which ain't gonna happen. :p

z16:
@jerryhebert No it wouldn't, I haven't been playing for years :) Can even have one of my accounts to work on! But let's discuss that elsewhere ;D

PR #1444, opened 2016-05-19
Added a time remaining count down for aby zones

smd111:


smd111:
its a little off because the countdown gets rest ever time the time remaining gets updated but it has to do this or the timer will get vary far off

addons/pointwatch/FIXING POINTWATCH.txt
```diff
@@ -17,6 +17,7 @@ Method 1 - POLUtils
 5) Wait for it to load, then export the entire table as xml
 6) Open the xml file up in a text editor, like Notepad or Notepad++.
 7) Control-F for the key phrase.
+ --look for: ≺Set Color #5≻Visitant Light Intensity≺Set Color #1≻
```
z16: Would remove that, it's a very specific example, and the `<Set Color #x>` stuff will probably only confuse people.

smd111: if you search for "≺Set Color #5≻Visitant Light Intensity≺Set Color #1≻" its exactly what you want for the offset

smd111:
any news on if this will be accepted??

smd111:
thx byrth

PR #1449, opened 2016-06-05
close #1448 [GearSwap] Export file with `%Y-%m-%d %H-%M-%S` format

### LINKED ISSUE #1448: [GearSwap] Change exported filename format for accessibility
When we execute `gs export`, we get a file named like `Character 12 50 47PM  16-22-05.lua`.
But it's hard to find because they can not be sorted with datetime by filenames.
And this date format is not familiar to JP users.

So I suggest to use RFC3339 date format like `2001-02-03 04:05:06+09:00` (Timezone is unnecessary).

Source: https://github.com/Windower/Lua/blob/dev/addons/GearSwap/export.lua#L171

``` lua
        path = path..os.date(' %H %M %S%p  %y-%d-%m')
```

nekonok:
ref #1448 

Now we get files named like `Character 2016-06-05 20-15-10.lua`.

### DIFF
--- a/addons/GearSwap/export.lua
+++ b/addons/GearSwap/export.lua
@@ -168,7 +168,7 @@ function export_set(options)
     elseif use_subjob_in_filename then
         path = path..'_'..windower.ffxi.get_player().main_job..'_'..windower.ffxi.get_player().sub_job
     else
-        path = path..os.date(' %H %M %S%p  %y-%d-%m')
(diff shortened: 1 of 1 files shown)


z16:
While I'm all for this, @Byrth may wanna weigh in on this. Alternatively it could be a setting that users can adjust.

nekonok:
I considered about adding format options. But, is there anyone who wants to get previous format?
I don't know tools that use previous date format to export files.

PR #1455, opened 2016-06-17
Adding TaruChat addon

geno3302:
TaruChat is an addon that creates a server chat for each server.
Uses /taru or /a to chat and //taru and //a as commands.


z16:
This is not mergeable at this point. You seem to have another addon (_RAWR_) included in this PR, which causes merge problems, since the addon already exists. Please create a clean branch to work off of, then submit a new pull request.

Also, while I don't mind you exposing your server like this, it could potentially be abused, so just a fair warning. Additionally, the FFOChat addon already provides intra-server chatrooms. Again, this is not an issue, just reminding you of that.

PR #1463, opened 2016-07-06
Added MacroPolo addon

ghost:
MacroPolo is a simple addon for macro navigation enabling the use of /macro back to return to the previous book and set.

This allows easy sharing between overlapping macro lines.

### DIFF
--- a/addons/MacroPolo/MacroPolo.lua
+++ b/addons/MacroPolo/MacroPolo.lua
@@ -25,7 +25,7 @@ ON ANY THEORY OF LIABILITY, WHETHER IN CONTRACT, STRICT LIABILITY, OR TORT
 (INCLUDING NEGLIGENCE OR OTHERWISE) ARISING IN ANY WAY OUT OF THE USE OF THIS
 SOFTWARE, EVEN IF ADVISED OF THE POSSIBILITY OF SUCH DAMAGE.]]
 
-_addon.version = '0.51'
+_addon.version = '0.52'
 _addon.name = 'MacroPolo'
 _addon.author = 'Omnys@Valefor'
 _addon.commands = {'macropolo','mp'}
@@ -38,10 +38,10 @@ mHistory = {}
 
 windower.register_event('outgoing text',function(original,modified)
 	if windower.regex.match(original,"^/macro(polo)? [0-9]+[-][0-9]+ [0-9]+[-][0-9]+") then
-		local coms = split(string.lower(original)," ")
+		local coms = string.lower(original):split(" ")
 		if windower.regex.match(coms[2],"[0-9]+[-][0-9]+") and windower.regex.match(coms[3],"[0-9]+[-][0-9]+") then
-			local pos2 = split(coms[2],"-")
-			local pos3 = split(coms[3],"-")
+			local pos2 = coms[2]:split("-")
+			local pos3 = coms[3]:split("-")
 			windower.send_command("input /macro book "..pos3[1].."; input /macro set "..pos3[2])
 			mHistory[#mHistory+1] = {pos2[1],pos2[2]}
 			return ""
@@ -81,27 +81,3 @@ windower.register_event('addon command', function(...)
 		end
 	end
 end)
-
---This function was made by Byrth. It's used to split strings
---at a specific character and store them in a table
-function split(msg, match)
-    if msg == nil then return '' end
-    local length = msg:len()
-    local splitarr = {}
-    local u = 1
-    while u <= length do
-        local nextanch = msg:find(match,u)
-        if nextanch ~= nil then
-            splitarr[#splitarr+1] = msg:sub(u,nextanch-match:len())
-            if nextanch~=length then
-                u = nextanch+match:len()
-            else
-                u = length
-            end
-        else
-            splitarr[#splitarr+1] = msg:sub(u,length)
-            u = length+1
-        end
-    end
-    return splitarr
-end
--- a/addons/MacroPolo/MacroPolo.lua
+++ b/addons/MacroPolo/MacroPolo.lua
@@ -1,0 +1,102 @@
+--[[
+Copyright © 2016, Omnys of Valefor
+All rights reserved.
+
+Redistribution and use in source and binary forms, with or without
+modification, are permitted provided that the following conditions are met:
+
+    * Redistributions of source code must retain the above copyright
+    notice, this list of conditions and the following disclaimer.
+    * Redistributions in binary form must reproduce the above copyright
+    notice, this list of conditions and the following disclaimer in the
+    documentation and/or other materials provided with the distribution.
+    * Neither the name of MacroPolo nor the
+    names of its contributors may be used to endorse or promote products
+    derived from this software without specific prior written permission.
+
+THIS SOFTWARE IS PROVIDED BY THE COPYRIGHT HOLDERS AND CONTRIBUTORS "AS IS" AND
+ANY EXPRESS OR IMPLIED WARRANTIES, INCLUDING, BUT NOT LIMITED TO, THE IMPLIED
+WARRANTIES OF MERCHANTABILITY AND FITNESS FOR A PARTICULAR PURPOSE ARE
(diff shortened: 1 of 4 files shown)


ghost:
Is this addon permissible? I know I hijack a unique syntax involving the /macro command but I wanted to make it easy for people to use it in macros without the console command as that confuses a lot of people.

I searched for a little bit but it seemed impossible to derive the the current macro position without the user specifying it.

z16:
It's permissible, I just didn't have that much time to review it yet.

One thing to note is that you don't need the `split` function, as there is a `string.split` function:

``` lua
print('one two three':split(' ')) -- Will print "[one, two, three]"
```

I would advocate to not use that syntax, instead go back to `//macro`, but that's really up to you. I'll give a more detailed review soon, maybe tomorrow.

ghost:
Thanks,

Updated to use the built in split function.

Yeah, I mean I can use outgoing text to capture anything, and of course it _is_ running briefly at every occurrence of outgoing text, every **/p hi**, etc. Again, the only motivation for using this method was to let people easily use it in macros rather than try to find out why **//macropolo...** doesn't directly work in macros (instead, having to use say, **//console macropolo...**)

Since it is running at every outgoing text (every /p, every /ws, etc), would it benefit to use a simpler if surrounding the regex match like..

`if original:sub(1,6) == '/macro' then`
`if windower.regex.match...`

z16:
Yes, it would be more efficient, though I'd use this:

``` lua
windower.register_event('outgoing text', function(original)
    if not original:startswith('/macro') then
        return
    end

    local current_book, current_set, target_book, target_set = original:split(' ', 2)[2]:match('(%d+)-(%d+) (%d+)-(%d+)')
    if not target_set then -- Failed to parse
        error('Invalid command')
    end

    -- Logic here
end)
```

However, I'm not convinced of the argument for `/macro` vs `//macro`. It has been a very well-known Windower shortcoming for many years, and I don't think anyone who would use an addon such as this does not know about it. And even if there are such people, this could be explained in a _ReadMe_ file.

Which brings me to your help text, imo it's too long and should really only be a one-liner about command usage. Explaining what it does is what a _ReadMe_ file is for.

I'll comment on it again at some point this weekend, sorry it's taking me so long.

ghost:
Thanks. I understand you're a busy individual.

I never minded using **//macro** over **/macro**. My thoughts were that //macro won't work when used directly inside a macro. **/console ...** will, but I'm trying to make a simple syntax available. Over my years in FFXI, people have asked me several times "Why won't my macro execute this windower command?".

I tried to use **//macro**, with the outgoing text event and it would not fire. I don't run many addons and nothing unusual so perhaps something pretty typical is overriding it.

At one time, test code literally was

```
windower.register_event('outgoing text', function(original)
    print(original)
end)
```

Since I couldn't get it to work using **//macro** in macros, I migrated everything to support **/macropolo** (via outgoing text) and **//macropolo** & **/console macropolo ...** (via addon command). This way it will work for novices and people familiar with using windower commands in macros. It no longer employs /macro in any way.

I actually expect more people that use it to be new/returning, and working on macros for macro-intense jobs like Scholar, rather than people very comfortable with their current macro layout.

I also moved most of help text to a README.md as you suggested.

Thanks also for for the condensing of my regex using split().

Edited for typo.

ghost:
Also, I removed mp as an addon command. Since this isn't intended to be typed out each time, the length of the command makes little difference. I removed it mainly because it's too easy that people might use mp as an alias for something, since it's a game term.

ghost:
Hi, any updates?

z16:
Sorry, I'll review this in the coming days.

z16:
Ouch, sorry, this totally passed me by. I'll review it now, but I'll note one thing: There are suddenly two addons here. Could you please remove _TrustWorthy_ from this and submit a separate pull request for it? I'm gonna focus on reviewing _MacroPolo_ here. That makes it easier to handle them separately.

addons/MacroPolo/MacroPolo.lua
```diff
@@ -0,0 +1,72 @@
...
+
+function Commander(...)
+    local args = T{...}:map(string.lower)
+    if args[1] == nil or args[1] == "help" then
+        log("Usage: /macropolo [current_book]-[current_set] [destination_book] [destination_set]")
+        log("/macropolo back to return to the previous set.")
+        log("Please see README.md for more details.")
+    elseif args[1] == "back" then
+        if #mHistory then
+            windower.send_command("input /macro book "..mHistory[#mHistory][1].."; input /macro set "..mHistory[#mHistory][2])
```
z16, line 46: Should use `windower.chat.input('/something')` instead of `windower.send_command('input /something')`. That accesses FFXI's input box directly and doesn't go through our console handler. A bit more efficient and less error prone. Should also split the two to separate function calls:

``` lua
            windower.chat.input('/macro book ' .. mHistory[#mHistory][1])
            windower.chat.input('/macro set ' .. mHistory[#mHistory][2])
```

addons/MacroPolo/MacroPolo.lua
```diff
@@ -0,0 +1,72 @@
...
+        else
+            error("No previous macro location recorded.")
+        end
+    else
+        local argstring = table.concat(args)
+        local current_book, current_set, target_book, target_set = argstring:match('(%d+)-(%d+) (%d+)-(%d+)$')
+        
+        if target_set then
+            windower.send_command("input /macro book "..target_book.."; input /macro set "..target_set)
```
z16, line 55: See above.

addons/MacroPolo/MacroPolo.lua
```diff
@@ -0,0 +1,72 @@
...
+            mHistory[#mHistory+1] = {current_book,current_set}
```
z16, line 56: Just to make the code more readable I'd suggest the following things, but they're really up to you:
1. I'd use our lists library: `mHistory = L{}`. Then you can add an element by doing: `mHistory:add({book, set})`. You don't have to work with `#` to get the length all the time. This helps above as well, you can just do `mHistory:last()` to get the last element (or `mHistory:last(2)` to get the second to last, and so on).
2. I'd use named table keys for the book/set record: `{book = current_book, set = current_set }`. That way you can do `mHistory:last().book` and `mHistory:last().set`, instead of using `1` and `2`, which does not make it obvious what it's supposed to do.

addons/MacroPolo/README.md
```diff
@@ -0,0 +1,11 @@
...
+
+Usage within macros: 
+/macropolo [current_book]-[current_set] [destination_book] [destination_set]
+/console macropolo [current_book]-[current_set] [destination_book] [destination_set]
+
+Example: '/macropolo 3-1 1-5': records the current macro position, as Book 3 Set 1 and then sends the user to Book 1, Set 5.
+
+You may then use either of these commands in a macro to return to the previous book and set.
+/macropolo back
+/console macropolo back
```
z16, line 11: Nothing to critique here, but I'd beef up the file with some markdown. That's why it's called `.md` after all :)

addons/MacroPolo/MacroPolo.lua
```diff
@@ -0,0 +1,72 @@
...
+            return ""
+        else
+            error('Invalid Command: Syntax should be: /macropolo current_book-current_set target_book-target_set')
+            log('See README.md for more details.')
+        end
+    end
+end
+
+windower.register_event('outgoing text', function(original)
```
z16, line 65: As I mentioned before, I'm not a fan of reacting to both outgoing text and addon commands separately, but again, this is up to you. I'd pick one (addon command) and only use that, and trust the users to be smart enough to figure it out. If you insist on outgoing text, I'd make it react to just `/macro`, so it looks like it's extending the in-game macro command. Right now it's separate from the in-game command, yet separate from standard Windower commands as well, in a strange middle place.

addons/MacroPolo/MacroPolo.lua
```diff
@@ -0,0 +1,72 @@
...
+    if original:startswith('/macropolo ') then
+        Commander(original:sub(tonumber(original:find(" "))))
```
z16, line 67: `string.find` already returns a number, the `tonumber` call is unnecessary.

PR #1465, opened 2016-07-06
Update slips.lua

abira2000:
LL Charuqs+1


z16:
What was the issue with this? Was it the wrong item ID?

abira2000:
Mistake ID of Luhlaza Charuqs+1 was the same as the ID of the Luhlaza Charuqs
27358→27359

PR #1508, opened 2017-03-08
New addon Omen

Sechs:
Added a new Omen addon and all the relative information into addons.xml etc
Addon has a few minor bugs still, but it has been tested live and works pretty well within the limits of what I befined in the description file

### DIFF
--- a/addons/Omen/Omen.lua
+++ b/addons/Omen/Omen.lua
@@ -26,11 +26,12 @@
 
 _addon.name    = 'Omen'
 _addon.author  = 'Braden, Sechs'
-_addon.version = '1.3'
+_addon.version = '1.4'
 _addon.command = 'omen'
 
 config = require ('config')
 texts = require('texts')
+--require('omen_test')
 
 defaults = T{}
 defaults.text_R = 255 --Color values are in RGB, ranging from 0 to 255
@@ -45,6 +46,7 @@ defaults.bad_B = 0
 defaults.pos_x = 0
 defaults.pos_y = 0
 defaults.font_size = 11
+defaults.bg_alpha = 255
 
 settings = config.load(defaults)
 
@@ -54,46 +56,48 @@ omens = 0
 obj_time = 0
 floor_obj = "Waiting for objectives..."
 floor_clear = ""
-IMAGE = texts.new("IMAGE")
+--image = texts.new("image", settings)
+image = texts.new("image")
 
-texts.color(IMAGE,settings.text_R,settings.text_G,settings.text_B)
-texts.size(IMAGE,settings.font_size)
-texts.pos_x(IMAGE,settings.pos_x)
-texts.pos_y(IMAGE,settings.pos_y)
+texts.color(image,settings.text_R,settings.text_G,settings.text_B)
+texts.size(image,settings.font_size)
+texts.pos_x(image,settings.pos_x)
+texts.pos_y(image,settings.pos_y)
+texts.bg_alpha(image,settings.bg_alpha)
 
 function reset_objectives()
-objectives = {
-[1] = {id=1,mes=0,amt=0,req=0},
-[2] = {id=2,mes=0,amt=0,req=0},
-[3] = {id=3,mes=0,amt=0,req=0},
-[4] = {id=4,mes=0,amt=0,req=0},
-[5] = {id=5,mes=0,amt=0,req=0},
-[6] = {id=6,mes=0,amt=0,req=0},
-[7] = {id=7,mes=0,amt=0,req=0},
-[8] = {id=8,mes=0,amt=0,req=0},
-[9] = {id=9,mes=0,amt=0,req=0},
-[10] = {id=10,mes=0,amt=0,req=0}
-}
-obj_time = 0
-floor_clear = ""
+    objectives = {
+    [1] = {id=1,mes=0,amt=0,req=0},
+    [2] = {id=2,mes=0,amt=0,req=0},
+    [3] = {id=3,mes=0,amt=0,req=0},
+    [4] = {id=4,mes=0,amt=0,req=0},
+    [5] = {id=5,mes=0,amt=0,req=0},
+    [6] = {id=6,mes=0,amt=0,req=0},
+    [7] = {id=7,mes=0,amt=0,req=0},
+    [8] = {id=8,mes=0,amt=0,req=0},
+    [9] = {id=9,mes=0,amt=0,req=0},
+    [10] = {id=10,mes=0,amt=0,req=0}
+    }
+    obj_time = 0
+    floor_clear = ""
 end
 reset_objectives()
 
 function refresh()
-header = floor_clear..floor_obj.."\\cr     Omens: "..omens
-body = "\n Bonus Objectives    "..os.date('%M:%S', obj_time)
+    header = floor_clear..floor_obj.."\\cr     Omens: "..omens
+    body = "\n Bonus Objectives    "..os.date('%M:%S', obj_time)
     for k,v in pairs (hide_timer) do
         if string.find(header,v) then
             body = ""
-            texts.text(IMAGE,header)
+            texts.text(image,header)
             return
         end
     end
-    for v=1,table.getn(objectives),1 do
-        if objectives[v].mes ~= 0 then
-            local msg = objectives[v].mes
-            local cur = objectives[v].amt
-            local fin = objectives[v].req
+    for v, objective in ipairs(objectives) do
+        if objective.mes ~= 0 then
+            local msg = objective.mes
+            local cur = objective.amt
+            local fin = objective.req
             if cur == fin then
                 body = body.."\n "..good_col..v..": "..messages[msg].short.." ["..cur.."/"..fin.."]\\cr"
             elseif obj_time < 1 and cur < fin then
@@ -103,41 +107,12 @@ body = "\n Bonus Objectives    "..os.date('%M:%S', obj_time)
             end
         end
     end
-body = string.gsub(body,"%-1","%?%?%?")
-texts.text(IMAGE,header..body)
+    body = string.gsub(body,"%-1","%?%?%?")
+    texts.text(image,header..body)
 end
 
-hide_timer = {"Kin","Gin","Kei","Kyou","Fu","Craver","Gorger","Thinker","Treasure","Waiting"}
+hide_timer = {"Kin","Gin","Kei","Kyou","Fu","Ou","Craver","Gorger","Thinker","Treasure","Waiting"}
 refresh()
-
-
-function test(str)
-windower.add_to_chat(161,tostring(str))
-end
-
-function floor_test()
-windower.add_to_chat(161,"Vanquish 18 sweetwater foes."..string.char(0x7f).."1")
---windower.add_to_chat(161,"Vanquish the Glassy Gorger."..string.char(0x7f).."1")
-coroutine.sleep(2)
-windower.add_to_chat(161,"You have 90 seconds remaining.")
-coroutine.sleep(2)
-windower.add_to_chat(161,"5: You have reduced your foe's HP by 1500 in a single auto-attack.")
-coroutine.sleep(1)
-windower.add_to_chat(161,"5: You have reduced your foe's HP by at least 2000 in a single auto-attack.")
-windower.add_to_chat(161,"3: You have reduced your foe's HP by 1500 using a single magic attack without performing a magic burst.")
-windower.add_to_chat(161,"3: You have reduced your foe's HP by 5000 using a single magic attack without performing a magic burst.")
-windower.add_to_chat(161,"3: You have reduced your foe's HP by 14999 using a single magic attack without performing a magic burst.")
-windower.add_to_chat(161,"3: You have reduced your foe's HP by at least 15000 using a single magic attack without performing a magic burst.")
-coroutine.sleep(1)
-windower.add_to_chat(161,"1: Vanquish 6 foes.")
-windower.add_to_chat(161,"2: Use 30 weapon skills on your foes.")
-windower.add_to_chat(161,"3: Reduce your foe's HP by at least 15000 using a single magic attack without performing a magic burst.")
-windower.add_to_chat(161,"4: Use 12 abilities on your foes.")
-windower.add_to_chat(161,"5: Reduce your foe's HP by at least 2000 in a single auto-attack.")
-coroutine.sleep(1)
-windower.add_to_chat(161,"A spectral light flares up."..string.char(0x7f).."1")
-end
-
 
 windower.register_event('prerender', function()
     if obj_time < 1 then return end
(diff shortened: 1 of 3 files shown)


addons/Omen/Omen.lua
```diff
@@ -0,0 +1,287 @@
...
+    end
+end)
+
+texts.visible(IMAGE,false)
+if windower.ffxi.get_info() and windower.ffxi.get_info().zone == 292 then
+texts.visible(IMAGE,true)
+end
+
+windower.register_event('incoming text', function(original, modified, mode)
+    if mode == 161 then -- Omen messages are 161 color, except total time extension messages which are 121 and irrelevant
```
alphaONE2: I'm sure this works, but it seems like a very messy way to retrieve objectives. You may want to consider using a packet-based solution in the future.

A packet-based solution has the benefit of also working on the Japanese client, without any extra work.

alphaONE2 (approved):
I don't see any glaring issues. I'll wait for @z17 to review this too, but looks OK to merge.

addons/Omen/Omen.lua
z17: Maybe you mean @z16 ?

alphaONE2: Yes, I did. My appologies for the typo.

Sechs: When I asked about it on Windower4 thread on BG I got replied that nobody knew which packets exactely Omen messages used. See reference https://www.bluegartr.com/threads/115131-Windower-4-Discussion-and-Support?p=6860317&viewfull=1#post6860317
Plus like 99,9999% of this addon has been developed by Braden, I only fixed a couple of bugs, added a few things, and my main interest in publishing the addon here is because me and my friends find it really useful, and I think it would be cool to have it on the official windower launcher.
I honestly lack the skill to use packets and even more to reverse engineer to find which bytes exactely are the Omen data.

Aside from these practical limitations and on a theoretical plan I 101% agree with you packets would be the best solution, not a coincidence I asked for those first when the addon didn't exist yet.

alphaONE2: It was just a suggestion. We certainly won't reject the PR for this.

addons/Omen/Omen.lua
```diff
@@ -0,0 +1,287 @@
...
+
+settings = config.load(defaults)
+
+good_col = "\\cs("..tostring(settings.good_R)..","..tostring(settings.good_G)..","..tostring(settings.good_B)..")"
+bad_col = "\\cs("..tostring(settings.bad_R)..","..tostring(settings.bad_G)..","..tostring(settings.bad_B)..")"
+omens = 0
+obj_time = 0
+floor_obj = "Waiting for objectives..."
+floor_clear = ""
+IMAGE = texts.new("IMAGE")
```
z16: Would change this to:
```lua
IMAGE = texts.new("IMAGE", settings)
```

That way anytime the text box is dragged on the screen it will immediately save its location and people don't need to mess with the settings file. It also means you can remove the four lines below. Probably better test it though, since it's been a while since I used this, so not 100% sure.

addons/Omen/Omen.lua
```diff
@@ -0,0 +1,287 @@
...
+
+texts.color(IMAGE,settings.text_R,settings.text_G,settings.text_B)
+texts.size(IMAGE,settings.font_size)
+texts.pos_x(IMAGE,settings.pos_x)
+texts.pos_y(IMAGE,settings.pos_y)
+
+function reset_objectives()
+objectives = {
```
z16: The indentation for the entire file is pretty fucked and makes the code awful to read and comprehend. I'd really urge you to format it first before submitting it. If you dread doing it by hand I'm sure there are tools that can do it.

addons/Omen/Omen.lua
```diff
@@ -0,0 +1,287 @@
...
+header = floor_clear..floor_obj.."\\cr     Omens: "..omens
+body = "\n Bonus Objectives    "..os.date('%M:%S', obj_time)
+    for k,v in pairs (hide_timer) do
+        if string.find(header,v) then
+            body = ""
+            texts.text(IMAGE,header)
+            return
+        end
+    end
+    for v=1,table.getn(objectives),1 do
```
z16:
```lua
for v, objective in ipairs(objectives) do
```

Then use `objective` instead of `objective[v]` below.

addons/Omen/Omen.lua
```diff
@@ -0,0 +1,287 @@
...
+    end
+body = string.gsub(body,"%-1","%?%?%?")
+texts.text(IMAGE,header..body)
+end
+
+hide_timer = {"Kin","Gin","Kei","Kyou","Fu","Craver","Gorger","Thinker","Treasure","Waiting"}
+refresh()
+
+
+function test(str)
```
z16: Are the two test functions required to be here? If you only needed them for development consider removing them. For future tests you can keep them in a separate file.

addons/Omen/Omen.lua
```diff
@@ -0,0 +1,287 @@
...
+    floor_obj = "Waiting for objectives..."
+    reset_objectives()
+    if zone == 292 then -- Henge
+        texts.visible(IMAGE,true)
+    end
+end)
+
+texts.visible(IMAGE,false)
+if windower.ffxi.get_info() and windower.ffxi.get_info().zone == 292 then
+texts.visible(IMAGE,true)
```
z16:
```lua
IMAGE:hide()
```
and
```lua
IMAGE:show()
```

Also, it bothers me more than it probably should that it's called `IMAGE` and not `image`…

addons/Omen/Omen.lua
```diff
@@ -0,0 +1,287 @@
...
+    texts.visible(IMAGE,false)
+    floor_obj = "Waiting for objectives..."
+    reset_objectives()
+    if zone == 292 then -- Henge
+        texts.visible(IMAGE,true)
+    end
+end)
+
+texts.visible(IMAGE,false)
+if windower.ffxi.get_info() and windower.ffxi.get_info().zone == 292 then
```
z16: The first check is unnecessary, `windower.ffxi.get_info` always returns a table.

addons/Omen/Omen.lua
```diff
@@ -0,0 +1,287 @@
...
+texts.visible(IMAGE,true)
+end
+
+windower.register_event('incoming text', function(original, modified, mode)
+    if mode == 161 then -- Omen messages are 161 color, except total time extension messages which are 121 and irrelevant
+        if string.match(original,"^%d") then
+            for k,v in pairs (messages) do
```
z16:
```lua
local objective = objectives[tonumber(original:match("^%d"))]
```

Then use `objective` instead of that whole thing below.

addons/Omen/Omen.lua
```diff
@@ -0,0 +1,287 @@
...
+                    objectives[tonumber(original:match("%d"))].amt = tonumber(string.sub(original:match(v.check),1,-2))
+                    if objectives[tonumber(original:match("%d"))].mes == 0 then -- if loading mid-floor
+                        objectives[tonumber(original:match("%d"))].mes = tonumber(v.id)
+                        objectives[tonumber(original:match("%d"))].req = -1
+                    end
+                end
+                refresh()
+            end
+        elseif string.find(original,"%d+ omen") then
+            omens = original:match("%d+")
```
z16, line 160: I don't know the format of the messages, but this could theoretically match a different number than the `if` check above. I'd restrict the `find` string some more and match the `match` string accordingly.

addons/Omen/Omen.lua
```diff
@@ -0,0 +1,287 @@
...
+            refresh()
+        elseif string.find(original,"You have %d+ seconds remaining.") then
+            if obj_time == 0 then
+                obj_time = tonumber(original:match("%d+"))
```
z16, line 164: Same as above.

addons/Omen/Omen.lua
z16: I agree with @alphaONE2 but I can see the argument. Anyway, since the addon is only meant to act on `mode` `161` I would invert the if check:
```lua
    if mode ~= 161 then
        return
    end

    …
```

This removes unnecessary indentation.

z16 (commented):
Most of the things I mentioned are just minor/optional changes. I'd really like to see the text settings change, because it would really improve the UX. The other stuff is just to make the code better and a bit more stable, but offers no improvement to the user. And while I won't reject it because of it I'd really like to see you fix the formatting, it triggers my OCD in all the unfun places...

addons/Omen/Omen.lua
Sechs: This is a nice idea, but it's saving a lot of stuff that shouldn't be into the xml, I mean A LOT. Any way to reduce that? I fear it might be a bit overwhelming for someone opening and trying to customize things, plus I would have to define more defaults at the beginning of the addon. (I already intended to add the BG alpha among the settings, anyway)

addons/Omen/Omen.lua
Sechs: I fixed some of it and removed all tabs substituting them with spaces, but I left out a lot of stuff (remember, 98% of the addon was built by Braden, not by me). I agree with this, will try to fix it the best I can and resubmit.

addons/Omen/Omen.lua
Sechs: Tested it. Seems to be working.

addons/Omen/Omen.lua
Sechs: I guess I can put em in a separate file and then custom-include it at start when I need to test. And yeah, they absolutely don't need to be there. I've been just using them to test objective tracking and colours and blahblah when outside of Omen. Good point on removing them, will do.

addons/Omen/Omen.lua
Sechs: Replaced lines accordingly, lowercased IMAGE everywhere

addons/Omen/Omen.lua
Sechs: Works.

addons/Omen/Omen.lua
Sechs: I think we gonna be safe with this, it's already restricted by the 161 code and stuff before. Nobody really knows the exact meaning of those messages anyway, it's just a cool icing on the cake to have the addon display the last "omen number" that got announced.
Format is exactely that btw, a sentence that says "xxx omens" at the end. Match string rightfully stops before the plural "s", at "omen".

Sechs:
Updated with (most) of the requests that came out.
Mafai appearently found all the dats and messages and bytes in the packets for the messages, if he can find free time maybe the addon in the future can be upated to use packet logic, I don't have the skills to do that, and until then the addon works fine as is.

Small list of changes:
1) Added bg_alpha to configurable stuff
2) Added new boss "Ou" to the list of bosses
3) furtherly improved text format and indentation, lowercased "IMAGE" to "image"
4) removed test functions to an external file, inserted a commented require at the beginning so that when I want to test it locally I can just uncomment that line
5) some quality fixes with objectives as suggested by Arcon

Basically everything that was suggested except the Settings thing, the 161 reverse check, the stricter string match for "omens" messages.
I tested the (image,settings) thing, it works fine but it creates a mess in the xml file and I don't really like that.

Sechs:
Good and bad news. Finally took a chance to test the addon in the "alternative path".
To my big surprise it works perfectly except for one thing which I haven't bee able to solve.
The addon only manages to track 9 additional objectives out of ten.
The Tenth "overwrites" the first in the list, and appears as objective number 1.

I'm pretty confident the source of this is the fact the addon is handling only single digit. So single digit 10 becomes 1, and hence it overwrites the original number 1.
I haven't been able to fix it though, did a few tests but none worked.
If anybody has an idea on how to solve this small thing I'm all ears.

Sechs:
Any news to this? :3

alphaONE2:
I haven't merged this primarily because of the bug you mentioned with the tenth objective. If you feel like that's not a particularly pressing concern, and would like to move ahead with it as is, we can. I would just request that the bug be documented somewhere.

Sechs:
A bug is bad nonetheless, period. There are three current known bugs in Omen, I've been unable to fix them. It's not like I didn't have time, I simply can't tell how to fix em.

Bug #1: If you get a free floor on "monsters" floor, the addon won't track the additional objectives (since the "main" obj gets completed before the additional objs get even loaded up)
Bug #2: On miniboss, megaboss and treasure floors, once the main obj is complete, the window sill shrink a bit on the left side. Doesn't happen on "monsters" floors.
Bug #3 on the alternate floor 3 path, the 10th and 1st objective will overwrite each other, the addon will hence display 9 objectives


Bug 1 and 2 are hardly relevant, and I have no clue what causes them.
Bug 3 I have a rough idea of what causes it (I bet 10 gets truncated to 1 somewhere, so it occupies the same spot as the real obj 1). I've done a lot of tests but have been unable to fix it.
Is it important? It's not that big of a deal, but it's something that has to be fixed imho.
If someone has a clue I'm open to suggestions :x

addons/Omen/Omen.lua
```diff
@@ -0,0 +1,259 @@
...
+    if texts.visible(image) then
+        image:hide()
+    else
+        image:show()
+    end
+end)
+
+messages = {
+[1] = {id="1",long="Weapon Skill Damage",short="WS Damage",check="%d+%su",
+init="%d: Reduce your foe's HP by %a*%s*%a*%s*%d+ using a single weapon skill.",
```
z16, line 204: The bug you mentioned could be caused by incorrect number matching. Is the `%d` here meant to represent the floor number? Because if so, that would be wrong, it should be `%d+`, since `%d` will only match a single digit. So `10` would be matched as `0`. I would try replacing all the `%d` in the following lines with `%d+`.

addons/Omen/Omen.lua
```diff
@@ -0,0 +1,259 @@
...
+    end
+end)
+
+image:hide()
+if windower.ffxi.get_info().zone == 292 then -- 292 is the code for Reisenjima Henge
+    image:show()
+end
+
+windower.register_event('incoming text', function(original, modified, mode)
+	local objective = objectives[tonumber(original:match("^%d"))]
```
z16: Actually, this one is likely the cause for the bug. It should be `^%d+`. `^%d` would only match the `1` out of `10`. Generally you should replace all `%d` with `%d+` unless you explicitly only want to match one digit (which is very rarely). See if that fixes it.

addons/Omen/Omen.lua
z16: Imprecise matching is precisely what led to the bug you're experiencing, which is why you should always try to be explicit :)

z16 (changes requested):

KenshiDRK:
The Omen messages ids are on ROM/353/58.dat and I guess the packets are the action message and the kill message but Im not sure. I'll try to track those messages next time I go to omen but the spam could be very bad.

z16 (approved):

addons/Omen/Omen.lua
Sechs: Yeah, tested it and it works. Since I didn't 100% trust my custom test function, I went in and tested in real omen, seems to be working. Thanks!

Sechs:
Updated readme and addon with the 10 objectives tracking fix! :)

PR #1509, opened 2017-03-10
Delete AutoLockStyle

cantreadcode:
Originally was made to reapply /lockstyle after zoning. Was replaced by SE's own code. Obsolete addon.


trv6:
I don't know how this is handled, but you'd have to delete the entry in the addons manifest as well.

z16:
As someone who has never used this addon before - does it do anything that `/lockstyle` doesn't? Or is that a full replacement for the addon?

zetaepyon:
I haven't used it myself either, but I guess it also lets you specify the set that you want to /lockstyle. 

The primary purpose though, I think, was to automatically reapply lockstyle each time you zoned, from prior to when lockstyle was persistent across zoning in general.

Byrth:
I did this and also updated addons.lua.

The addon had no functionality that isn't now provided by the game, as far as I can see.

PR #1512, opened 2017-03-29
config and fields libs Update

KenshiDRK:
Updated config lib to solve an error if t_merge was nil and updated the
packet 0x027 on fields.lua

### DIFF
--- a/addons/libs/config.lua
+++ b/addons/libs/config.lua
@@ -142,116 +142,118 @@ function merge(t, t_merge, path)
     for key in pairs(t) do
         keys[tostring(key):lower()] = key
     end
+
+    if not t_merge then
+        return t
+    end
     
-    if t_merge then
-        for lkey, val in pairs(t_merge) do
-            local key = keys[lkey:lower()]
-            if not key then
-                if type(val) == 'table' then
-                    t[lkey] = setmetatable(table.copy(val), getmetatable(val) or _meta.T)
-                else
-                    t[lkey] = val
-                end
-
+    for lkey, val in pairs(t_merge) do
+        local key = keys[lkey:lower()]
+        if not key then
+            if type(val) == 'table' then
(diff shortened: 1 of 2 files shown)


addons/libs/config.lua
```diff
@@ -142,114 +142,116 @@ function merge(t, t_merge, path)
     for key in pairs(t) do
         keys[tostring(key):lower()] = key
     end
+    
+    if t_merge then
```
z16: This diff is awful to read due to the indentation. Was all you changed the if check? If so, I'd prefer for you to just use this:
```lua
    if not t_merge then
        return t
    end
```

addons/libs/packets/fields.lua
```diff
@@ -1554,15 +1554,19 @@ fields.incoming[0x026] = L{
     {ctype='data[22]',          label='_unknown2',          const=0},           -- 06
 }
 
--- Encumbrance Release
+local function mid(val)
```
z16: I assume this should stand for "message ID". I find the name a bit odd. Personally I'd just use a function like this:
```lua
local function sub(amount, val)
    return val - amount
end
```

Then you can use it like so:
```lua
    {ctype='unsigned short',    label='Message ID',         fn=sub+{0x8000}},   -- 0A
```

This is more general and we could use it in other places as well (there are several packets that contain values with constants offsets). You might also add an `add` function that does the same but adding the amount.

I would also place the function at the top, along with the others. Take a look at [line 65](https://github.com/Windower/Lua/blob/dev/addons/libs/packets/fields.lua#L65).

addons/libs/config.lua
KenshiDRK: yea the if check is all I changed

addons/libs/packets/fields.lua
KenshiDRK: I'll make the changes as soon as possible

PR #1519, opened 2017-04-18
DynamisHelper: refactor and update with additional Treasury changes

Skyrant:
Added nicer interface with more options.  works with dynamis campaign
event. general refactoring of code and optimization.


addons/dynamishelper/dynamishelper.lua
```diff
@@ -27,19 +27,307 @@
...
+        player = windower.ffxi.get_player().name
+        obtained = nil
+    end
+end)
+
+-------------------------------------------------------------------------------
+-- Parse the chat messages for drops, staggers and time extensions ------------
+-------------------------------------------------------------------------------
+windower.register_event('incoming text',function (original, new, color)
+	a,b,fiend = string.find(original,"%w+'s attack staggers the (%w+)%!")
```
z16: Consider making these local.

addons/dynamishelper/dynamishelper.lua
```diff
@@ -1,4 +1,4 @@
---Copyright (c) 2013, Krizz
+--Copyright (c) 2013, Krizz, Skyrant
```
z16: Please change the `(c)` to `©`.

addons/dynamishelper/dynamishelper.lua
```diff
@@ -24,22 +24,427 @@
...
+defaults.pposy = 0
+
+-------------------------------------------------------------------------------
+-- Load defaults from settings.xml --------------------------------------------
+-------------------------------------------------------------------------------
+settings = config.load(defaults)
+
+-- upgrade to new settings file if old settings still exist -------------------
+if settings.trposx and settings.trposy then
+	-- copy position from old config to new one
```
z16: Throughout the file you're mixing spaces and tabs for indentation. Please use 4 spaces uniformly.

addons/dynamishelper/dynamishelper.lua
```diff
@@ -24,22 +24,427 @@
...
+        player = windower.ffxi.get_player().name
+        obtained = nil
+    end
+end)
+
+-------------------------------------------------------------------------------
+-- Parse the chat messages for drops, staggers and time extensions ------------
+-------------------------------------------------------------------------------
+windower.register_event('incoming text',function (original, new, color)
+	a,b,fiend = string.find(original,"%w+'s attack staggers the (%w+)%!")
```
z16: Actually, it seems like you never care about `a` or `b`, only for `fiend`. If that's the case, you should use `string.match` instead:
`local fiend = original:match('%w+'s attack staggers the (%w+)%!')`

You should only use `string.find` if you need the start and end index of the match.

And this goes for all the other `a,b,x` lines in this function as well.

addons/dynamishelper/dynamishelper.lua
```diff
@@ -24,22 +24,427 @@
...
+time_col = "\\cs(255,255,0)"
+
+StaggerCount = 0
+current_mob = "unknown"
+current_proc = "unknown"
+currenttime = 0
+obj_time = 0
+end_time = 0
+
+image = texts.new("image")
```
z16: I don't know if you actually want this, but if you do, you can make text object changes automatically save to the settings file. So every time you change the position for example, it will automatically save.

To do that, you need to call the constructor like this:
```lua
image = texts.new('image', settings, settings)
```

This will make the settings you currently have obsolete, you'd need to change them to the ones the text library uses. But in return you get many features for free, and the user can completely customize the text box if they want to. You can try it once and you'll see what settings it will generate. If you go that route I'd recommend putting the text settings into its own sub-table:
```lua
defaults = {}
defaults.image = {}

-- and later

image = texts.new('image', settings.image, settings)
```

So the first argument is the settings table for the image object, the second argument is for the root settings object.

addons/dynamishelper/dynamishelper.lua
Skyrant: fixed

addons/dynamishelper/dynamishelper.lua
Skyrant: investigating, is there some documentation for the texts library?

addons/dynamishelper/dynamishelper.lua
Skyrant: done

addons/dynamishelper/dynamishelper.lua
Skyrant: done

addons/dynamishelper/dynamishelper.lua
Skyrant: done

addons/dynamishelper/dynamishelper.lua
z16: Unfortunately no, but there's a short explanation of the [constructor](https://github.com/Windower/Lua/blob/dev/addons/libs/texts.lua#L129) in the Lua file itself. Check the link for more info.

addons/Treasury/Treasury.lua
```diff
@@ -103,7 +104,15 @@ function force_check()
...
-    if settings.AutoDrop then
+    if areas.Cities:contains(res.zones[windower.ffxi.get_info().zone].name) then
+        if settings.AutoDrop and settings.CityDrop then
+            for index, item in pairs(items.inventory) do
+                if type(item) == 'table' and code.drop:contains(item.id) then
+                    drop(item.id, index, item.count)
+                end
+            end
+        end
+    elseif settings.AutoDrop then
```
z16, line 115: This entire block you added can be merged with the previous block if you just change the condition:
```lua
    if settings.AutoDrop and (settings.CityDrop or not areas.Cities:contains(res.zones[windower.ffxi.get_info().zone].name)) then
```

addons/Treasury/Treasury.lua
```diff
@@ -183,8 +192,13 @@ windower.register_event('incoming chunk', function(id, data)
...
         end
-
-        if id == 0x020 and settings.AutoDrop and code.drop:contains(chunk.Item) then
+        if areas.Cities:contains(res.zones[windower.ffxi.get_info().zone].name) then
+            if id == 0x020 and settings.AutoDrop and settings.CityDrop and code.drop:contains(chunk.Item) then
+                drop(chunk.Item, chunk.Index, chunk.Count)
+            else
+                stack()
+            end
+        elseif id == 0x020 and settings.AutoDrop and code.drop:contains(chunk.Item) then
```
z16, line 201: Can be merged same as above:
```lua
        if id == 0x020 and settings.AutoDrop and code.drop:contains(chunk.Item) and (settings.CityDrop or not areas.Cities:contains(res.zones[windower.ffxi.get_info().zone].name)) then
```

addons/Treasury/Treasury.lua
```diff
@@ -276,6 +290,17 @@ windower.register_event('addon command', function(command1, command2, ...)
...
+    elseif command1 == 'citydrop' then
+        if command2 then
+            settings.CityDrop = bool_values[command2:lower()]
+        else
+            settings.CityDrop = not settings.CityDrop
+        end
+
+        config.save(settings)
+        log('CityDrop %s':format(settings.CityDrop and 'enabled' or 'disabled'))
+        --log('Current Zone: %s':format(res.zones[windower.ffxi.get_info().zone].name))
```
z16, line 302: I guess that was just for debugging, can be removed.

addons/Treasury/Treasury.lua
```diff
@@ -324,6 +350,35 @@ windower.register_event('addon command', function(command1, command2, ...)
     end
 end)
 
+areas = {}
+areas.Cities = S{
```
z16, line 354:
```lua
areas.cities
```

A common style please :)

z16 (changes requested):
The changes to Treasury are fine, although the checks can be consolidated, but that's optional. But please submit a different pull request from the DynamisHelper one. They need to be reviewed and handled separately, which is difficult if they're in the same request.

addons/dynamishelper/dynamishelper.lua
```diff
@@ -24,408 +24,597 @@
...
+window = texts.new(" ",settings.window,settings)
+w = T{}
+
+-------------------------------------------------------------------------------
+-- Initialize the Currency array. We need this to keep track of the drops -----
+-------------------------------------------------------------------------------
+function init_currency()
+    Currency = T{"Ordelle Bronzepiece","Montiont Silverpiece","One Byne Bill",
+                "One Hundred Byne Bill","Tukuku Whiteshell","Lungo-Nango Jadeshell",
+                "Forgotten Thought","Forgotten Hope","Forgotten Touch","Forgotten Journey","Forgotten Step"}
```
z16, line 83: Since you're using T-tables you can use their advanced iteration features:
```lua
    for c in Currency:it() do
        w[c] = 0
    end
```

Same goes for the `init_granules` function below.

addons/dynamishelper/dynamishelper.lua
```diff
@@ -24,408 +24,597 @@
...
+        obj_time = end_time - os.time()
+        w.time = os.date('!%H:%M:%S', obj_time)
+        window:update(w)
+    end
+end)
+-------------------------------------------------------------------------------
+-- Did we enter a Dynamis Zone? -----------------------------------------------
+-------------------------------------------------------------------------------
+windower.register_event('zone change', function(zone)
+    if ProcZones:contains(windower.ffxi.get_info().zone) then
```
z16, line 153: No need to call `windower.ffxi.get_info` here, just to get the zone; it's already a function argument.

addons/dynamishelper/dynamishelper.lua
```diff
@@ -24,408 +24,597 @@
...
+    end
+end
+init_window()
+
+-------------------------------------------------------------------------------
+-- Register a prerendere event for the display refresh ------------------------
+-------------------------------------------------------------------------------
+windower.register_event('prerender', function()
+    if obj_time < 1 then return end
+    if obj_time ~= (end_time - os.time()) then
```
z16, line 143: Can merge these conditions:
```lua
    if obj_time < 1 or obj_time == end_time - os.time() then
        return
    end

    --[[ rest of the function ]]
```

addons/dynamishelper/dynamishelper.lua
```diff
@@ -24,408 +24,597 @@
...
+-------------------------------------------------------------------------------
+-- The on screen window structure ---------------------------------------------
+-------------------------------------------------------------------------------
+function init_window()
+    local showCurrencyDivider = false
+    window.text(window,Yellow.."Time remaining: ${time|initializing...}")
+    window.appendline(window,"\\cr————————————————————")
+    window.appendline(window,"${current_mob|(unknown)}\n"..Green.."${current_proc|(none)}")
+    window.appendline(window,"\\cr————————————————————")
+    for i=1, #Currency do
```
z16, line 111: See above. As you're using the Currency table now it doesn't need to be a T-table at all.

addons/dynamishelper/dynamishelper.lua
```diff
@@ -24,408 +24,597 @@
...
+    end
 end
 
-windower.register_event('unload',function ()
- 	windower.text.delete('dynamis_box')
- 	windower.text.delete('proc_box')
+-------------------------------------------------------------------------------
+-- Process options and save settings ------------------------------------------
+-------------------------------------------------------------------------------
+windower.register_event('addon command',function (...)
```
z16, line 344: Instead of using `:lower()` dozens of times below, you can use it just once at the top by designing the function with a more fitting signature:
```lua
    local options = {...}
```

This will replace the first 7 lines of the function, and you can remove all `:lower()` calls on command below

z16 (commented):
I added the review for the DynamisHelper changes as well, but like I said before, I'd really appreciate if you could split it up into two pull requests.

PR #1522, opened 2017-04-19
New addon temps

Snapster-amb:
This addon accepts commands to buy temporary items from escha npcs.


addons/temps/temps.lua
```diff
@@ -0,0 +1,348 @@
...
+
+Redistribution and use in source and binary forms, with or without
+modification, are permitted provided that the following conditions are met:
+
+* Redistributions of source code must retain the above copyright
+notice, this list of conditions and the following disclaimer.
+* Redistributions in binary form must reproduce the above copyright
+notice, this list of conditions and the following disclaimer in the
+documentation and/or other materials provided with the distribution.
+* Neither the name of timestamp nor the
```
z16: Use `temps` here instead of `timestamp`.

addons/temps/temps.lua
```diff
@@ -0,0 +1,348 @@
...
+    else
+        notice("Enabling turbo.")
+        settings.turbo = true
+    end
+    settings:save()
+end
+    
+local function handle_command(...)
+    local cmd  = (...) and (...):lower()
+    local args = {select(2, ...)}
```
z16: You only use `args` to unpack it again later, so you could save yourself the effort by just doing `handlers[cmd](select(..., 2))` later (or, if you took my suggestion above, `handlers[cmd](...)`).

addons/temps/temps.lua
```diff
@@ -0,0 +1,348 @@
...
+        settings.turbo = false
+    else
+        notice("Enabling turbo.")
+        settings.turbo = true
+    end
+    settings:save()
+end
+    
+local function handle_command(...)
+    local cmd  = (...) and (...):lower()
```
z16: To save some effort here, you can change the function signature:
```lua
local function handle_command(cmd, ...)
```

That will parse the first argument for you.

You can also default to 'help' if it's not passed, which would be more user-friendly and save you the last case below:
```lua
    cmd = cmd and cmd:lower() or 'help'
```

z16 (changes requested):
Added some minor suggestions and some required changes to the copyright notice. It would also be nice to include a ReadMe.md file, which will be displayed if the user clicks on the "About" link on the addons tab.

addons/temps/temps.lua
z16: For some reason it removed my other comment... below it says "Giuliano Riccio", replace that with your name (either real or alias as above, preferably the same name as in the opening line).

joshk6656:
This is interesting, I recently researched this and I was pretty sure the NPC in Reisenjima did not work like other menu NPC's. Did you do any testing in that area? I'm not doubting but kind of baffled why I couldn't find this info. Mojo do you ever use IRC anymore, I have an unrelated question.

Snapster-amb:
I pushed a few commits addressing the feedback here.  I am rarely on IRC but you can usually message me on ffxiah or bg.

z16:
Ah crap.... now I pulled prematurely. One thing is still missing: an entry into the `addons.xml` file. Please provide that in a new pull request, otherwise the addon won't show up in the Windower launcher, so people won't be able to download it.

PR #1524, opened 2017-04-22
Nostrum version bump

trv6:


z16:
So I finally got around to "reviewing" it, but it's really shallow, since I feel it would take hours if not days of proper analysis to get into the details of how it works, so I won't go into the functionality of the addon. I'll trust your skill (and your tests, which are nice btw) to cover that. Instead I'll remark on a few general things that stood out. As for functionality, I only want to know this:
Will it break peoples' current *Nostrum* setup with this update? I saw some code that deletes old files, did you implement an upgrade process for users' settings? If so, did you make sure the upgrade process works right, so people don't end up with broken installations? Are the commands the same?

Note that breaking peoples' current setup is not necessarily a bad thing, sometimes breaking changes are needed for updates. I just want to know if that's the case so I know what to recommend. For example, if it can break peoples' macros or settings files I think you should display a message after loading it, so people know what to expect.

As I said before, I'd very much appreciate if you submitted the widget library as a proper library to our collection. But I saw you remarked it being "temporarily" in there, so I have some hope for it :)

As for the code, since I didn't look at everything I won't go into much detail, this is just what stood out when I glanced over it:

1. Why did you switch from `require('foo')` to `require 'foo'`? :(
2. I saw a few places where you left commented out code in there, especially throughout some files of the widget library. I'd go over those once more, see if some of them can be deleted if no longer needed, as things like that tend to pile up during debugging.
3. I saw you implemented your own packet parsing mechanism for things we already provide (such as buffs and some player stats). Is there a reason for that?
4. You set the addon's language to the client's language, which is (or *should* be) unnecessary. The addon's language is used to look up things in resources, which you effectively reimplemented manually here. Instead of `resources.english` or `resources.japanese` you can just do `resources.name`, which will go by `_addon.language`, if provided, otherwise by `windower.ffxi.get_info().language`. The only reason an addon can specify a language is so it can hardcode values and look them up by using `name`, which is generally not a good idea and prevents good localization. So you should either specify a language directly, or not use one at all, in your case it has no practical effect. Except for the bug that you lowercase it, but later compare it with `'Japanese'`, which will always be `false`. Can't blame you for not noticing though, since the `if` and `else` branch of that check are identical :)
5. Is the sandboxing feature so people can write their own plugins for it? Like the *MsJans* thing? Currently that does not reside inside the `data` folder, so people will have a number of issues writing code for it. If it's meant to be user-writable for the addon, I would move it into the `data` folder, if it's something you mean to redistribute yourself through PRs to the addon, this is the way to go. Of course you can also have a mix of the two, by looking for plugins in different directories.
6. There are a bunch of very minor things to remark which are more of style irregularities rather than issues (missaligned comments, stray variables, etc.), but since it's a pain to work on such a large PR, I'll leave that out for now. If you're curious you can peruse the PR yourself and see if there's something worth changing to you, but I didn't find anything in particular that was worth mentioning separately.

trv6:
Thanks! Feedback of any depth helps-- I really appreciate it. 
I'm mostly limited to my phone for the next two days. I hope you don't mind if I leave covering the numbered items for when I'm home free on Wednesday.

Regarding the settings files and breaking a user's setup: yes and no. I don't think I wrote anything to handle converting settings related to the visual aspects of the display (color, fonts, etc.), but that block of code does convert the spell/ability part of the old settings. As a backup, the old settings file is renamed rather than deleted.
The new settings file allows for any spell/ability in the resources file to be added to the display, whereas the old only allowed a small list. This is the file that was output when I tested that block of code: https://gist.github.com/trv6/8bace733c598a4da05d4377e1ee49fa8 It uses the same format as the _Timers_ plugin (pipes!). It's not ideal: adding a new spell involves editing three lines in the settings (the spell, label, and icon must be specified in the position they'll appear in the display), but it's certainly not worse than the old file.

There is a small message printed to the console after the conversion that mentions an update. I can probably make it more detailed.

The commands are mostly the same (I think? I haven't looked in a while).

In short, a user _should_ (fingers crossed) be able to load the addon after the update and get back to where they were with only some mild annoyance.

Regarding the widgets library: Parts of it are... not great.

z16:
Most of our libraries are barely functional, and some not even that :) I can't force you to, just saying I'd appreciate it.

And no rush, I waited long with my review, can't blame you for waiting a few days :)

trv6:
1) I remain enamored with that syntax. It is swell. (b‘ .’)b

2) I’ll search through them. You mention the widgets library, though, which might be an exception.
	
3) The bit library route in the packet parsing table is probably to grab one piece of information out of a fairly large packet. In the case of buffs, I think I didn't want to iterate through the string indices:
	```
	for i = 1, 32 do
		local buff = packet['Buff ' .. tostring(i)]
	end
	```

4) I think the only thing the language is used for (apart from grabbing the correct spell/ability string) is deciding whether or not to surround the spell/ability in quotes when inputing a command. Nothing in the Japanese resources contained a space, so I wasn't sure if the client would even parse quotes.
	
5) _Nostrum_ is only an interface. It wraps a few of windower’s events and creates a bunch for party related info. Overlay files (I lacked a better word?) can register functions to the events to create different displays and controls.
    The _MsJans_ files are an overlay recreation of the previous versions of _Nostrum_ (originally a clone of the standalone MsJans program) with a few extra options. I don’t think I included it in this pull request, but there’s an unfinished clone of _Greed_ as well.

- [ ] `{}.name` vs `_addon.language`
- [ ] settings to Lua
- [ ] terrible widgets
- [ ] Monstrosity via `player.job == 'Monipulator'` -- check before BLU

PR #1525, opened 2017-04-27
Added xivbar addon

SirEdeonX:
This addon displays vital bars for easy tracking.


addons/xivbar/xivbar.lua
```diff
@@ -0,0 +1,49 @@
...
+_addon.name = 'XIV Bar'
+_addon.author = 'Edeon'
+_addon.version = '1.0'
+_addon.commands = {'xivbar'}
+_addon.language = 'english'
+
+-- Windower Libs
+config = require('config')
+texts  = require('libs/non-draggable-texts')
+images = require('libs/images')
```
trv6: You've included a copy of windower's texts and images libraries in the directory of the addon, 
each with a minor change that switches the default setting for the dragging behavior to false.
It would be easier to require the unedited libraries and define your settings so that your
objects are not draggable.
This text object will not be draggable:
```
texts.new({
	flags = {
		draggable = false,
	},
})
```
I believe the images library has the same settings structure.

trv6: You also found an error in the images library, which you could open a separate pull request for.

The first line of images.draggable should read:
`if drag == nil then ....`

addons/xivbar/events.lua
```diff
@@ -0,0 +1,92 @@
...
+
+windower.register_event('mpmax change', function(new, old)
+    mp_update = true
+end)
+
+windower.register_event('tp change', function(new, old)
+    tp_update = true
+end)
+
+windower.register_event('prerender', function()
```
trv6: The function windower.ffxi.get_player is among the most expensive functions in windower's API.
As written, your addon will potentially call it ninety times per second (180 if the player's
framerate is set to 60) for information that is available to you elsewhere. 

Would it not be more simple to split the &lt;stat&gt;_update functions for HP and MP into two functions-- one for
updating the text, and one for updating the bar? You can use the '<hpp/mpp> change' events for
the bars, and only update the text objects in the '<hp/mp> change' events. This would negate your need for the 'hpmax change' event. Also, you could remove the &lt;stat&gt;_update variables, as both your old and new  &lt;stat&gt; are available within the scope of the events.

addons/xivbar/variables.lua
```diff
@@ -0,0 +1,97 @@
...
+-- setup variables
+background = images.new()
+
+hp_foreground = images.new()
+mp_foreground = images.new()
+tp_foreground = images.new()
+
+hp_text = texts.new(settings.Texts.Text)
+mp_text = texts.new(settings.Texts.Text)
+tp_text = texts.new(settings.Texts.Text)
```
trv6: settings.Texts.Text is not defined anywhere.

addons/xivbar/xivbar.lua
```diff
@@ -0,0 +1,49 @@
...
+        ON ANY THEORY OF LIABILITY, WHETHER IN CONTRACT, STRICT LIABILITY, OR TORT
+        (INCLUDING NEGLIGENCE OR OTHERWISE) ARISING IN ANY WAY OUT OF THE USE OF THIS
+        SOFTWARE, EVEN IF ADVISED OF THE POSSIBILITY OF SUCH DAMAGE.
+]]
+
+-- Addon Description
+_addon.name = 'XIV Bar'
+_addon.author = 'Edeon'
+_addon.version = '1.0'
+_addon.commands = {'xivbar'}
```
trv6: Your addon has no commands.

addons/xivbar/main.lua
```diff
@@ -0,0 +1,231 @@
...
+    hp_foreground:show()
+    hp_text:show()
+    mp_foreground:show()
+    mp_text:show()
+    tp_foreground:show()
+    tp_text:show()
+    ready = true
+end
+
+function initialize()
```
trv6: This function is called every time you log in, but really only needs to be called once.

addons/xivbar/main.lua
```diff
@@ -0,0 +1,231 @@
...
+        elseif old_hp_width == new_hp_width then
+            if new_hp_width == 0 then
+                hp_foreground:hide()
+            end
+            hp_update = false
+        end
+    end
+
+    hp_text:clear()
+    hp_text:append('' .. info.vitals.hp)
```
trv6: The tostring function will convert a number to a string.

trv6 (commented):
I am not Arcon, and I imagine he'll give you more thorough feedback, but I was stuck in a train
station and did a quick lookover.

You nailed the style of the xi HP/MP/TP bars.

addons/xivbar/main.lua
SirEdeonX: Thank you so much for your insights! As my first addon (and first time messing around with Lua), they're invaluable! I'll make the necessary changes :)

addons/xivbar/events.lua
Byrth: For the record, it's get_party() that is very expensive.

get_player() is comparably quite light.

trv6: I timed a few of the functions a while ago. I think I looped them 5,000 times.

```
function               time(s)
get_items (no args):   6.377
get_player:            0.677
get_items(0):          0.483
get_party (6 players): 0.322
get_party (1 players): 0.167
get_mob_by_id:         0.035 (full zone)
get_mob_by_id:         0.030 (empty zone)
get_mob_by_index:      0.028
```

z16: Just to add to that, `get_party` scales with party members (as was observed), but generally mostly pushed the mob array entry for each party member, so it's basically `<some overhead> + get_mob_by_index * <number of party members>`. `get_player` on the other hand pushes a lot that's not immediately obvious, like all merits, all job points (which are each many tables with subentries), buffs, skills, vitals, jobs, etc., all of which add up to quite a bit. Still, as is obvious from the timing, `get_items` trumps them all, which is why I added the specific overloads for specific bags/slots as well as the `add item` and `remove item` events, to not have to constantly query all bags to monitor changes.

As a side note, the performance of `get_mob_by_id` depends on the type of the ID. If it's an NPC ID with a non-zero first byte, it just falls back on `get_mob_by_index`. If it's a player ID it iterates over the player index range and compares IDs. However, empty zones are not necessarily faster in that regard, as indices cycle, so it's just a matter of luck.

addons/xivbar/defaults.lua
```diff
@@ -0,0 +1,77 @@
...
+defaults.Images.FFXIV.BackgroundCompact = windower.addon_path..'resources/ffxiv_bar_compact.png'
+defaults.Images.FFXIV.Hp = windower.addon_path..'resources/ffxiv_hp_fg.png'
+defaults.Images.FFXIV.Mp = windower.addon_path..'resources/ffxiv_mp_fg.png'
+defaults.Images.FFXIV.Tp = windower.addon_path..'resources/ffxiv_tp_fg.png'
+defaults.Images.FFXI = {}
+defaults.Images.FFXI.Background = windower.addon_path..'resources/ffxi_bar_bg.png'
+defaults.Images.FFXI.BackgroundCompact = windower.addon_path..'resources/ffxi_bar_compact.png'
+defaults.Images.FFXI.Hp = windower.addon_path..'resources/ffxi_hp_fg.png'
+defaults.Images.FFXI.Mp = windower.addon_path..'resources/ffxi_mp_fg.png'
+defaults.Images.FFXI.Tp = windower.addon_path..'resources/ffxi_tp_fg.png'
```
z16: Using `windower.addon_path` in the default settings table will mean that the user won't be able to copy their settings file without manually adjusting the paths, since those will be written to the settings file. They also won't be able to share settings files with others, and what's worse, it may reveal details of their filesystem structure to others without their intent. I would change this to just hold relative paths, and then prepend `windower.addon_path` when used, or after parsing the settings (see also `config.register` for the latter).

addons/xivbar/events.lua
```diff
@@ -0,0 +1,109 @@
...
+-- ON LOAD
+windower.register_event('load', function()
+    if windower.ffxi.get_info().logged_in then
+        initialize()
+        show()
+    end
+end)
+
+-- ON LOGIN
+windower.register_event('login', function()
```
z16: This can be shortened:
```lua
windower.register_event('login', show)
windower.register_event('logout', hide)
```

z16 (changes requested):
Aside from the paths in the default settings everything looks pretty good, surprisingly little I would change. There are a few design choices that I'd make differently, but they're just that - choices. I'll still mention them so you can evaluate the reasons and decide if you wanna do it or not, but it's up to you.

This only regards the structure of your program. You `require('main')` in the end, which contains the initial execution logic of your program. I would instead paste all of that in your entry file (`xivbar.lua`), and something similar for `variables.lua`/`events.lua`.

The reason to package things into Lua modules is to make them modular, but the way your program is right now is not modular. You're using different files just to differentiate parts of your code. Which, again, is not wrong, it's just a different style than what's commonly used.

Here are some things I'd change (aside from just appending all of `main.lua` onto `xivbar.lua`):

Parsing settings in a different file is fine, I'd keep that (especially since your settings are rather large). But instead of executing it by using `require` I'd make it local inside the file and then return it.

`defaults.lua` would then look like this:
```lua
--[[ copyright notice ]]

local defaults = {}

--[[ This stays the same ]]

return defaults
```

Now you have a proper module. The file won't write into the global namespace, but just prepare the default settings table and return it, and the caller can then do with it as they please:

```lua
-- load settings
defaults = require('defaults')
config.save(settings)
```

As for `variables.lua` and `events.lua`, you could do something similar, but I think Lua makes that a bit clumsy. Right now a person reading your code would see a variable `font_size` in `main.lua`, but wouldn't see where it's from. It's not defined in the file, and the file also does not require any other files. That means, the file relies on it being called after `variables.lua` is loaded. Which it is right now, but that's not very modular.

Lua has the `require` function for this. It also assures that a file is never included twice. So you can do `require('variables')` both in `main.lua` and again in `events.lua`. Then it doesn't matter if `events.lua` is included before or after `variables.lua`, it will always have the variables available. And a curious reader will know where to look if they need to find something.

Finally, the reason for this split is to separate declarations and logic. To that end you could do something else which some people choose to do: define events in `events.lua`, but do not register them. Instead register them later in the main file.

`events.lua` would look something like this:
```lua
local events = {}

events.load = function() --[[ stuff ]] end
events.login = function(name) --[[ stuff ]] end
--[[ and so on ]]

return events
```

`xivbar.lua` or `main.lua`, whichever you go with in the end, would receive this table of functions and register the appropriate events:
```lua
events = require('events')

windower.register_event('load', events.load)
windower.register_event('login', events.login)
--[[ and so on ]]
```

This has the slight advantage that you can combine functions for event hookup, which you're not doing right now, but which you could in the future:
```lua
windower.register_event('load', events.load, events.initialize)
windower.register_event('hp change', events.hp_change, events.show)
```

So it would allow you to design your functions themselves in a more modular way. Instead of one function for `load`, one for `login` and so on, you'd have simple functions doing simple things, then combine them as you see fit when registering events. Again, you don't actually need this in your code (your individual functions are very simple already, which is a good thing), but just to give an outlook on what can be done with this.

But again, all of this is optional, though I'd strongly urge you to consider removing paths from default settings. As a rule of thumb, the default settings should never rely on anything user-specific. As for the rest, let me know if you want to give it a shot or if you want us to merge it as it is now.

addons/xivbar/events.lua
SirEdeonX: Thank you for your awesome feeback! As this was my first time messing with Lua and creating an Addon, I struggled to find a structure and went mostly by example, but I'll go ahead and change a couple of things.

z16:
Oh and one more thing I noticed, you forgot to make an entry in the `addons.xml` file. If you don't do that this addon will not show up in our launcher. It's in the root directory of this repositry and you can copy the structure from any of the other addons, just replace with values of your own.

addons/xivbar/player.lua
```diff
@@ -0,0 +1,50 @@
...
+player.mpp = 0
+player.tpp = 0
+player.current_hp = 0
+player.current_mp = 0
+player.current_tp = 0
+
+function player:calculate_tpp()
+    self.tpp = self.current_tp
+
+    if self.current_tp ~= 0 then
```
z16: This check is unnecessary. If it is `0`, `self.tpp` will just be assigned `0 / 10`, which is again `0`.

And if you remove that if check, you may as well remove the first line. And if you want to use an additional Lua function the entire function could be:
```lua
function player:calculate_tpp()
    self.tpp = math.min(self.current_tp / 10, 100)
end
```

addons/xivbar/xivbar.lua
```diff
@@ -0,0 +1,216 @@
...
+            elseif flag == 2 then
+                xivbar.update_mp = false
+            elseif flag == 3 then
+                xivbar.update_tp = false
+            end
+        else
+            local x = old_width
+
+            if old_width < new_width then
+                x = old_width + math.ceil(((new_width - old_width) * 0.1))
```
z16: There are extra parentheses around this and the lower expression... yeah, that's the best kind of criticism I can come up with now :D

addons/xivbar/xivbar.lua
```diff
@@ -0,0 +1,216 @@
...
+
+                if x > theme_options.bar_width then
+                    x = theme_options.bar_width
+                end
```
z16: If you want to limit values to a certain lower bound, you can use `math.max`, and `math.min` for an upper bound. Can either do it like this:

```lua
            x = math.max(x, theme_options.bar_width)
```

or combine it into the upper expression, though some would find that's a bit too much to read:

```lua
            x = math.max(old_width + math.ceil((new_width - old_width) * 0.1), theme_options.bar_width)
```

You can do the same thing below with `if x < 0 then`, with `math.min`.

z16 (approved):
I added some minor optional adjustments you can make but overall it looks good and I'd be ready to merge. Let me know if you want to adjust something still or if you want me to merge it as it is.

SirEdeonX:
Thank you again for all your feedback. I made some last adjustments and I think it's ready to be merged :D

PR #1527, opened 2017-05-01
Patch 1

Kinkles:


addons/findAll/findAll.lua
```diff
@@ -48,6 +48,7 @@ res = require('resources')
 defaults = {}
 defaults.Track = ''
 defaults.Tracker = {}
+defaults.Tracker.Key_Item_Display = true
```
z16: Is the key item display related to the tracker? If not, I'd move the setting out of that table. Also please remove the underscores from settings names:

```lua
defaults.KeyItemDisplay = true
```

addons/findAll/findAll.lua
```diff
@@ -282,7 +283,7 @@ function search(query, export)
     end
     
     for id, names in pairs(key_item_names) do
-        if terms_pattern == '' or key_item_names[id].name:find(terms_pattern) then
+        if terms_pattern == '' or key_item_names[id].name:find(terms_pattern) and settings.Tracker.Key_Item_Display == true then
```
z16: While this works, the loop as it is is inefficient. I'd rewrite the entire loop to this:

```lua
if settings.KeyitemDisplay then
    for id in pairs(key_item_names) do
        if terms_pattern == '' or key_item_names[id].name:match(terms_pattern) then
            results_key_items:add(id)
        end
    end
end
```

PR #1533, opened 2017-05-07
New Addon: capetrader

OdinBurntwaffle:
Hello there. Cape Trader was designed to automate the process of augmenting ambuscade capes.

I currently have two concerns about the capetrader addon. If possible, for one concern I could use some suggestions on how to go about implementing a solution if you think it would be worthwhile.

1. It is currently possible for a user of capetrader to lose their their augment items if they try to augment a cape with a different augment path than is already on their cape. For example trying to augment a cape with a STR path when the DEX path has already been augmented into the cape. It is possible for the user to avoid this by using the //ct prep command carefully. I've tried to highlight this possibility in the README, but the possibility still worries me. If you all think a safeguard for this would be a good idea to add let me know. Suggestions on how to implement such a safeguard would be appreciated as well, I'm still not sure how I would go about it at the moment.

2. There is currently a bug relating to augmenting a cape with dye for the first time. I have implemented what I think will fix this by increasing the trade delay for dye items but I'm not entirely sure this will fix the problem. Unfortunately I don't think I can test the fix until I do some ambuscade after the upcoming May update.

If you have any concerns that I missed please let me know and I will try to address them.

This is my fist time working with git fork/pull requests so I hope I didn't mess it up. Also I'd like to use this opportunity to thank you all for the work you do on Windower.


Byrth:
Haven't really looked at the code, but you need to edit addons.xml and Sync so the changes are added to your pull request. Just replicate an existing entry.

OdinBurntwaffle:
Hello again. I think my latest push resolves both of my own initial concerns that I raised above. I tested the fix for the augmenting with dye bug and it seemed to work, but I wonder if that problem may have just been random packet loss on my end. 

Just to make sure, it is the dev branch I need to stay in sync with right? I suspect I will have to keep an eye out for updates after this upcoming May update. Thanks for pointing out I needed to edit the addons.xml file.

z16:
Any particular reason why this was closed? I was in the middle of reviewing it. Do you want me to finish?

OdinBurntwaffle:
Doh! I just made a new pull request, made a few changes to the readme and to some variable names from this one. Sorry about that.

z16:
No problem! You don't need to make a new PR for that btw, any changes you commit and push to the same branch will automatically be added to the pull request.

PR #1535, opened 2017-05-08
Adding custom Addons.  Fixing up packets\fields.lua

SammehFFXI:
libs\packets\fields.lua - Fixed 0x03c to reflect Craft Skill and craft
Rank.
Addons added:
Sleeper: Bard Lullaby Tracker
Snub: Cleanup Chat Log for things like treasury drop/pass items; exp;
etc.
JobChange: command line Job Change for fast reset/job change.
DistancePlus:  Mods the distance plugin to add in distance to pet, and
color coordiantes JA's and ranged attacks and magic casting.
PetCharges: Displays information aobut pet charges / timers remaining on
screen.


z16:
I know this will be more work, but I'd like to ask you to open individual pull requests for each addon and the libs change. That will make life much easier for us, we may be ok with one addon, but not with another, or one may be fine while another one needs adjustments. It's just much easier to review if you did it that way. If you're not yet familiar with git, you should look up how branching works, which actually makes that quite simple to do.

SammehFFXI:
I kind of assumed this would be the case but had no idea etiquette.  I'm not a real git user nor a real developer.. .tinker mostly.  I'll submit individuals.

SammehFFXI:
Pulling my hair out on this one.  I re-submitted then tried to add another one.  Not sure it did right.  Let me know. I apologize if wrong again.

cairthenn:
What you want to do is create a branch for each addon locally (so instead of SammehFFXI:dev, SammehFFXI:DistancePlus) with just that addons changes in dif, and submit a pull request from that branch. You can use git shell to create a new branch or do it on the website within your actual repository.

SammehFFXI:
ok - Well kill my next pull request haha :(    I'll try that way again! Sorry!

PR #1536, opened 2017-05-09
TH Tracker Update

tehkrizz:
TH Tracker API update and Battlemod compatibility


tehkrizz:
Hmm, what I pulled and modified isn't equal to what it is comparing against. Going to cancel and review.

addons/thtracker/thtracker.lua
```diff
@@ -24,70 +24,162 @@
...
+local posy = 200
+local see = true
+local name = nil
+local box_name = 'addon:gr:thbox'
+mob = nil
+count = 0
+
+local settingtab = nil
+local settings_file = 'data\\settings.xml'
+local settingtab = config.load(settings_file)
```
cairthenn, line 48: Switching to requiring the user to have the settings XML to create defaults probably isn't the way to go, and I'm pretty sure the config library isn't actually intending for load to just take a string path. What's the reason for this change?

addons/thtracker/thtracker.lua
```diff
@@ -24,70 +24,162 @@
...
+			a,b,name,count = string.find(original,'Additional effect: Treasure Hunter effectiveness against the (.*)%s? increases to (%d+).')
+			if name == nil then
+				--log('No name')
+				a,b,name,count = string.find(original,'Additional effect: Treasure Hunter effectiveness against (.*)%s? increases to (%d+).')
+			end
+		end
+		
+		if string.find(original, "defeats") and not string.find(original, "TH Tracker:") then
+			a,b,deadmob = string.find(original,'%w+ defeats the (.*)\46')
+			if deadmob == nill then
```
cairthenn, line 164: You probably want "nil" here

addons/thtracker/thtracker.lua
```diff
@@ -24,70 +24,162 @@
 --(INCLUDING NEGLIGENCE OR OTHERWISE) ARISING IN ANY WAY OUT OF THE USE OF THIS
 --SOFTWARE, EVEN IF ADVISED OF THE POSSIBILITY OF SUCH DAMAGE.
 
-
-_addon.name = 'THTracker'
+_addon = _addon or {}
+_addon.name = 'TH Tracker'
 _addon.author = 'Krizz'
-_addon.version = 1.0
-_addon.commands = {'thtracker', 'th'}
```
cairthenn, line 31: You should keep the commands and update the version.

addons/thtracker/thtracker.lua
```diff
@@ -24,70 +24,162 @@
...
+function initialize()
+	--[[windower.text.create(box_name)
+	windower.text.set_bg_color(box_name,200,30,30,30)
+	windower.text.set_color(box_name,255,200,200,200)
+	windower.text.set_location(box_name,posx,posy)
+	windower.text.set_visibility(box_name, see)
+	windower.text.set_bg_visibility(box_name,1)
+	windower.text.set_font(box_name,'Arial',12)
+	windower.text.set_text(box_name, thoutput)]]
+	--log('Initializing TH Tracker')
```
cairthenn, line 67: You can probably remove this commented out code!

addons/thtracker/thtracker.lua
```diff
@@ -24,70 +24,162 @@
...
+    windower.text.delete(box_name)
+end
+
+function thunload()
+	dispose()
+	windower.send_command('unalias th')
+end 
+
+windower.register_event('load', function(...)
+	windower.send_command('alias th lua c thtracker')
```
cairthenn, line 103: You won't need this if you leave the commands there at the top :)

addons/thtracker/thtracker.lua
```diff
@@ -24,70 +24,162 @@
...
+		elseif params[1]:lower() == "hide" then
+			see = false
+			thbox:hide()
+		elseif params[1]:lower() == "show" then
+			see = true
+			thbox:show()
+		elseif params[1]:lower() == "current" then
+			print('Box text should be: '..thoutput)
+		elseif params[1]:lower() == "delete" then
+			--windower.text.delete(box_name)
```
cairthenn, line 142: I'm guessing this and create were left in for debug purposes? You can probably remove delete altogether. No reason for the user to be deleting the box anyway!

addons/thtracker/thtracker.lua
```diff
@@ -24,70 +24,162 @@
...
+	--[[windower.text.create(box_name)
+	windower.text.set_bg_color(box_name,200,30,30,30)
+	windower.text.set_color(box_name,255,200,200,200)
+	windower.text.set_location(box_name,posx,posy)
+	windower.text.set_visibility(box_name, see)
+	windower.text.set_bg_visibility(box_name,1)
+	windower.text.set_font(box_name,'Arial',12)
+	windower.text.set_text(box_name, thoutput)]]
+	--log('Initializing TH Tracker')
+	t = texts.new('test text box')
```
cairthenn, line 68: Stray textbox!

cairthenn (commented):
The only major concern I have is that you didn't really switch away from the config library, instead you're using it an a very nonstandard way. I'd recommend leaving the defaults in lua and doing:

settingstab = config.load(defaults)

And then for your textbox you should be able to do:

thbox = texts.new('${boxtext}', settingstab)

cairthenn:
Didn't realize you closed this, sorry!

tehkrizz:
I appreciate the feedback regardless.

addons/thtracker/thtracker.lua
z16: Just to weigh on this, the ability to create and manage default settings was added onto the `config` library a while after it already existed. Originally it loaded whatever file it found at the provided path, which is why originally many addons shipped with existing settings files.

And you can use the config library for multiple files, it merely defaults to `data/settings.xml`.

So this usage is not really wrong, only outdated. It defaults to that file anyway, so no use specifying it. Also no need to set `settingtab` to `nil` beforehand. And instead of providing a settings file, you should use a default settings table and pass it to `config.load`. Most addons use it in that way as well, so you can check any one of them if you need examples.

PR #1537, opened 2017-05-09
TH Tracker Update

tehkrizz:
TH Tracker Battlemod compatibility.


addons/thtracker/thtracker.lua
```diff
@@ -24,162 +24,83 @@
...
-
-
-windower.register_event('addon command', function(...)
-	local params = {...};
-	if #params < 1 then
-		return
+windower.register_event('incoming text', function(original, new, color)
+	original = original:strip_format()
+	if string.find(original, "Treasure Hunter") and not string.find(original, "TH Tracker:") then
+		name, count = original:match('Additional effect: Treasure Hunter effectiveness against the (.*) increases to (%d+)')
```
cairthenn: I'd recommend "local name, count" so your variables don't exist in the global namespace. Name and count are very generic, so (future) issues could easily arise.

addons/thtracker/thtracker.lua
```diff
@@ -24,162 +24,83 @@
...
+    if name and count then
+        mob = name
+        th:text(' '..name..'\n TH: '..count);
+        th:show()
+		mob = nil
+		deadmob = nil
+		count = nil
+    end
+
+    deadmob = original:match('%w+ defeats the (.*).')
```
cairthenn: Just another local variable change for you..

`local deadmob = original:match('%w+ defeats the (.*).') or original:match('%w+ defeats (.*).')`

addons/thtracker/thtracker.lua
```diff
@@ -24,162 +24,83 @@
...
-			deadmob = nil
-		end
+	
+    if name and count then
+        mob = name
+        th:text(' '..name..'\n TH: '..count);
+        th:show()
+		mob = nil
+		deadmob = nil
+		count = nil
```
cairthenn: You can probably delete the following entirely:

```
mob = name

mob = nil
deadmob = nil
count = nil
```

cairthenn (commented):
There are just a few minor changes regarding variable scope I think you should make before this gets merged.

I think the core functionality of the addon could use an overhaul since the text matching makes no guarantees which mob is having their TH upgraded or being killed, but since you aren't changing that from how it was, it's outside the scope of this review.

addons/thtracker/thtracker.lua
tehkrizz: I'm actually in the middle of correcting the issue with the mob being killed at least having the same name. I want to eventually do better matching to support multiple targets, but that is beyond me right now. I should have the scopes corrected shortly.

addons/thtracker/data/settings.xml
```diff
@@ -1,9 +1,50 @@
...
+            <stroke>
+                <alpha>255</alpha>
+                <blue>0</blue>
+                <green>0</green>
+                <red>0</red>
+                <width>0</width>
+            </stroke>
+        </text>
+    </global>
 </settings>
```
z16: You should remove the settings file entirely. See my comments on the outdated PR.

tehkrizz:
Just commited another change.

Line 54 gives it text on init so the 'th show' command has something to display after loading.

The message checks are now in different statements with returns because they were conflicting with each other at one point. I tried to make the variables local (mob is really the only one that needs to be global), but it completely stopped working. I'm open to suggestions.

addons/thtracker/thtracker.lua
```diff
@@ -61,33 +62,51 @@ windower.register_event('addon command', function(command, ...)
...
-        th:show()
-    end
-
-    name = original:match('%w+ defeats the (.*).')
-    if name and mob == name then
-        th:text('')
-        th:hide()
-    end
+	original = original:strip_format()
+	if string.find(original, "Treasure Hunter") then
```
z16: The `string.find` here is not saving much in terms of efficiency, compared to the latter match. And since the latter match is much more precise, I'd remove this if check entirely, along with the `return` at the end of this inner scope.

addons/thtracker/thtracker.lua
```diff
@@ -61,33 +62,51 @@ windower.register_event('addon command', function(command, ...)
...
+		
+		if name and count then
+			mob = name
+			th:text(' '..name..'\n TH: '..count);
+			th:show()
+		end
+		return
+	end
+	
+	if string.find(original, "defeats") then
```
z16: Same as above, this check can be removed completely

addons/thtracker/thtracker.lua
```diff
@@ -61,33 +62,51 @@ windower.register_event('addon command', function(command, ...)
...
-    if name and mob == name then
-        th:text('')
-        th:hide()
-    end
+	original = original:strip_format()
+	if string.find(original, "Treasure Hunter") then
+		name, count = original:match('Additional effect: Treasure Hunter effectiveness against the (.*) increases to (%d+).')
+		if not name and not count then
+			name, count = original:match('Additional effect: Treasure Hunter effectiveness against (.*) increases to (%d+).')
+		end
```
z16: While it would be more expensive, you could consider using a regex match for this. It would save you the double match and if check, so you'd effectively reduce this code (as well as the lower part) to one line.

addons/thtracker/thtracker.lua
```diff
@@ -27,11 +27,12 @@
```
z16, line 27: You should update the copyright notice for the current year, as well as replace the `(c)` with `©`.

z16 (approved):
Added some suggestions as well. All rather minor, overall this looks good to me.

addons/thtracker/thtracker.lua
tehkrizz: That was my original goal, but I have issues with Lua regex. I'll investigate further.

tehkrizz: I've managed to get it to find it with or without the string, but it adds 'the ' before the mob's name in the variable. Not the desired result. I'm currently using:
`against[%s%a%a%a]- (.*) increases`

I can't seem to wrap my head around how to ignore the 'the'.

z16: You can't do that with Lua patterns, sadly, so either you keep it the way it's in this PR, or you change to regex:
```lua
name, count = select(2, unpack(windower.regex.match(original, 'Additional effect: Treasure Hunter effectiveness against (?:the)? (.*) increases to (\\d+).')[1]))
```

The syntax is a bit more verbose, since `windower.regex.match` returns a different structure for all matches and packs it into a table, so it looks like this.

tehkrizz: The windower.regex was exactly what I was looking for (obviously didn't look hard enough). That will be so much better as it should be similar, if not the same, to what I use for JavaScript. Thanks.

PR #1540, opened 2017-05-11
DistancePlus Addon

SammehFFXI:
Distance Plus added -  Provides an updated version of Distance which
correlates between model sizes for more accurate off-sets of casting
range, shooting ranges, etc.  Will color coordiante for ranged attacks,
magic, and Job abilities.  When a pet is out - will add a
distance-from-pet which is also color coordinated for Ready/Reward
charges etc.


addons/DistancePlus/DistancePlus.lua
```diff
@@ -0,0 +1,392 @@
...
+heighttxt.flags.right = true
+height_upper_threshold = 8.5
+height_lower_threshold = -7.5
+
+distance = texts.new('${value||%.2f}', defaults)
+petdistance = texts.new('${value||%.2f}', pettxt)
+abilities = texts.new('${value}', abilitytxt)
+height = texts.new('${value||%.2f}', heighttxt)
+
+debug.setmetatable(nil, {__index = {}, __call = functions.empty})
```
z16: Are you aware what this line does? It was really a hack I added to save two lines of code in *Distance*, and I'm even regretting using it there. I'd recommend removing it and properly checking your code for `nil` instead.

addons/DistancePlus/DistancePlus.lua
```diff
@@ -0,0 +1,392 @@
...
+
+option = "Default"
+showabilities = false
+showheight = false
```
z16: These seem like settings, but it doesn't look like they can be adjusted by the user. You should consider using the `config` library to manage user settings.

addons/DistancePlus/DistancePlus.lua
```diff
@@ -0,0 +1,392 @@
...
+
+self = windower.ffxi.get_player()
```
z16: You set this variable but don't always use it. It's also not updated when a player relogs. You should set it in a `load` and `login` event.

addons/DistancePlus/DistancePlus.lua
```diff
@@ -0,0 +1,392 @@
...
+
+
+function displayabilities(distance,master_pet_distance)
```
z16: This function is potentially called on every `prerender` event, which means up to hundreds of times each second. In the very first lines it also calls some expensive functions itself. It's overall very inefficient, and for what it does it doesn't need to be called so often, but only on job change.

addons/DistancePlus/DistancePlus.lua
```diff
@@ -0,0 +1,392 @@
...
+  abilitylistrecast = windower.ffxi.get_ability_recasts()
```
z16: You're using global variables for almost everything. To easier debug these things you should make it a habit to use `local` wherever it makes sense (which is almost everywhere).

addons/DistancePlus/DistancePlus.lua
```diff
@@ -0,0 +1,392 @@
...
+	else
+		windower.add_to_chat(8,'Mode: Default.')
+	    option = "Default"
+	    MaxDistance = 25
+	end
+end
+
+
+windower.register_event('prerender', function()
+    t = windower.ffxi.get_mob_by_index(windower.ffxi.get_player().target_index or 0)
```
z16: `windower.ffxi.get_player` is an expensive function, and you call it twice in each `prerender` event, despite having an unused `self` variable you could use. And you don't even need that, should rewrite it to:
```lua
    t = windower.ffxi.get_mob_by_target('t') or windower.ffxi.get_mob_by_target('st')`
    s = windower.ffxi.get_mob_by_target('me')
    pet = windower.ffxi.get_mob_by_target('pet')
```

The last call you make to get the pet is especially puzzling. You get the pet mob, then get its index and get the same object again by calling another expensive function. You should keep in mind that none of this is free and you can save yourself a lot of effort when doing it right.

addons/DistancePlus/ja_distance.lua
```diff
@@ -0,0 +1,110 @@
+-- Automatically generated file: Job Abilities
```
z16: That's a lie. This file was not automatically generated :)

I see it's become a trend to manually copy things from the resources into a new file, and I will again advise against it. Instead query the proper resources. It even has a `distance` value already, which it calls `range`, so no need to ask the community to contribute. The only difference is that the `range` parameter is not in game units, but has to be translated by a table. Can read some more about it in a [post about it](http://forums.windower.net/index.php?/topic/412-distance-and-gearswap-question/?p=2332).

z16 (changes requested):
I'm a bit torn on the addon. Its purpose is clear and good, its implementation can be cleaned up a little, but is otherwise ok, but mainly I don't know why this is its own addon, instead of incorporating this into the distance addon. I'm not a big fan of having two addons with essentially the same purpose, where they only differ in the amount of features they provide for that purpose.

That said, it's not gonna stop me from accepting it, but I'd like you to consider the option of integrating it into Distance as it currently is (or replacing Distance entirely, which would also be fine).

There are a few things in the code that could be done better, I mentioned them in the detailed sections below. One thing I didn't mention though: please sanitize your indentations. You mix spaces of varying length and tabs throughout the file. Please use a consistent 4 space indentation, to make it easier to review and work on by others in the future.

SammehFFXI:
Thanks for the code reviews.  This was one of my first; and have been running it for quite some time w/o issue and will be happy to clean it up.  I have several more that I'm going to submit but I can't even figure out git well enough to do that multi-threaded.   (Tried branches, but the multiple pull requests just went back into my main dev branch instead of Windower's).  (If you want to look at others - github.com/SammehFFXI/FFXIAddons  - I'll be hoping to submit PetCharges; JobChange; Sleeper; Snub; and potentially React (if you guys don't consider it too much automtaion).

Anyway - I appreciate the input since I'm not a developer but I do have some general good ideas. If you want to 'replace' Distance or 'integrate' we can easily do that - Again just didn't know etiquette.  Let me see what I can do to clean this up and merge changes!

addons/DistancePlus/DistancePlus.lua
SammehFFXI: I'm confused if you're referencing the 'self' or the displayabilities.   The displayabilities has to be called as part of the pre-render as it is what does the color coordination based on distance.  So it's very valid.  I have moved self to load/login events. Very open to ideas or clarification here! Thanks for the feedback.

addons/DistancePlus/DistancePlus.lua
SammehFFXI: Cleaning up thanks. Actually wasn't using the ability recast at all so that pointed it out. That was an idea in motion that never took flight until really another addon.

addons/DistancePlus/DistancePlus.lua
SammehFFXI: I hear ya.  Took the 't' right from the existing Distance plugin and then did very little other than replace some variables to get the pet bit.   Fixing up.

addons/DistancePlus/DistancePlus.lua
z16: You can extract the expensive compution and only leave certain parts in there. For example, you can calculate the ability list once on load, login and job change, then just replace the distance value in the text on the `prerender` event. The texts library even provides a mechanism for that, by declaring variables:
```lua
t = texts.new('Some text with two variables: ${varA}, ${varB}')

t.varA = 'this'
t.varB = 'that'

t:show() -- This will show a text box with this text: Some text with two variables: this, that'
```

So you could use this to make the color and the distance variable (or even the ability list) and only update any one of the components when it changes.

SammehFFXI:
I really appreciate many of the suggestions.  I don't think I was able to implement them all but most.   I look forward to continued feedback.

On the 'duplicating' resources bit - I was going to use built in resources but found much of it incomplete or wrong; but never realized I could contribute to that as well.  So a new version I'll likely make a move to using resources and submit what I believe to be correct values there.

Thanks again =)

z16:
You can't contribute to resources, they only read FFXI DAT data, so they're certainly not wrong (unless the DATs themselves are wrong, which has happened before). What are the values you think are wrong?

SammehFFXI:
I'll have to remember; I wrote this several months ago.  I vaguely recall that card abilities were showing wrong for corsair;   The 'range's felt very off - For example - 'Fight' (bst pet command) is 16 yalms + model offset,   the resources has it listed as 11.   Bounty shot in resources is range of 12 - yet I have it recorded as 20.     Shadowbind similarly in resources is 12 but is definitely over 20.

trv6:
Those numbers fit with the table Arcon linked to.
Fight: 11 * 1.490909 = 16.39999
Bounty shot: 12 * 1.7 = 20.4

z16:
The post I linked to showed the calculation, here it is again:

```lua
range_mult = {
      [2] =  1.70,
      [3] = 1.490909,
      [4] = 1.44,
      [5] = 1.377778,
      [6] = 1.30,
      [7] = 1.20,
      [8] = 1.30,
      [9] = 1.377778,
     [10] = 1.45,
     [11] = 1.490909,
     [12] = 1.70,
}

distance = target.model_size + spell.range * range_mult[spell.range]
```

Here `spell` is the spell/ability from the resource and `target` is the mob object of your target. The formula itself may need tweaking, but in terms of yalms it yields decent enough results.

SammehFFXI:
Ok I misread that - I'll see what i can get going.   Thanks for the patience.

SammehFFXI:
Ok - Deleted the ja_distance extra file; I did end up having to mod some of the ranges from the page you referenced but they were pretty close.  In the equation though I'm also adding in s.model_size (because that's different between taru vs hume/elv/mithra vs Galka.    

Thanks for the constant feedback =)

addons/DistancePlus/readme.md
```diff
@@ -0,0 +1,37 @@
+--[[
```
z16: The readme does not need a copyright notice, and it's unusual to have it (no other addon does). I'm not sure if that would even be valid, since the license is explicitly about code (`[…] in source and binary forms […]`).

addons/DistancePlus/DistancePlus.lua
```diff
@@ -0,0 +1,420 @@
...
+        windower.add_to_chat(8,'Mode: Default.')
+        option = "Default"
+        MaxDistance = 25
+    end
+end
+
+
+windower.register_event('prerender', function()
+    t = windower.ffxi.get_mob_by_target('t') or windower.ffxi.get_mob_by_target('st')
+    s = windower.ffxi.get_mob_by_target('me')
```
z16: These values are still missing a `local` keyword and as such bleed into the global table. Is that what you intended? Nothing else uses them, so that feels weird, and it's one potential error source.

addons/DistancePlus/DistancePlus.lua
```diff
@@ -0,0 +1,420 @@
...
+        [9] = 1.377778,
+        [10] = 1.45,
+        [11] = 1.454545454545455,
+        [12] = 1.666666666666667,
+    }
+
+    list = 'Abilities:\n'
+    if abilitylist then 
+      for key,ability in pairs(abilitylist) do
+        ability_en = res.job_abilities[ability].en
```
z16: Two things when using resources, first be aware that `en` is not meant to be used - use `english` instead.

Second, if you use `name` instead of `english` (or `en`) the name will display correctly on Japanese clients as well, so you'd be doing them a favor.

addons/DistancePlus/DistancePlus.lua
```diff
@@ -0,0 +1,420 @@
...
+        [6] = 1.30,
+        [7] = 1.15,
+        [8] = 1.25,
+        [9] = 1.377778,
+        [10] = 1.45,
+        [11] = 1.454545454545455,
+        [12] = 1.666666666666667,
+    }
+
+    list = 'Abilities:\n'
```
z16: Same as below, missing a `local` keyword.

z16 (changes requested):
Added one more review, though this one is somewhat minor. You do still have some tabs in your code. I'd search/replace them en masse with your text editor/IDE.

Aside from that the resource thing with `en` is the only thing I really want to see changed. The rest is up to you.

addons/DistancePlus/DistancePlus.lua
SammehFFXI: t and s are used in the 'displayabilities' function which are global.   I could pass them as parameters, but wasn't sure which was more expensive?

SammehFFXI:
Ok - I think all changes made unless you have some recommendation about the 't' and 's' in the displayabilities section.

addons/DistancePlus/DistancePlus.lua
alphaONE2: The difference is negligible. You would never be able to tell the difference in practice, but... passing as function arguments is faster since they'll end up in registers, avoiding a hash table look up.

PR #1541, opened 2017-05-11
New Addon: CapeTrader

OdinBurntwaffle:
Hello. The capetrader addon is designed to automate the process of augmenting ambuscade capes.

Unfortunately I think I goofed my original pull request for this addon. Hopefully this request worked out. Looking forward to any suggestions and input you all might have. Thanks.


addons/capetrader/capetrader.lua
```diff
@@ -0,0 +1,495 @@
...
+require('strings')
+local res = require('resources')
+local packets = require('packets')
+local augPaths = require('allAugPaths')
+local maxAugMap = require('maxAugMap')
+local extData = require('extdata')
+local jobToCapeMap = require('jobToCapeMap')
+
+local playerIndex = nil
+local gorpaID = 17797273
```
z16: Two things to keep in mind about using NPC IDs like this:
1. It will likely change. NPC IDs frequently do, so this would need to be maintained after every update. That may warrant putting it in a more visible place, where others could maintain it better as well, just in case you stop playing at some point.
2. I'd recommend writing it down as a hex number. That allows you to distinguish some parts more easily:
`0x010F9099`
This means it's a regular NPC (starts with `0x01`), is in Mhaura (zone `0x0F9`, or `249` in decimal) and has index `0x099` (`153`). That also means you can derive the following few lines from it and don't need to explicitly list them, as they may change as well.

addons/capetrader/capetrader.lua
```diff
@@ -0,0 +1,495 @@
...
+local threadIndex = 1
+local dustIndex = 2
+local dyeIndex = 3
+local sapIndex = 4
+local maxAugKey = nil
+local zoneHasLoaded = true
+local inventory = nil
+
+windower.register_event('addon command', function(input, ...)
+    local cmd = string.lower(input)
```
z16: I would recommend not relying on the implicit `arg` value Lua 5.1 provides. It was even removed in Lua 5.2 because it can confuse people and pollute the local variable space, and has hardly any advantage. It's easily implemented in one line:
```lua
    local args = {...}
```

This will also make it easier to convert the addon to newer versions of Windower as well, which don't support that.

addons/capetrader/capetrader.lua
```diff
@@ -0,0 +1,495 @@
...
+    end
+
+    if id == 0x01D then
+        if not zoneHasLoaded then
+            zoneHasLoaded = true
+        end
+        if tradeReady then
+            tradeReady = false
+
+            if not string.contains(string.lower(pathItem.en), 'dye') then
```
z16: This may just be your preference, but this is more succintly typed when using the object syntax:
```lua
            if not pathItem.en:lower():contains('dye') then
```

This can be applied to many cases throughout this file, and with functions too, such as in the next line:
```lua
                startAugmentingCape:schedule(dustSapThreadTradeDelay, numberOfTimesToAugment - timesAugmentedCount + 1, false)
```

addons/capetrader/capetrader.lua
```diff
@@ -0,0 +1,495 @@
...
+        firstTimeAug = false
+        optionIndex = 512
+    else
+        optionIndex = 256
+    end
+
+    local packet = packets.new('outgoing', 0x05B)
+    packet["Target"] = gorpaID
+    packet["Option Index"] = optionIndex
+    packet["_unknown1"] = pathIndex
```
z16: Did you actually figure out what `_unknown1` is, or have you just observed its value in this particular case?

If it's the former, I'd welcome it if you could update our `Windower/addons/libs/packets/fields.lua` file with the new info.

addons/capetrader/capetrader.lua
```diff
@@ -0,0 +1,495 @@
...
+    packet["Target Index"] = gorpaTargetIndex
+    packet["Automated Message"] = true
+    packet["_unknown2"] = 0
```
z16: If you know what this is, see the comment above. If not and you're just setting it to `0` to be safe, you don't need to bother. All fields you don't explicitly specify are initialized to `0` by default.

addons/capetrader/capetrader.lua
```diff
@@ -0,0 +1,495 @@
...
+    packet["Zone"] = mhauraID
+    packet["Menu ID"] = gorpaMenuID
+    packets.inject(packet)
+
+    packet["Automated Message"] = false
+    packets.inject(packet)
+
+    packet = packets.new('outgoing', 0x016, {
```
z16: I'd recommend choosing a new variable for a new packet. In the above case I'd say it's ok, because the packet is essentially the same, just slightly modified, but here it's entirely new. Having a new variable for it reduces the risk of mixing things up when refactoring the code. Which could be especially problematic for injected outgoing packets.

addons/capetrader/capetrader.lua
```diff
@@ -0,0 +1,495 @@
...
+    end
+
+    local augmentTable
+    if extData.decode(capeItem).augments then
+        augmentTable = extData.decode(capeItem).augments
+    end
+
+    local augValue
+    if augmentTable then
+        if string.contains(string.lower(pathItem.en), 'thread') then
```
z16: I'm not sure if it's applicable here, but if the name should not just *contain* `thread` but actually *end* with it, you can use `string.endswith`. It has the advantage of avoiding false positives and is also more efficient.

And in the interest of avoiding false positives you can also add a space to the word you're checking for:
```lua
        if pathItem.en:lower():endswith(' thread') then
```

addons/capetrader/capetrader.lua
```diff
@@ -0,0 +1,495 @@
...
+local dustSapThreadTradeDelay = 1
+local dyeTradeDelay = 2
+local tradeReady = false
+local threadIndex = 1
+local dustIndex = 2
+local dyeIndex = 3
+local sapIndex = 4
+local maxAugKey = nil
+local zoneHasLoaded = true
+local inventory = nil
```
z16: You rely on a lot of global state. This is a simple way to implement certain mechanisms, but is also prone to some issues. For example, you always have to remember to clean up the state wherever you leave certain functions. This makes those functions harder to design, since you can really only have one exit route.

Instead I'd consider passing more things as arguments. One simple example is the `getCapeIndex` function. It uses the `currentCape` variable, which is defined globally. Instead, you can simply pass the value to the function where it's called inside `checkCapeCount`. Which, in turn, can also receive it from its calling function. This both avoids the need to clean up, since it doesn't persist by default, as well as allows it to trace easily which functions need which values, since you see it in their argument list. That helps others (like me :D) to understand the code better and will help you if you decide to refactor it in the future.

The same can be applied to many other variables.

On the other hand, certain variables make sense to handle globally, namely those that are either inefficient to calculate (like player ID/index or zone info). Those can be stored globally and certain events can be registered which shall update those variables (like login/logout/zoning).

z16 (changes requested):
I added some comments, and most of it looks pretty good. Even the comments I added are merely suggestions.

There's just one thing that I didn't mention in the code that I'd like changed before I merge this, and that's the use of `item.en` throughout the code. It's not your fault, I don't think we have it documented anywhere, but that's actually an internal value. Instead please use `english` (or even `name`, which will cause it to be Japanese on a Japanese client). The abbreviated versions are not guaranteed to work in future versions of Windower.

So as long as that is changed, the rest is fine to merge. Just let me know if you want some time to work on some of my suggestions, or if you want it to be merged as is.

addons/capetrader/capetrader.lua
trv6: I bumped into this problem not too long ago and learned a few things which I thought were interesting. 
The `arg` variable is a leftover from Lua 5.0, which sneaks into 5.1 via a compatibility option that is on by default. It's off in LuaJIT.
There's at least one very good reason not to use it: because creating the `arg` table adds overhead to functions with the `...` syntax, Lua 5.1 only defines it if you don't directly use the syntax elsewhere in your function (`select(n, ...)`, etc.). This can trip you up if you're not aware -- suddenly you'll reach for `arg` and it won't be there -- but it has an upside! Defining `local args = {...}` probably doesn't cost much because Lua would have spent the time defining `arg`, but won't when it sees you using the ellipsis.

addons/capetrader/capetrader.lua
OdinBurntwaffle: I added some code in order to quickly check these values after a version update. If you could  have a look at the printGorpaPacketInfo() function and the windower.register_event('outgoing chunk') portion of code and maybe tell me if there is a better way to get that information that would be great. I also added some comments to the capetrader.lua file to explain how to use those functions to get the relevant information in case I am not around to patch the addon.

addons/capetrader/capetrader.lua
OdinBurntwaffle: Keeping track of the variables here  was definitely a cause of many difficulties for me while writing this addon! Unfortunately I only managed to get rid of three of these relating to item index values as you suggested. I think I could manage to get rid of some more of them if I combined the prepareCapeForAugments() and startAugmentingCape() functions into one large function but I'm not sure if that would be worth the effort. If you see any other simpler solutions for removing some of these variables let me know. Thanks for the suggestion.

addons/capetrader/capetrader.lua
OdinBurntwaffle: Fixed!

addons/capetrader/capetrader.lua
OdinBurntwaffle: Fixed!

addons/capetrader/capetrader.lua
OdinBurntwaffle: Fixed!

addons/capetrader/capetrader.lua
OdinBurntwaffle: I've made a couple of small addons similar to this one that use this 0x05B packet as well. It seems like "_unknown1" will take values other than zero if the 0x05B dialog packet you are sending results in you exiting the related menu in game. Although it seems there are a few exceptions to this in the capetrader addon, for example unknown1 will be zero if you augment a cape and exit the menu with any of the following pairs of items/augmentpaths: thread/hp, dust/acc+atk, sap/wsd, dye/hp. 

I didn't add a note to the packets/fields.lua file because I'm not entirely sure how unknown1 works. I think if there are multiple dialog options that can result in exiting a menu unknown1 won't necessarily be equal to zero and can take different values. If you think that is plausible and worth adding to the fields file let me know and I can add an explanatory comment of some kind.

addons/capetrader/capetrader.lua
OdinBurntwaffle: Removed the line, I'm not sure what "_unknown2" is for.

addons/capetrader/capetrader.lua
OdinBurntwaffle: Fixed!

OdinBurntwaffle:
I think the latest push addressed most of your suggestions. I wasn't sure whether to replace item.en with item.english or item.name, I ended up using .name. Let me know if I should use .english instead. Let me know if you see any problems I missed or may have introduced while changing things.  I tested the addon again as much as I could and everything seems to be working properly after these changes. As far as I can tell it is ready to be merged, let me know otherwise.

Thanks again for the feedback!

addons/capetrader/capetrader.lua
```diff
@@ -0,0 +1,516 @@
...
+
+_addon.name = 'CapeTrader'
+_addon.author = 'Lygre, Burntwaffle'
+_addon.version = '1.0.2'
+_addon.commands = {'capetrader', 'ct'}
+
+--NOTE: It is possible that the correct values for the following four variables can change after a version update.
+local gorpaID = 0x010F9099
+local mhauraID = 0x0F9
+local gorpaTargetIndex = 0x099
```
z16: While this helps outline the values, what I meant is to only define one of them. Then only one needs to be updated. For example, you can define only Gorpa's ID and derive the other two from it:
```lua
local gorpaID = 0x010F9099
local zoneID = math.floor(gorpaID / 0x100) % 0x1000
local gorpaIndex = gorpaID % 0x1000
```

Or you can define only the index and get Mhaura's ID from the resources:
```lua
local gorpaIndex = 0x099
local mhauraID = res.zones:with('english', 'Mhaura').id
local gorpaID = 0x01000000 + mhauraID * 0x1000 + gorpaIndex
```

Actually looking at it now it's not exactly prettier :D

The point being only one value should be variable when the others can be derived from it. This simply reduces the maintenance work and the surface area for manual bugs. But in this case the gains are admittedly rather small, so I'll leave that up to your judgment.

addons/capetrader/capetrader.lua
```diff
@@ -0,0 +1,516 @@
...
+
+function getItemIndex(capeOrAugItem)
+    for itemIndex, item in pairs(inventory) do
+        if item.id == capeOrAugItem.id then
+            return itemIndex
+        end
+    end
+end
+
+function getItem(itemName)
```
z16, line 153: This function searches through the item resources. All resource tables already provide a mechanism to do that:
```lua
    return res.items:with('name', itemName)
```

This will not match case insensitively, so the case would need to be correct. Otherwise you can do the more verbose:
```lua
    return res.items:with('name', function(name) return name:lower() == itemName end)
```

Both functions (as well as your original loop) are annoyingly inefficient. The item resources are huge with several thousand entries, and every time the function is called it searches through them all and even lowercases their names. You can avoid much of that effort by not lowercasing names unless you have to (such as when handling user input), but sometimes there's no way around it.

In this case however, you can do one of two things:
1. Memoization. That's a term describing saving function input and a hash table of results, so that only the first time the resources are searched. The second time the hashed value is taken. One possible implementation would be this:
```lua
do
    local results = {}
    getItem = function(itemName)
        itemName = itemName:lower()
        if not results[itemName] then
            results[itemName] = res.items:with('name', function(name) return name:lower() == itemName end)
        end

        return results[itemName]
    end
end
```
2. Precomputing a table of known values. This method is especially applicable if there are only a certain known number of valid values, which I think is the case here (although I'm not familiar with the augmenting mechanism, so I could be wrong). In this case you'd iterate through the resources on addon load and create a name-item mapping table. Then the `getItem` function wouldn't need to search at all, nor manage its own hash table, it would just look up the value in the global table you computed before.

z16 (changes requested):
I added some more comments, but mostly I just wanted to mention the `name` vs. `english` thing again. Since you asked when to use which, simply think about how you use them.

Take this example:
```lua
    print(res.items[19747].name == 'Mandau')
```

Since the value of `name` depends on the client, this would compare `'Mandau' == 'Mandau'` on an English client, which is true. But on a Japanese client it would compare `'マンダウ' == 'Mandau'`, which is false.

Basically, use `name` when you're printing or reading, i.e. talking to the user. Use `english` when you compare against hard-coded values.

So in your case, you check if `item.name:endswith(' dust')` at some point. This will now break for Japanese users, since the name will not end with ' dust' for them. Instead use `english` here.

But on line 260, where you print a message to the console, using `name` would be fine. It would still be English text, and no one expects you to translate it, but at least a Japanese player will figure out that the error is regarding the correct item. Item names are especially hard to figure out in another language, so even a Japanese item name in an English sentence would help users.

As for the debugging function you added, that's fine. Unfortunately there' no sure-fire way to handle such cases. Nothing we can do after an update but figure this stuff out again.

Fortunately you don't even have to wait until the update finishes to update it. [POLUtils] is a tool that can read FFXI DAT files and it can find out NPC IDs/indices. So if you want to provide maximum security for your users after an update, before the servers are up you should try to update as soon as possible and then get the new value from POLUtils. Then submit a pull request with the new value. A bit of a hassle, but that's what it is :)

z16:
Actually, I just thought of something else you could do that would probably make updating at all unnecessary. You could simply look for the NPC name... that usually doesn't change. So hard-code the name and just check if a mob with that name is in the vicinity. We have a function to get a mob by name:
```lua
    local gorpa = windower.ffxi.get_mob_by_name('Gorpa something')
```

Since Mhaura's ID will likely always be the same (and you can still get it from the resources to be super sure), you can get the ID and index from that object. And then you already have the object just need to check if it's not `nil` and if it's in range.

The more I think about it, the more I think this is the way it should be. I'll ponder it a bit more, and I won't require you to change it to that, just something to think about, since it could save yourself some effort in the long run.

OdinBurntwaffle:
Hello again. I think I replaced item.name with item.english where appropriate.

I made a first attempt at changing how we get the gorpa npc information. Is 
`windower.ffxi.get_mob_by_name('Gorpa-Masorpa')` a particularly expensive function? I think I could probably get rid of a few more global variables related to the gorpa npc if we just replace instances of `gorpaID `and `gorpaTargetIndex` with `windower.ffxi.get_mob_by_name('Gorpa-Masorpa').id` and `windower.ffxi.get_mob_by_name('Gorpa-Masorpa').index` respectively assuming we don't mind calling that function too much.

I also made a first attempt at fixing searching `res.items ` problem. I tried to follow your suggestion of precomputing a table of known values so we only need to search `res.items` once.

Thanks again for the feedback.

z16:
As you may suspect, the function is usually somewhat expensive. But the way I see it you only need to call it when the user enters a command. For single-use it's really quite ok to use. Also, you can store the result, since once you have the ID/index it won't change for the entire of the FFXI version until the next update. So you can use that to determine the ID and then just call `get_mob_by_id`. Although you won't even need that... since his position also never changes. So if you have it once, you could just store its position, and always compare against that.

But that's all just micro-optimizing at this point. Like I said, `get_mob_by_name` may be "expensive", but only in relative terms to the other `get_mob_by_*` functions. It's still fast enough to carefree use it every time the user enters a command. Its raw execution time is very like still under the millisecond threshold for most machines.

Let me know if there's still something you want to implement, or if you're ready on your end.

OdinBurntwaffle:
Made a few more small changes. I tested some things after these changes again and it seems the addon is working correctly. (At least I hope so!)

Unless you find something else objectionable I think it is ready to merge. 

Thank you for the help.

z16 (approved):

PR #1543, opened 2017-05-15
New Addon - JobChange

SammehFFXI:
Addon Job Change submission.  Allows command line job changing when
close to a moogle npc. Great for macro'ing when doing ambuscade to
quickly reset JA's and/or main/sub's to run back and forth to gather
KI's.

### DIFF
--- a/addons/JobChange/jobchange.lua
+++ b/addons/JobChange/jobchange.lua
@@ -28,11 +28,12 @@ SOFTWARE, EVEN IF ADVISED OF THE POSSIBILITY OF SUCH DAMAGE.
 
 _addon.name = 'Job Change'
 _addon.author = 'Sammeh'
-_addon.version = '1.0.2'
+_addon.version = '1.0.3'
 _addon.command = 'jc'
 
 -- 1.0.1 first release
 -- 1.0.2 added 'reset' command to simply reset to existing job.  Changes sub job to a random starting job and back.
+-- 1.0.3 Code clean-up
 
 require('tables')
 packets = require('packets')
@@ -41,35 +42,21 @@ res = require ('resources')
 function jobchange(job,main_sub)
     windower.add_to_chat(4,"JobChange: Changing "..main_sub.." job to:"..res.jobs[job].ens)
     if job and main_sub then 
-        if main_sub == 'main' then 
-            local packet = packets.new('outgoing', 0x100, {
-            ["Main Job"]=job,
-            ["Sub Job"]=0,
-            ["_unknown1"]=0,
-            ["_unknown2"]=0
-            })
-            packets.inject(packet)
-            coroutine.sleep(0.5)
-        elseif main_sub == 'sub' then
-            local packet = packets.new('outgoing', 0x100, {
-            ["Main Job"]=0,
-            ["Sub Job"]=job,
-            ["_unknown1"]=0,
-            ["_unknown2"]=0
-            })
-            packets.inject(packet)
-            coroutine.sleep(0.5)
-        end
-    end
+        local packet = packets.new('outgoing', 0x100, {
+            [main_sub:sub(1,1):upper()..main_sub:sub(2)..' Job'] = job,
+        })
+        packets.inject(packet)
+        coroutine.sleep(0.5)
+    end    
 end
 
 windower.register_event('addon command', function(command, ...)
+    local self = windower.ffxi.get_player()
     local args = L{...}
     local job = ''
     if args[1] then 
         job = args[1]:lower()
     end
-    local currentjob = windower.ffxi.get_player()
     local main_sub = ''
     if command:lower() == 'main' then
         main_sub = 'main'
@@ -84,15 +71,15 @@ windower.register_event('addon command', function(command, ...)
         windower.add_to_chat(4,"JobChange Syntax: //jc reset -- Resets Current Job")
         return
     end
-    local conflict = find_conflict(job)
-    local jobid = find_job(job)
+    local conflict = find_conflict(job,self)
+    local jobid = find_job(job,self)
     if jobid then 
         local npc = find_job_change_npc()
         if npc then
             if not conflict then 
                 jobchange(jobid,main_sub)
             else
-                local temp_job = find_temp_job()            
+                local temp_job = find_temp_job(self)            
                 windower.add_to_chat(4,"JobChange: Conflict with "..conflict)
                 if main_sub == conflict then 
                     jobchange(temp_job,main_sub)
@@ -111,8 +98,7 @@ windower.register_event('addon command', function(command, ...)
     
 end)
 
-function find_conflict(job)
-    local self = windower.ffxi.get_player()
+function find_conflict(job,self)
     if self.main_job == job:upper() then
         return "main"
     end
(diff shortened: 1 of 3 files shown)


addons/JobChange/jobchange.lua
```diff
@@ -0,0 +1,218 @@
...
+        -- Zones with a nomad moogle / green thumb moogle, taken from Resources
+        -- All other zones check if mog_house
+        [26] = {id=26,en="Tavnazian Safehold",ja="タブナジア地下壕",search="TavSafehld"},
+        [53] = {id=53,en="Nashmau",ja="ナシュモ",search="Nashmau"},
+        [247] = {id=247,en="Rabao",ja="ラバオ",search="Rabao"},
+        [248] = {id=248,en="Selbina",ja="セルビナ",search="Selbina"},
+        [249] = {id=249,en="Mhaura",ja="マウラ",search="Mhaura"},
+        [250] = {id=250,en="Kazham",ja="カザム",search="Kazham"},
+        [252] = {id=252,en="Norg",ja="ノーグ",search="Norg"},    
+    }
```
z16: Copied from the resources again :) Here's what you could do instead:

```lua
    if not (info.mog_house or S{'Selbina', 'Mhaura', --[[and so on]]}:contains(res.zones[info.zone].english)) then
```
```

File: addons/JobChange/jobchange.lua [hunk 2]
```diff
@@ -0,0 +1,218 @@
...
+    if not (valid_zones[info['zone']] or info['mog_house']) then
+        windower.add_to_chat(4,'JobChange: Not in a zone with a Change NPC')
+        return
+    end
+    for i,v in pairs(windower.ffxi.get_mob_array()) do
+        if v['name'] == 'Moogle' or v['name'] == 'Nomad Moogle' or v['name'] == 'Green Thumb Moogle' then
+            found = 1
+            target_index = i
+            target_id = v['id']
+            distance = windower.ffxi.get_mob_by_id(target_id).distance
```
z16: This can just be:
```lua
            distance = v.distance
```

But why have all these global variables, when you can just do:
```lua
    for i, v in pairs(windower.ffxi.get_mob_array()) do
        if v.distance < 36 and S{'Moogle', 'Nomad Moogle', 'Green Thumb Moogle'}:contains(v.name) then
            return v
        end
    end
```

That way this function returns the mob object itself. Then you don't need to mess around with a separate variable for checking whether or not you found it. In the calling function you can just do `if mob then` now. I also adjusted `math.sqrt(distance) < 6` to `distance < 36`, that way you save a `sqrt` operation. Not at all necessary, just saying, one more way to consider performance.

addons/JobChange/jobchange.lua
```diff
@@ -0,0 +1,218 @@
...
+    local self = windower.ffxi.get_player()
+    local ki = windower.ffxi.get_key_items() 
+    
+    for index,value in pairs(res.jobs) do
+        if value.ens:lower() == job then 
+            jobid = index
+        end
+    end
+    --[[
+        Small rant/request/can't figure out a better way.
```
z16: There was no better way. But there is now! I just adjusted the new hook to gather that info and *LuaCore* now reports level `0` for jobs that are not unlocked yet.

addons/JobChange/jobchange.lua
```diff
@@ -0,0 +1,218 @@
...
+        windower.add_to_chat(4,'JobChange: Not in a zone with a Change NPC')
+        return
+    end
+    for i,v in pairs(windower.ffxi.get_mob_array()) do
+        if v['name'] == 'Moogle' or v['name'] == 'Nomad Moogle' or v['name'] == 'Green Thumb Moogle' then
+            found = 1
+            target_index = i
+            target_id = v['id']
+            distance = windower.ffxi.get_mob_by_id(target_id).distance
+            if math.sqrt(distance)<6 then
```
z16: As I wrote before, it's a common trick to save yourself one square root operation to just square the other side of the comparison:
```lua
        if distance < 36 then
```

addons/JobChange/jobchange.lua
```diff
@@ -0,0 +1,218 @@
...
+    end
+end
+
+windower.register_event('addon command', function(command, ...)
+    local args = L{...}
+    local job = ''
+    if args[1] then 
+        job = args[1]:lower()
+    end
+    local currentjob = windower.ffxi.get_player()
```
z16: You call the expensive `windower.ffxi.get_player` function, yet you never use the result. However, below you call the same expensive function several times again, instead of using this value. I would pass this value to `find_conflict` and `find_temp_job` to avoid expensive recomputations of the same value.

addons/JobChange/jobchange.lua
```diff
@@ -0,0 +1,218 @@
...
+
+function jobchange(job,main_sub)
+    windower.add_to_chat(4,"JobChange: Changing "..main_sub.." job to:"..res.jobs[job].ens)
+    if job and main_sub then 
+        if main_sub == 'main' then 
+            local packet = packets.new('outgoing', 0x100, {
+            ["Main Job"]=job,
+            ["Sub Job"]=0,
+            ["_unknown1"]=0,
+            ["_unknown2"]=0
```
z16: The default values for packet fields are `0`, so you don't need to explicitly specify them. Only specify what you actually need to specify. The same goes for the packet definition below.

addons/JobChange/jobchange.lua
```diff
@@ -0,0 +1,218 @@
...
+        elseif main_sub == 'sub' then
+            local packet = packets.new('outgoing', 0x100, {
+            ["Main Job"]=0,
+            ["Sub Job"]=job,
+            ["_unknown1"]=0,
+            ["_unknown2"]=0
+            })
+            packets.inject(packet)
+            coroutine.sleep(0.5)
+        end
```
z16: The `if` and `else` branch do the same thing, only differing on one value. You could easily rewrite lines 44 through 62 to this:
```lua
    packets.inject(packets.new('outgoing', 0x100, { [main_sub:capitalize() .. ' Job'] = job }))
    coroutine.sleep(0.5)
```

SammehFFXI:
Thanks for the good recommendations.  I think I've incorporated them all =)   I couldn't find a capitalize function (not sure if I needed an extra include) so I slightly changed it with substrings.  Otherwise good to go on rest =).

z16:
Right sorry, you would need to `require('strings')` for the string helper functions to become available. But that's ok, this way works fine.

addons/JobChange/jobchange.lua
```diff
@@ -0,0 +1,145 @@
...
+    end
+    local main_sub = ''
+    if command:lower() == 'main' then
+        main_sub = 'main'
+    elseif command:lower() == 'sub' then
+        main_sub = 'sub'
+    elseif command:lower() == 'reset' then
+        windower.add_to_chat(4,"JobChange: Resetting Job")
+        main_sub = 'sub'
+        job = windower.ffxi.get_player().sub_job:lower()
```
z16, line 68: You use `get_player` again here, even though you have access to the `self` variable, which is the same value (as you just set it before).

addons/JobChange/jobchange.lua
```diff
@@ -0,0 +1,145 @@
...
+        coroutine.sleep(0.5)
+    end    
+end
+
+windower.register_event('addon command', function(command, ...)
+    local self = windower.ffxi.get_player()
+    local args = L{...}
+    local job = ''
+    if args[1] then 
+        job = args[1]:lower()
```
z16, line 58: Here and below you call `:lower()` on the job, only to call `:upper()` on it later for the string comparison. This is very nitpicky at this point but I'd recommend just doing one, either lowercase or uppercase, and then sticking with it for the remainder of the addon. Since both *LuaCore* and *FFXI* itself use only uppercase to refer to job abbreviations, I'd use the same as well. Then you only need one `:upper()` call right here, and you can remove the `:lower()` call on line 68 and the two `:upper()` calls inside the `find_conflict` function.

Like I said, rather nitpicky at this point :) Just a recommendation to reduce unnecessary code where possible.

addons/JobChange/jobchange.lua
```diff
@@ -0,0 +1,145 @@
...
+    end
+    local main_sub = ''
+    if command:lower() == 'main' then
```
z16, line 61: Speaking of nitpicky... a common idiom for the command handler is this first line:
```lua
```

That way you don't need to `:lower()` every time you compare it, and it defaults to 'help', which you may or may not check for.

addons/JobChange/jobchange.lua
```diff
@@ -0,0 +1,145 @@
...
+        return "main"
+    end
+    if self.sub_job == job:upper() then
+        return "sub"
+    end
+end
+
+function find_temp_job(self)
+    local starting_jobs = S { 'WAR', 'MNK', 'WHM', 'BLM', 'RDM', 'THF' } 
+    for i in pairs(starting_jobs) do
```
z16, line 112: About iteration! You may have noticed that you need `for k,v in pairs(t) do` for normal tables, yet `for v in pairs(s) do` for sets. This confusion has led us to create a helper function which behaves universally (and is a bit shorter):
```lua
    for job in starting_jobs:it() do
```

This works the same on T, S and L by returning the value first and the index second, so you can use that if you want. But again, just a minor suggestion. You can also use that below:
```lua
            for value, index in res.jobs:it() do
```

addons/JobChange/jobchange.lua
```diff
@@ -0,0 +1,145 @@
...
+        if not find_conflict(i,self) then 
+            for index,value in pairs(res.jobs) do
+                if value.ens == i then
```
z16, line 115: Please don't use `ens`, use `english_short` instead. `ens` is just an internal name, which just happens to work. I already wanted to drop support for it, but people were already using it, so it was too late, but we don't need to perpetuate this :)

addons/JobChange/jobchange.lua
```diff
@@ -0,0 +1,145 @@
...
+                    return index
+                end
+            end
```
z16, line 118: This loop can be replaced by a helper function from the resources library. You need to find a job with a certain short name. The resource library provides a mechanism to search for a specific resource with:
```lua
    for job in starting_jobs:it() do -- same as above
        if not find_conflict(job, self) then -- same as above
            return res.jobs:with('english_short', job).id -- inner loop replaced with this
        end
    end
```

Now, you could also exploit that the starting jobs have the IDs 1 through 6, and use another helper mechanism of the resources library:
```lua
funtion find_temp_job(self)
    -- This creates a list of starting jobs, including all the fields from the resources (id, name, english, english_short, etc.)
    local starting_jobs = res.jobs:id(function(id) return id >= 1 and id <= 6 end)
    -- This applies a filter function to the list of starting jobs. The result is a list of all starting jobs that don't have a conflict)
    local valid_jobs = starting_jobs:filter(function(job) return not find_conflict(job, self) end)
    -- Then you just select the first one of those and return its ID (which is safer to go for than the index)
    return valid_jobs[1].id
end
```

addons/JobChange/jobchange.lua
```diff
@@ -0,0 +1,145 @@
...
+        end
+    end
+end
+
+function find_job(job,self)
```
z16, line 123: Again, this function can be shortened quite a bit. First, note what the `jobLevel` variable does. If it's 0, then the loop goes through all jobs, but the condition will always be false. I.e. it's not reliant on the loop variable. In that case you can remove it from the loop entirely:
```lua
function find_job(job, self)
    if self.jobs[job:upper()] == 0 then
        return
    end

    -- loop here
end
```

And like I said before, you could save yourself the `:upper()` if you did it on `args[1]` earlier, since the value is already uppercase from any other source.

addons/JobChange/jobchange.lua
```diff
@@ -0,0 +1,145 @@
...
+    local jobLevel = self.jobs[job:upper()]
+    for index,value in pairs(res.jobs) do
+        if value.ens:lower() == job and jobLevel > 0 then 
+            return index
+        end
+    end
```
z16, line 129: The entire loop itself can be replaced with the same function call as I suggested previously, using the resource library's helper mechanism:
```lua
    return res.jobs:with('english_short', job:upper()).id
```

Note that I switched from lowercasing the resource value to uppercasing the function's argument job value, which will amount to the same. And same as before, you can remove that entirely if you uppercase `args[1]` in the command handler.

addons/JobChange/jobchange.lua
```diff
@@ -0,0 +1,145 @@
...
+
+function find_job_change_npc()
+    local info = windower.ffxi.get_info()
+    if not (info.mog_house or S{'Selbina', 'Mhaura', 'Tavnazian Safehold', 'Nashmau', 'Rabao', 'Kazham', 'Norg'}:contains(res.zones[info.zone].english)) then
+        windower.add_to_chat(4,'JobChange: Not in a zone with a Change NPC')
+        return
+    end
+
+    for i, v in pairs(windower.ffxi.get_mob_array()) do
+        if v.distance < 36 and S{'Moogle', 'Nomad Moogle', 'Green Thumb Moogle'}:contains(v.name) then
```
z16, line 141: A minor note on using `S{}` here. I do not know how much Lua tries to optimize code, I know it does not do a lot. So it wouldn't surprise me, if it couldn't optimize the set creation out of this. Here is what your code *says* to do:

1. Iterate over every mob in the zone
2. Compare the distance
3. Create a new table, containing three values, `'Moogle'`, `'Nomad Moogle'` and `'Green Thumb Moogle'`
4. Turn that table into a set using `S`
5. Check if the mob's name is in that set

You can see number 3 and 4 are repeated for every mob, which can be up to 2k mobs. But the result of those steps (a set containing the names of moogles) is not reliant on the loop variable, so in a similar move to the `jobLevel` thing I mentioned above, we can move it outside of the loop.

While that will realistically not to impact anyone much, the idea behind this would be to save the work where possible. So you can create the set outside of the loop, then it would only be created once:
```lua
    local moogles = S{'Moogle', 'Nomad Moogle', 'Green Thumb Moogle'}
    for i, v in pairs(windower.ffxi.get_mob_array()) do
        if v.distance < 36 and moogles:contains(v.name) then
            return v
        end
    end
```

z16 (approved):
I added a bunch more comments, but none of them will prevent me from merging this. I decided to go a bit more in-depth with them mainly because you seem to be an active developer and our API is shoddily documented, so you could use some of the stuff I mentioned here better in other addons you submit (especially utilizing the resources library more).

But like I said, all of the suggestions I wrote down are very optional and minor fine-tuning. If you want me to merge the code as it is, just say so, it's certainly good enough. You can also implement the changes another time and submit a new pull request, if you feel like it. All up to you now.

SammehFFXI:
I really appreciate the commentary.   I haven't taken a programming class since high school (18-20 years ago) and am not a developer.  Takes a lot for me to even understand some of your suggestions but I get every one.  Most of what I've done is based on combining various aspects from other people's code and changing - and most folks are as inefficient as I am ;)

Hell if I knew git better you'd have 5-6 addons to review right now but I can't even figure out branching/commits more than one at a time - so I'm just pushing through them.

Thanks for the good insights, Merge as-is if you don't mind and I'll try to incorporate for next version =)

-Sam

alphaONE2:
@SammehFFXI: GitHub has a great guide explaining the basic [git workflow](https://guides.github.com/introduction/flow/).

The specifics about how to branch, etc., depend on which git client you're using. If you're using the GitHub Desktop client, GitHub has guides for [branching](https://help.github.com/desktop/guides/contributing/creating-a-branch-for-your-work/), [committing](https://help.github.com/desktop/guides/contributing/committing-and-reviewing-changes-to-your-project/) and [submitting pull requests](https://help.github.com/desktop/guides/contributing/sending-a-pull-request/).

If you're just using git from the command line, then just do `git checkout -b <new_branch>` before making and committing your changes as usual. If you want to switch to another branch that already exists (say you're working on two different things at once) then use `git checkout <branch>`.

If you're using another client, then you'll need to consult the documentation to learn how to perform each step in your particular client.

PR #1544, opened 2017-05-18
Library refactoring

trv6:
This would prevent careless use of global variables (eg. reassigning `set`) from altering each library's behavior.


z16:
It's a good idea, though the entire system really needs an overhaul. If you want you could do this for most libraries. Would also help if the libraries would return their tables, instead of making things global, but unfortunately that would break shit for many people. But if you reeeeeeally feel like it you could refactor them to do both :D That way they would really be protected and isolated (for example instead of setting `_libs.sets = true` you could do `_libs.sets = set` and store the table there). The entire system is revamped for v5, so personally I wouldn't put too much effort into it if I were you, but like I said, if you're super bored you could give it a try.

Anyway, I'd merge this but there are tabs instead of spaces used in this. Please adjust that first :)

trv6:
Whitespace is taken care of.

I think I could switch the libraries around pretty quickly, but it wouldn't be very pretty. Since the really common libraries don't return their tables, there would be a long chain of something like this at the beginning of each file:
 ```
if not _libs.math then
    require 'maths'
end

local math = _libs.math
```
Since at least a few libraries require math, string, sets, lists, tables... etc., it could get kind of ugly.

trv6:
Err, actually, I think `require` handles the ugly bit, anyway. Someone could go through and set parts of `_libs` to nil, but then they'd kind of deserve the errors.

trv6:
Please don't merge: I haven't tested the changes.

z16:
Please don't use Lua's dumbass syntax :( `require('tables')`, anything else should be a criminal offense…

alphaONE2:
I agree on the syntax. I would be much more willing to accept the "no parentheses" syntax if you could use it consistently. If you're going to restrict it to functions taking a single string literal or table literal argument, then why even have the syntax in the first place? It just makes your code look inconsistent.

I can see some value in the table literal exception if you really like "named arguments", but personally I find the `func{name = value}` syntax confusing. It looks like an initializer from C++ with strange syntax.

z16:
Let me know when you tested it enough to merge it.

trv6:
I'm not sure I can offer a proper defense for the syntax. Consistent use of parentheses makes it very clear that a function is being called, but I find no parentheses a little easier to read (a little more language-like?), and I like to use it in _some_ cases where there is no loss of meaning.
I especially like to use it for `require`. Calls to require are usually all grouped together at the top of a file, so there's no other code around to cause confusion. Also, require's use is well understood, so it's less important to see `require('tables')` as a call to require than it is to simply see `'tables'` and know that you have access to that library. Similarly, `T{} S{} L{}`, etc.

I've added them back in, though, and corrected a few instances of `(c)`.

trv6:
Should be safe to merge.

PR #1551, opened 2017-06-12
Added craft addon.

Snapster-amb:
I added a lot of changes today so I will need to testing tomorrow but I figured I would throw this up for review as I'll probably need to adjust for feedback anyways.

### DIFF
--- a/addons/craft/craft.lua
+++ b/addons/craft/craft.lua
@@ -1,5 +1,5 @@
 --[[
-Craft v1.03312017
+Craft v1.1.0
 
 Copyright © 2017 Mojo
 All rights reserved.
@@ -30,7 +30,7 @@ SOFTWARE, EVEN IF ADVISED OF THE POSSIBILITY OF SUCH DAMAGE.
 
 _addon.name     = 'craft'
 _addon.author   = 'Mojo'
-_addon.version  = '1.03312017'
+_addon.version  = '1.1.0'
 _addon.commands = {'craft'}
 
 require('chat')
@@ -69,15 +69,21 @@ local supported = false
 local appropriated = {}
 local inventory = {}
 
-local bags = {
-    safe = 1,
-    storage = 2,
-    locker = 4,
-    satchel = 5,
-    sack = 6,
-    case = 7,
-    safe2 = 9
-}
+local function filter_bag(v)
+    return not (v.name:match("Inventory") or
+        v.name:match("Temporary") or
+        v.name:match("Wardrobe"))
+end
+
+local function get_bag_command(k)
+    return res.bags[k].command
+end
+
+local function get_bag_id(bag)
+    return bag.id
+end
+
+local bags = res.bags:filter(filter_bag):key_map(get_bag_command):map(get_bag_id)
 
 local support_npcs = {
     {name = "Orechiniel", zone = 230, menu = 650, buff = 240},
@@ -93,7 +99,7 @@ local support_npcs = {
 }
 
 local exceptions = {
-    ['Terra Crystal'] = 6509,
+    ['Geo Crystal'] = 6509,
 }
 
 local clusters = {
@@ -260,7 +266,7 @@ local function busy_wait(block, timeout, message)
     end
     if os.time() - start >= timeout then
         conditions[block] = false
-        return "timed out - %s":format(message)
+        return "Timed out - %s":format(message)
     else
         inventory = windower.ffxi.get_items()
     end
@@ -358,9 +364,9 @@ local function fetch_ingredient(ingredient)
                 return fetch_ingredient(ingredient)
             end
         end
-        return "unable to locate %s":format(ingredient)
+        return "Unable to locate %s":format(ingredient)
     else
-        return "unknown item %s":format(ingredient)
+        return "Unknown item %s":format(ingredient)
     end
 end
 
@@ -498,8 +504,8 @@ local function check_queue()
             if food then
                 consume_food()
             end
-            local item = queue:pop()
-            local msg = item[1](item[2])
+            local fn, arg = unpack(queue:pop())
+            local msg = fn(arg)
             if msg then
                 error(msg)
             end
@@ -539,12 +545,12 @@ local function handle_status()
 end
 
 local function handle_delay(seconds)
-    local status, err = pcall(tonumber, seconds)
-    if not err then
-        return "invalid delay %s":format(seconds)
+    local n = tonumber(seconds)
+    if n == nil then
+        return "Invalid delay %s":format(seconds)
     else
         notice("Setting delay to %s":format(seconds))
-        delay = seconds
+        delay = n
     end
 end
 
@@ -578,10 +584,10 @@ local function handle_jiggle(key)
 end
 
 local function handle_repeat(count)
-    count = count or 1
-    local status, err = pcall(tonumber, count)
-    if not err then
-        return "invalid count %s":format(count)
+    local count = count or 1
+    local n = tonumber(count)
+    if n == nil then
+        return "Invalid count %s":format(count)
     end
     notice("Adding %d repeat commands to the queue":format(count))
     for i = 1, count do
@@ -592,10 +598,10 @@ local function handle_repeat(count)
 end
 
 local function handle_make(item, count)
-    count = count or 1
-    local status, err = pcall(tonumber, count)
-    if not err then
-        return "invalid count %s":format(count)
+    local count = count or 1
+    local n = tonumber(count)
+    if n == nil then
+        return "Invalid count %s":format(count)
     end
     local recipe = fetch_recipe(item)
     if not recipe then
@@ -620,7 +626,7 @@ local function handle_food(item)
             notice("Setting auto food to %s":format(name.en))
             food = name.en
         else
-            return "invalid food %s":format(item)
+            return "Invalid food %s":format(item)
         end
     end
 end
@@ -629,7 +635,7 @@ local function handle_put(ingredient, bag)
     if bag then
         bag = bag:lower()
         if not bags[bag] then
-            return "unknown bag %s":format(bag)
+            return "Unknown bag %s":format(bag)
         end
     end
     local search = res.items:name(ingredient)
@@ -639,7 +645,7 @@ local function handle_put(ingredient, bag)
         local args = {
             ['id'] = id,
             ['bag'] = bag,
-            ['name'] = name.en,
+            ['name'] = name.english,
         }
         local item = {put, args}
         if bag then
@@ -651,7 +657,7 @@ local function handle_put(ingredient, bag)
         queue:push(item)
         process_queue()
     else
-        return "unknown item %s":format(ingredient)
+        return "Unknown item %s":format(ingredient)
     end
 end
 
@@ -664,20 +670,20 @@ end
 
 local function handle_display()
     if display then
-        notice("Disabling display.")
+        notice("Disabling display")
         display = false
     else
-        notice("Enabling display.")
+        notice("Enabling display")
         display = true
     end
 end
 
 local function handle_support()
     if support then
-        notice("Disabling support.")
+        notice("Disabling support")
         support = false
     else
-        notice("Enabling support.")
+        notice("Enabling support")
         support = true
     end
 end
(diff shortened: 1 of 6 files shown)


z16:
At work, so I can't review it fully yet, but one thing that immediately comes to mind is to significantly condense the recipes file. It uses strings for everything, which increases file size, load times, memory consumption (immensely) and decreases run-time performances. I would make it fully ID-based and optionally single-row (shouldn't matter for parse speed or memory size, but for file size and as such download times, as well as load times for an editor. Just the line breaks and indentations are adding close to 40% of the file size. It would obviously not be human-readable that way, but it doesn't need to be either. When new recipes should be added, they can easily be converted (could even include a short converter tool that people can use, in case you stop maintaining it at some point). Doing those things could cut the file size by 65%, and it would improve run-time performance.

I'll look over the rest of the addon tonight.

Snapster-amb:
Alright.  Recipe uniqueness is based on the ingredients and not the result.  So it can't be completely id based although changing all of the crystal/ingredients to ids and making each recipe a single line should be an easy change.  I will also look into compressing it (simply zipping the recipe files resulted in a 90% reduction in file size.)  I generated all of the recipes using bg wiki, BeautifulSoup, and items.lua in the windower resources.  I will attach the script to the push request later tonight.

alphaONE2:
I think a better way to structure the database would be to use two tables. The first would simply be a sequential list of unique recipes, and the second, a mapping from item ids to a list of recipe indices. When it's time to look up an item, you first look up the list of recipes for that item. If there is only one recipe then use it, otherwise report an ambiguous recipe error and list the possible recipes for the user to choose from.

Byrth:
Having used this, it's actually pretty easy to memorize the different recipe keys (ex. 'aquamarine 2' vs. 'aquamarine',) but I guess moving towards a less ambiguous model ('aquamarine' = error, 'aquamarine 1'/'aquamarine 2'=synth) is also an option.

This addon also doesn't cause performance issues for me, but optimization is always welcomed. I would avoid single-lining it, though, because it's nice to let users add novel recipes.

z16:
I don't think single-lining it would make it harder, changing it to IDs would, which I'd consider the more important change. However, if you're fine with longer load times, you could also pre-process the list on load and convert it to IDs for faster access later. This could even be done off the main thread via `coroutine.yield`, so the user wouldn't even notice it loading longer. And it wouldn't take minutes either, only like a second or two, so the user could still use it right after loading, because that's probably quicker than the user can type the command.

By "ID based" I didn't mean to use IDs for the recipe, this can (and probably should) remain a string, only for the actual item lookup later on. But that wouldn't matter anyway if you go the above route.

And sorry I didn't get to review it last night, something turned up after work so I didn't get to it. Better luck tonight.

Snapster-amb:
Haven't had time to really work on this (might not until Friday not.)  Here's a pastebin of the python script I used to generate the recipes.

https://pastebin.com/5E0zwpSH

There's some jiggery pokery in there due to inconsistencies on bg-wiki's recipe list.  I didn't add it to the push request because I'm not sure what the best deployment strategy is.

I also do think that trying to optimize the recipe lookup is perhaps not worth the effort in regards to performance.  It runs once every 20 seconds or so and I never noticed any slowdown or anything, so the human readability of the recipe list becomes more valuable than the gains from optimizing it.  However, it has been pointed out to me by others not on this review that it might be worth doing to make the usage more robust.  Searching the en/ja/enl/ja1 might be a nice feature I guess.  Are item names different in other regions of the word (french/german/etc.) or are they all english or japanese?

Snapster-amb:
https://github.com/SirAnthony/slpp also this is a dependency for the python script that you can't get on pip or whatever.

addons/craft/craft.lua
```diff
@@ -0,0 +1,738 @@
+--[[
+Craft v1.03312017
```
z16: While I somewhat notoriously abhor the "date as version" method of versioning the code, I've come to live with the fact that some people just want to watch the world burn. But please don't make it worse by using standard date ordering (especially US date ordering)... the year needs to be first, followed by the month, followed by the day. As it is now, a version that's released later (say January 2018) would have a lower version number than the current version. If we did add version comparison to base updates on (like we do with Plugins) it would refuse to update future versions. Please at least change it to `1.yyyymmdd` (or better yet, just use normal versioning).

addons/craft/craft.lua
```diff
@@ -0,0 +1,738 @@
...
+
+local bags = {
+    safe = 1,
+    storage = 2,
+    locker = 4,
+    satchel = 5,
+    sack = 6,
+    case = 7,
+    safe2 = 9
+}
```
z16: I notice that you're using this table to move items between bags. Is there a reason why some bags are omitted? If you do want to include all bags, you can use this code to generate such a table from the resources:
```lua
local bags = res.bags:key_map(function(id) return res.bags[id].command end):map(function(bag) return bag.id end)
```

This has the benefit of still working after they add new bags like Wardrobes 27 & 28. Also reduces error potential due to a manually created mapping.

addons/craft/craft.lua
```diff
@@ -0,0 +1,738 @@
...
+handlers['support'] = handle_support
+handlers['find'] = handle_find
+
+local function handle_command(...)
+    local cmd  = (...) and (...):lower()
+    local args = {select(2, ...)}
+    if handlers[cmd] then
+        local msg = handlers[cmd](unpack(args))
+        if msg then
+            error(msg)
```
z16: Maybe I missed something because I didn't grok the big picture of the program structure yet, but it looks as if you could just `error(...)` in the individual functions, instead of checking for the message and then calling `error` outside of the handler. Or is there something preventing that? I noticed that in other functions as well (`busy_wait`, `check_queue`, `handle_repeat`, maybe others). This would have the benefit of not having to write the check everywhere you call those functions, as well as freeing up the return value to use for proper return values, if needed.

addons/craft/craft.lua
```diff
@@ -0,0 +1,738 @@
...
+handlers['food'] = handle_food
+handlers['status'] = handle_status
+handlers['help'] = handle_help
+handlers['jiggle'] = handle_jiggle
+handlers['support'] = handle_support
+handlers['find'] = handle_find
+
+local function handle_command(...)
+    local cmd  = (...) and (...):lower()
+    local args = {select(2, ...)}
```
z16: You can simplify this by adjusting the signature to:
```lua
local function handle_command(cmd, ...)
    local args = {...}
```

Also, I'd recommend to default the command to 'help', but erroring is also fine, up to you:
```lua
    local cmd = ... and (...):lower() or 'help'
```

addons/craft/craft.lua
```diff
@@ -0,0 +1,738 @@
...
+    notice("paused", paused)
+    notice("display", display)
+    notice("auto food", food)
+    notice("auto support", support)
+    notice("jiggle", jiggle)
+    notice("queue size", queue:length())
+end
+
+local function handle_delay(seconds)
+    local status, err = pcall(tonumber, seconds)
```
z16: `tonumber` can't fail, instead it will return `nil` if the provided number is invalid.

addons/craft/craft.lua
```diff
@@ -0,0 +1,738 @@
...
+                commence_jigglin()
+            end
+            if support then
+                poke_npc()
+            end
+            if food then
+                consume_food()
+            end
+            local item = queue:pop()
+            local msg = item[1](item[2])
```
z16: You could name the table arguments, to be able to specify the names explicitly here. Might make it a bit clearer than using `[1]` and `[2]`, but not really that important either way.

addons/craft/craft.lua
```diff
@@ -0,0 +1,738 @@
...
+        end
+    end
+    local search = res.items:name(ingredient)
+    local id, name = next(search, nil)
+    if id then
+        local msg = nil
+        local args = {
+            ['id'] = id,
+            ['bag'] = bag,
+            ['name'] = name.en,
```
z16: Use `english` please instead of `en`.

z16 (changes requested):
Added a few comments to the code.

As for the Python tool, it's perfectly fine to include it (we just included a Python tool into the *boxdestroyer* addon [today](https://github.com/Windower/Lua/commit/e5016d4ee063eedde8ddab11125f75ca65b876ab)). If you include the third party library as well you need to make sure that the license is compatible with ours and follow its instructions for reproducing the license. Cursory research seems to suggest that it can be included into our code without problems (MIT in BSD) but I'd rather let @alphaONE2 weigh in on it before making that final decision.

And about the names, there are only English and Japanese names at this point, both the German and French versions were discontinued. Fortunately it shouldn't be too hard to make it Japanese-friendly. You can leave the recipes file as it is now and can just convert the user input. So when a user enters a recipe name (I assume they're all item name based, even if there are multiple entries?), you just look up the name in the resources. Then you look up the value in the recipe file by that item's English name.

As for the resources themselves, a minor note on its API: it currently supports indexing with `.en` and `.jp`, as it does in the corresponding resources file. But those are really just internal values and you should use `english` and `japanese` instead. There is also `name` property which resolves to the name based on the client. So using `print(res.items[19747].name)` would print `Mandau` on an English client and `マンダウ` on a Japanese one. You can use this to read and output in a user's corresponding language while internally either working with item IDs, or keep working with English names (as it is now).

And finally a note on performance. It's true that most gains we're talking about are marginal here on most PCs, and even if you don't notice anything you have to keep in mind that some people play on more than 10 year old hardware. Some may not be quite as old, but are using many more addons, and if they also multibox the cost is naturally multiplied. So while it's generally not necessary to adjust for performance (nor would I decline this pull request because of it, fyi, like most things I mention in these reviews they're just suggestions), it should be a goal to strive towards. As long as the cost of improving it is within reasonable margins compared to the gains I think it's worth considering. But I leave that entirely up to your judgment.

That said, the changes I actually do request are, as mentioned in the code:
* The change of the versioning mechanism
* The use of `english` for an English resource, instead of `en`

Everything else is just a note on my part.

addons/craft/craft.lua
Snapster-amb: Wardrobes/temporary/inventory need to be excluded from the list.  I added something that filters the resources now.

addons/craft/craft.lua
Snapster-amb: This is part of the design I opted for because it's simple and because I like the appearance of the code.  Functions can return something, or a string.  If it's a string it indicates something went wrong.

Snapster-amb:
I have submitted a new version with the python scripts + licensed python libraries and some changes to address the feedback.  I plan on reworking the recipes still though, but I haven't had the time to do that yet.

Byrth:
As another note, you need to check the item status to make sure it isn't in bazaar before you try to use it in a recipe / move it / eat it / etc.

alphaONE2:
I just wanted to weigh in on the issue of users adding recipes to the database. I agree that the changes to the database that both I and @z16 have suggested would make it much harder for users to manually add recipes to the database. However, my opinion has always been that requiring users to make manual changes to any kind of internal resource or settings files is bad.

Instead, an in-game interface should be provided to make it simple to modify the database. Ideally user modifications should be kept in a separate file to make updates to the built-in database easier.

Having said that, implementing such a thing is probably more than you wanted to take on right now. I only suggest it as a _possible_ future improvement.

@z16: Including the MIT licenced library here is perfectly fine; as you said, the two are entirely compatible. The MIT licence and the version of the BSD licence we use are essentially identical in terms of what they require and allow, they just use different legal wording.

z16:
Well that all sounds good to me then! @Snapsmojo Is there anything else you want to implement for this PR or is this ready to be merged? I see no issues preventing a merge at this point.

z16 (approved):

Snapster-amb:
There is still some work to do.

* Don't try and use items that are equipped or in bazaar.
* Upon second thought, I can use mog wardrobes, but only for certain ingredients.
* Update to the recipes list format.  I want to add localization (japanese + english) and allow users to search recipes using the english/japanese long format as well.

I imagine I will have this done on Saturday, although there is no reason it can't be accepted in the mean time and another push request opened as these changes won't affect the usage much (only make it more robust).

z16:
All right, I see no harm in merging it now then, we likely won't merge it to live until the weekend anyway, so this will give some people an option to try it out until then for some further feedback.

PR #1553, opened 2017-06-24
New addon closetCleaner

Fox214:
Initial official version


Fox214:
Reuses a lot of gearswap functions not sure if this is better stand alone or merged with gearswap.

z16:
You can consider making this a GS library, if that would help.

Fox214:
I think that's probably the best way to go since it knows the locations of the files and any reused functions are maintained. I just don't know how to do that.

Byrth:
Check out the organizer library for gearswap.

I believe that gearswap functions are exposed as ```gearswap['function name']()```

Fox214:
So I'm trying to make this into a library, first thing I did was put it in the libs directory (next to organizer-lib) I couldn't get that to work because the path couldn't find all the functions from the main gearswap I'm reusing, require '../flow.lua' (etc..) worked for most of them but I was getting an error about some chat file it couldn't see to resolve the path to which was in flow.lua

So at that point I tried to move it up to the main directory and just add a command line call right into gearswap.lua, I got that all working except now when it executes it can't find the include() function (where does this come from?) so it abends on my files since the first line in all my JOB.lua files is include('organizer-lib.lua') 

I had the same issue in my stand alone version but I coded around it be coding a dummy include function which just returns nothing so its basically a noop. The point of trying to integrate is so I don't have to and it will just process through that hopefully people won't have bugs if they run it.

trv6:
There's a little more work needed to convert this to a library (though probably not a lot). Moving it to the libs folder alone isn't enough.

I do not have an in depth knowledge of how your addon works, so I may be off and you'll definitely need to interpret this a little.

When you call `require` or `include` within a Gearswap user file, Gearswap loads the file located at `(gearswap directory)/data/[include path]/<args>` _into the user environment_ (not Gearswap's global environment).
Libraries are standalone scripts which are loaded into user files via `require` and `include` -- to emphasize, your code will end up in the user environment. As a result, your code's calls to `require` are searching the `.../data/[include path]/` directory and can't find the main Gearswap files. This is at the root of the conversion that you will need to do.

You need to account for your code existing within the user environment, which Gearswap refers to as `user_env`. There's actually a huge advantage to this: You tried to grab all of Gearswap's user environment functions (`include`, etc.) so that you could load a user file without getting errors, but you made a slight mistake and copied Gearswap's global environment instead (as a result, functions like `include` continued to throw errors). Once your code exists within `user_env` you won't need to do anything fancy.

Here's my suggestion, step by step:

From the user's perspective:
- The user will grab your code by adding `require('closet_cleaner')` somewhere in their file. This will dump all of the tables, globals, and functions you've defined into `user_env` (essentially copy/pasting your code). As a result, you'll need to take some precautions so that _your_ code doesn't interfere with _their_ code.
- The user will then need a command to run your code. You can add one via the `user_env` function `register_unhandled_command`, which is passed a variable number of arguments (the rest of the command). This is where the majority of your code will be called.

Within your code:

I _think_ your code currently does the following:

- Create a global sets table
- Go through a list of jobs and grab the user file via dofile (I'm pretty sure this dumps the user functions into Gearswap's global table, which is bad)
- Call the now global init_gear_sets or get_sets functions 
- Stash whatever has been added to the sets table

It doesn't look like you actually call any Gearswap functions (I only looked _very_ briefly).

Instead:
Your objective is to iterate through all of the user's files, load the contents, run the get_sets or init_get_sets functions, save the data from the sets table, and then dump everything else.

- Create a local table, say `local closet_cleaner = {}` and stash everything you need in there. This will hide your code from the user, however your eventual function for `register_unhandled_command` will still be able to see it.
- Create a record of equipment which will contain all of the user's sets, populated one file at a time: `closet_cleaner.master_closet = {}`
- Create a function, say `closet_cleaner.get_sets` which does the following:

    - You need a place to dump the contents of the user file, without messing up your closet_cleaner table, the global Gearswap environment, or the user environment: `closet_cleaner.sandbox = {}`.
    - Create your sets table within the sandbox: `closet_cleaner.sandbox.sets = {}`. Now the user's sets will have a spot to sit.
    - Since the `closet_cleaner.sandbox` table is where all of the user's code will end up, it needs access to the Gearswap user environment functions (this is your initial stumbling block). Tell the table to peak into `user_env` when it needs to: `setmetatable(closet_cleaner.sandbox, {__index = gearswap.user_env})` (since you're operating within `user_env` now, `setmetatable(closet_cleaner.sandbox, {__index = _G})` is equivalent, but the first is more explicit, I think).
    - Load the user file into `closet_cleaner.sandbox`. This is where you'll need to dance around Gearswap's code. You need the `loadfile` and `setfenv` functions, but they're not in `user_env`, so you'll need to grab them through `user_env.gearswap`.

    ``` 
    local user_file = gearswap.load_file(<path>) -- if the file is found, this will return the contents of the file wrapped in a function. If it is not found or the file contains an error, this will return `false` and an error message.

    if user_file then -- make sure the function containing the user file exists
       setfenv(user_file, closet_cleaner.sandbox) -- tell Lua to dump the contents of the user file into your sandbox
       user_file() -- run the user file

        -- run closet_cleaner.sandbox.get_sets or closet_cleaner.sandbox.init_get_sets if they exist
        local def_gear = closet_cleaner.sandbox.init_get_sets or closet_cleaner.sandbox.get_sets
        if def_gear then def_gear() end
    end 
    ```

    - `closet_cleaner.sandbox.sets` is now populated. Dump the contents into `closet_cleaner.master_closet`.
    - Wipe the sandbox: `closet_cleaner.sandbox = nil`

You've probably written the functions for dumping the contents of `closet_cleaner.sets` already, which is really the majority of the work. Everything else should be pretty quick.

The structure of your library would be (within Gearswap/libs/closet_cleaner.lua):
```
local closet_cleaner = {}

function closet_cleaner.get_sets = <...> -- as described

register_unhandled_command(function(command, ...)

    if command ~= 'cCleaner' then return end

    local jobs = L{...}

    for job in jobs:it() do
        -- find the filepath, which may be <name>_JOB.lua, JOB.lua, etc. (use pathsearch)
        closet_cleaner.get_sets(path)
    end

    -- now closet_cleaner.master_closet is populated
    -- Process closet_cleaner.master_closet (most of your existing code)
end)
```

The user would call your library with `gs c cCleaner JOB JOB JOB JOB JOB JOB JOB JOB JOB ....`.

trv6:
As a quick note, it looks like you defined `gearswap.pathsearch` in your main file. You can access it without redefining it from `user_env`. You can also also access all of the functions in validate.lua, which look like they'll be useful to you. 

Edit: Ohhh, I see. You had to redefine the file directory. Either way, you won't need to do that. (b'.')b

Fox214:
Ok thanks for all that, but yes the point of trying integrate to gearswap is to use any functions naturally and avoid the screwyness of trying to stitch paths back together. 

Effectively all my code does at a high level is this:
1. loads all data in the users inventory
2. loads all the data stored in the sets = {} tables in the <job>.lua files of the user
3. xrefs the 2 lists above with the resources data and counts up how much and where everything is used, prints an ascii file starting with 0 going upto as high as the user specifies in the config. This way you can dump gear/items you have no use for, or stuff that's used once maybe find an alternate piece to use instead and save an inv. slot. 

With respect to #2 I've created a table called supersets, then as I iterate through each job file which have data like sets.engaged.... sets.precast... (etc...) will all end up in the same table with this structure:
supersets.BLM.sets.engaged...
supersets.BLM.sets.precast....
supersets.BLU.sets.engaged...

(etc....) so Nothing should collide I don't think I can add another table prefix though if you like, and then rest of the code just crawls though the arbitrary nesting of all the tables down to the leafs and counts stuff up. Then its fun with printf and done.  So I think you got the gist of most of it. 

So now when I run this code:
`
```
local cc = {}
cc.sandbox = {}

register_unhandled_command(function(...)
    local cmds = {...}
    for _,v in ipairs(cmds) do
        if S{'closetCleaner','cc'}:contains(v:lower()) then
			setmetatable(cc.sandbox, {__index = gearswap.user_env})
			cc.sandbox.itemsBylongName = T{}
			cc.sandbox.itemsByName = T{}
			cc.sandbox.inventoryGear = T{}
			cc.sandbox.gsGear = T{}
			for k,v in pairs(cc.sandbox.res.items) do
				cc.sandbox.itemsBylongName[res.items[k].enl:lower()] = k
				cc.sandbox.itemsByName[res.items[k].en:lower()] = k
			end
			add_to_chat(122,'got here')
                        -- real call to the run_report function will go here 
			cc.sandbox = nil
        end
    end
    return false
end)
```
`

I'm getting 'attempt to index field 'res' (a nil value) , I can get around that by adding:
res = require 'resources'

but I think I should be inheriting it somehow unfortunately its not immediately clear to me how even after all that, I was hoping the setmetatable() function was the missing ingredient but it does not appear to be. 

(apologies I'm an engineer who's a total hack at programming)

trv6:
I don't think the resources are in the user environment. Try `gearswap.res` instead.

Fox214:
I think that did the trick, I'll try and finish this in the next couple days.

trv6:
It looks like you're mapping item string to id so you can convert the user's sets to ids. There must be a function for that in Gearswap. Maybe you could reuse that, if it's not nested within another function? 

I don't see it after a quick glance, though.

Fox214:
I confess I haven't thought about the actual code in a while, but if i remember correctly I map to the id, and sort on the id so its all consistent, otherwise I could get strange results if someone used a mix and long names and short names.  (because "Meg. Cuirie +1" and "Meghanada Cuirie +1" would count as separate items otherwise, also don't have to worry about case sensitivity (although that's an easy fix))

trv6:
Ehhhh, looks like it would be more complicated to go through Gearswap's functions in this case. They're converted from string name to bag location in `unpack_equip_list`. I guess you could go from name > bag location > id, but that doesn't seem much better.

It also wouldn't work for items that aren't in your bags. Aw well.

Fox214:
I think i tried to reuse as much as seemed reasonable when I originally wrote it, I'm not a big fan of reinventing the wheel. 

Slightly off topic: I wanted to also tell people what could be done with the items in the report (ie AH, bazaar, mule, storage slip X, voucher, etc..) is that data stored anywhere? (I know its on some webpages which I assume comes from the dats but I'm not feeling like writing something to scrape all that data is a good investment of my time)

trv6:
Auction house and bazaar stuff may be stored in the flags field, but I don't know.

Storage slips can be checked with the slips library.

```
require('slips')
slips.get_slip_by_item_id(<item id>)
```
Should do the trick.

Vouchers (event gear storage?) aren't available in the resources or libraries, but the checkstorage script has an up to date list, I think.

alphaONE2:
One problem I see with converting to ID, is you throw out information about augments. Instead I would suggest converting to (bag, index, id) tripples. Since bag + index uniquely identifies a *specific* item (or stack of items) a user owns, this lets you also track augments, the id part of the triple would just be a sanity check value, in case the user moves things unexpectedly and also used to build the final report.

Fox214:
Ok thanks for the tips on getting the extra info and not losing the augments (I'll have to look at mine, I think it considers them all the same item so if you have one you're no longer using, its obvious) 

I think I have the variables all straightened out now, and I think "dofile" is not compatible with the include() function. I was able to prove this by in a working file, I created a 2 line file called FOO.lua in my data directory with my JOB.lua files:
```
print('doing foo')
include('organizer-lib.lua')
```
I then just replaced the include line above (which is line 1 of my lua) with:
`dofile(windower.addon_path..'data/FOO.lua')`

And sure enough I get the include error. Obviously I do not want to execute anything in the lua, I just want to load the function to populate the sets = {} table so I can traverse through it and do the tallying. (I had actually original approached this project in having gearswap dump my inventory, then writing a perl script to parse that and the lua files and generate the report but after I started it made more sense to just leverage the lua data structure) 

Any ideas to replace the "dofile" then? I also tried a dummy wrapper but this did not appear to work:

```
function include(f)
	gearswap.include(f)
end
```

trv6:
Don't use `dofile`. It will load the contents of the file into Gearswap's environment (yikes) rather than `user_env`. The `include` function only exists within `user_env`. In Gearswap's environment, I think it's `include_user`, but you should not attempt to go through the `gearswap` table to get it.

Remember that everything you're doing is in `user_env`, you're simply having difficulty getting the files to load there alongside your code, rather than into Gearswap's environment. The method I described earlier, using `gearswap.loadfile` and `setfenv`, should be sufficient-- unless I goofed (which happens a **lot**).

Fox214:
Ah apologies I missed that amongst all the other info. 

Unfortunately I can't make it work, load_file doesn't exist. 

```
function cc.extract_sets(file) 
	local job_id = cc.sandbox.jobs[file]
	local user_file = gearswap.load_user_files(job_id, file)
	if user_file then 
	    setfenv(user_file, cc.sandbox) 
	    user_file() 
		print("loaded..."..file)

		local def_gear = cc.sandbox.init_get_sets or cc.sandbox.get_sets
		if def_gear then 
			def_gear() 
		end
		return cc.deepcopy(cc.sandbox.sets)
	else	
		print('lua file for '..file..' not found!')
	end 
end
```
I tried include_user() and load_user_files() neither seem to load anything, it just prints the "else" (lua file for ...not found). I think if I can get this to work I should be good.

trv6:
Is `gearswap.loadfile` nil? It's a core Lua function, so it should be there.

Fox214:
nope, I grepped through the whole directory, any of this look right?

```
steve@Ares:/cygdrive/c/games/playonline/windower4
$ find . -name "*.lua" | xargs -i grep -i load {} | grep function | grep -v unload | sort -u                                                                                                                   
    tracker:register_event('reload', function()
-- Can define functions to execute every time the settings are reloaded
-- Load logger settings (has to be after the logging functions have been defined, so those work in the config and related files).
--Added this function for first load on new version. Because of the newly added features that weren't there before.
function config.load(filepath, defaults)
function config.reload(settings)
function global_on_load()
function include_user(str, load_include_in_this_table)
function load_sidecar(job)
function load_user_files(job_id,user_file)
function load_user_globals()
function loadout(loadout_items)
function loadout_get(loadout_items, current_items)
function loadout_tidy(loadout_items, current_items)
function options_load( )
function org.export_loadout(run_type)
function thaw(file_name,bag,loadout)
windower.register_event("load", function()
windower.register_event('load', function()
windower.register_event('load', initialize:cond(function() return windower.ffxi.get_info().logged_in end))
windower.register_event('load', 'logout', 'login', function()
windower.register_event('load', update:cond(function() return windower.ffxi.get_info().logged_in end))
windower.register_event('load',function ()
windower.register_event('load',function()
```

trv6:
It wouldn't be defined in a gearswap file. Try `print (gearswap.loadfile)` in a blank gs file.

trv6:
Errr, sorry. That was worded really poorly. To clarify:
You won't find `loadfile` defined within GearSwap's directory. It's a stock Lua function (like print, dofile, etc.); Byrth didn't write it.
I don't see any trickery in Byrth's code that would prevent access to `loadfile` from `user_env` through the `gearswap` table. i.e. `gearswap.loadfile`.
Sorry for the confusion!

Fox214:
ok sorry been busy last couple days:
print(dofile) returns "function 3342B0C8"
print(loadfile) returns  "nil". 

I created a simple testcase I ran at the command line with cygwin and I got both loadfile and dofile to work, but obviously I can't get it to work inside windower/gearswap (seems like loadfile is missing?). Any other ideas?

trv6:
`print(loadfile)` is expected to print nil. As you guessed, `loadfile` is not in `user_env`. However, it is in Gearswap's global environment, and you should be able to access it through the `gearswap` table.

If `print(gearswap.loadfile)` prints nil, then there's trouble.

Fox214:
Ok that worked (I thought I tried it before, but apparently not or I found a secondary tweak I also had to do) 

The bad news is I think its going into an infinite loop (I either get out of memory error or the game crashes) I think it could be due to executing the same file with require('closetCleaner') in it? Any ideas to deal with that? (there's no loop if I skip the loadfile section of code)

trv6:
Ohhh, oops. It looks like `include_user` doesn't write anything to `package.loaded`, so it would load ClosetCleaner again for each file. I'm not sure that would cause an infinite loop unless you're testing it with the script's code in the main body of the file rather than the unhandled_command event, but it's a quick fix either way.
Within the sandbox, wrap the include/require functions:
```
function cc.include(str)
    if str ~= 'ClosetCleaner' then
        include(str, cc.sandbox)
    end
end

cc.sandbox.include = cc.include
cc.sandbox.require = cc.include
```

That will prevent the closet cleaner library from being reloaded. The second argument to `include` (cc.sandbox) tells GearSwap to put everything into the `cc.sandbox` table, preventing all of the files from ending up in `user_env` (which would probably require people to reload their files after running the closet cleaner command).

Fox214:
Ok so I'm still not getting it, an abbreviated version of what I have (all from global scope of closetCleaner.lua) (with the def_gear() line commented out there is no loop, but of course that's what populates the tables, so I don't expect it to work):

```
-- sets the 'sets' and puts them into supersets based off file name. 
function cc.extract_sets(file) 
	local user_file = gearswap.loadfile(file)
	if user_file then 
	    gearswap.setfenv(user_file, cc.sandbox) 
	    user_file() 
		print(gcnt.." loaded..."..file)
		gcnt = gcnt + 1
		local def_gear = cc.sandbox.init_get_sets or cc.sandbox.get_sets
		if def_gear then 
			-- def_gear() 
		end
		return cc.deepcopy(cc.sandbox.sets)
	else	
		print('lua file for '..file..' not found!')
	end 
end

function cc.include(str)
    if str ~= 'ClosetCleaner' then
        include(str, cc.sandbox)
    end
end

cc.sandbox.include = cc.include
cc.sandbox.require = cc.include

```

Fox214:
I should also add, gcnt is always one, which proves its coming from different calls to the file, that debug statement will come out.

trv6:
Hmm. Ruling out the easy stuff: within the gearswap files you're using to test the library, is it typed `include ('ClosetCleaner')` (same capitalization, no '.lua', etc?).

gcnt is probably being overwritten each time the file loads, unless it's defined 'gcnt = gcnt or 0'.

Fox214:
Yes, could the problem be that I am including it in my User-Globals file? Would I be best off doing it in each one? 

Also I only put it in there to see if it was looping because of something in my program (which should show > 1) or because my program is getting loaded multiple times (then only 1) since its the latter I know that's the issue.  (and oops I had require, but I changed to include same issue)

Fox214:
ok I did get it to work as expected but this work around is a hack:
I put the include in a job I don't care about go to mh change to that job (in this case it was COR) so COR.lua gets loaded (and closetCleaner through the include) and I have 17 other jobs in the ccjobs list and ran it and it worked like a charm.  I'm still at a loss how to make it work in a non-hacky way. 

Shall I open a gearswap branch or email you the file or what?

Fox214:
would it be possible for gearswap to load the include when the command is run then unload it after so its not in any of the JOB.lua files?

trv6:
Include vs require shouldn't matter, as you found. They're the same function. Loading within a shared file also shouldn't matter.

Some print statements might help you trace the issue. I'd start with printing `str` in `cc.include`. Your console will get spammed, but it will allow you to verify that the include wrapper is being called and the file causing the loop.

trv6:
If you commit your work on the library to your clone of windower/Lua/dev, I'll take a look at it on Thursday.

Fox214:
Ok i've commited 2 files in GearSwap/lib

Fox214:
Also I validated my cc.include function is getting called, I'll try and look at this some more tomorrow as well.

trv6:
I think you forgot to push your changes.

Fox214:
Repository: Fox214/Lua
Branch: closetCleaner
Files: addons\Gearswap\libs\ccConfig.lua and closetCleaner.lua 

If that's the wrong spot let me know my github client says they're there.

z16:
@trv6 Since the pull request is closed we don't get updates when @Fox214 pushes changes anymore.

Fox214:
Hi did you ever get a chance to look at this? Sorry I'd been on vacation and it sort of fell off my radar.

PR #1556, opened 2017-07-16
InfoBar Add-on

KenshiDRK:
InfoBar Add-on, replacement of infobar plugin


addons/InfoBar/InfoBar.lua
```diff
@@ -0,0 +1,298 @@
...
+defaults.display.text.font = 'Consolas'
+defaults.display.text.red = 255
+defaults.display.text.green = 255
+defaults.display.text.blue = 255
+defaults.display.text.alpha = 255
+defaults.display.text.size = 12
+
+settings = config.load(defaults)
+
+Notes = {}
```
z16: I don't notice this being used anywhere.

addons/InfoBar/InfoBar.lua
```diff
@@ -0,0 +1,298 @@
...
+
+notes = config.load('data/notes.xml')
+
+box = texts.new("", settings.display, settings)
+
+local infobar = {}
+infobar.new_line = '\n'
+
+windower.register_event('load',function()
+    db = sqlite3.open(windower.addon_path..'\database.db')
```
z16: Should replace the path separator by `/` or `\\`, as `\d` will be treated as an escape sequence. Which doesn't matter in this case, since `\d` in particular is no escape sequence, but should do it properly so renaming is easier. Speaking of which, `database.database` is a bit of a weird name, but nothing I'd really object to, just saying :)

xxxKaiserxxx:
@z16  Can you check it please and add it to windower addons.

addons/InfoBar/InfoBar.lua
```diff
@@ -0,0 +1,303 @@
...
+        (Degrees <= (11.25 * 17) and Degrees > (11.25 * 15) and 'S') or
+        (Degrees <= (11.25 * 19) and Degrees > (11.25 * 17) and 'SSW') or
+        (Degrees <= (11.25 * 21) and Degrees > (11.25 * 19) and 'SW') or
+        (Degrees <= (11.25 * 23) and Degrees > (11.25 * 21) and 'SWW') or
+        (Degrees <= (11.25 * 25) and Degrees > (11.25 * 23) and 'W') or
+        (Degrees <= (11.25 * 27) and Degrees > (11.25 * 25) and 'NWW') or
+        (Degrees <= (11.25 * 29) and Degrees > (11.25 * 27) and 'NW') or
+        (Degrees <= (11.25 * 31) and Degrees > (11.25 * 29) and 'NNW')
+	return direction
+end
```
z16: This function does a lot of branches and calculations, it can be simplified thus:
```lua
local dir_sets = L{'E', 'ESE', 'SE', 'SSE', 'S', 'SSW', 'SW', 'WSW', 'W', 'WNW', 'NW', 'NNW', 'N', 'NNE', 'NE', 'ENE', 'E'}
function DegreesToDirection(val)
    return dir_sets[((val + 8)/16):floor() + 1]
end
```

Note that this function does not take the value in degrees, but in radians, i.e. just `target.facing`. See my comments on `getDegrees` above. As such you may want to rename the function accordingly (or just inline it in the calling function).

addons/InfoBar/InfoBar.lua
```diff
@@ -0,0 +1,303 @@
...
+    notesdb:close()
+end)
+
+function getDegrees()
+    local target = windower.ffxi.get_mob_by_target('st') or windower.ffxi.get_mob_by_target('t') or windower.ffxi.get_mob_by_target('me')
+    local target_heading = V{}.from_radian(target.facing)
+    local angleInDegrees = (math.atan2(target_heading[1], target_heading[2]) * (180 / math.pi))
+    local Degrees = angleInDegrees % 360
+    return math.floor(Degrees)
+end
```
z16: This function can be drastically simplified:
```lua
function getDegrees(value)
    return math.round(360 / math.tau * value)
end
```

Yours returns north for 0°, but the game uses 0° for east. I don't know how InfoBar displayed degrees by value (if it used 0° for north or east or SSW or something), but I think it would make sense to part with convention in favor of game-compatibility, since it doesn't really matter where 0° points to, so it may as well match what the game has.

Just a suggestion though, if you think it makes sense to make 0° be north, then use this:
```lua
function getDegrees(value)
    return math.round(360 / math.tau * (value - math.pi / 2))
end
```

Note that I also changed the signature to get the `facing` value passed in. The `get_mob_by_x` functions are pretty expensive and you do it multiple times on every prerender. I'd suggest having it done once in the calling function (which you already have, the `pos` object), then passing it into both this function and the next one.

addons/InfoBar/InfoBar.lua
```diff
@@ -0,0 +1,303 @@
...
+            get_db(target.name, zone, player.main_job_level)
+        end
+    end
+    box:update(infobar)
+end
+
+windower.register_event('incoming chunk',function(id,org,modi,is_injected,is_blocked)
+    if id == 0xB then
+        zoning_bool = true
+    elseif id == 0xA and zoning_bool then
```
z16: Can remove the `and zoning_bool` check.

z16 (approved):
Wow, sorry for the long delay, I have written the review but somehow forgot to submit it... thanks for reminding me @xxxKaiserxxx .

addons/InfoBar/InfoBar.lua
KenshiDRK: I guess I forgot to submit the changes where I added the notes funtionality into another database.

addons/InfoBar/InfoBar.lua
KenshiDRK: I have a problem with this function so thats why I use the other one, with that function the facing is a little off, for example, when facing S (90º) it shows as SW.

z16: Hm, I just noticed this goes from 0 to 255, instead of proper radians... that's how it's transmitted in certain packets. To make it work for radians (which is what `target.facing` returns), change the scaling like this:
```lua
local dir_sets = L{'W', 'WNW', 'NW', 'NNW', 'N', 'NNE', 'NE', 'ENE', 'E', 'ESE', 'SE', 'SSE', 'S', 'SSW', 'SW', 'WSW', 'W'}
function DegreesToDirection(val)
    return dir_sets[math.round((val + math.pi) / math.pi * 8) + 1]
end
```

KenshiDRK: thats returning nil now D:

z16: Are you sure you used the right value, the one returned by `mob.facing` as input?

Here's some sample code I wrote:
image

And here is the result:
dir

(`target.me.facing` up there is the same as `windower.ffxi.get_mob_by_target('me').facing`.)

KenshiDRK: sorry, it was an error on my end, will update the pull request soon.

addons/InfoBar/InfoBar.lua
```diff
@@ -0,0 +1,303 @@
...
+                MOB_infobar.isnm = isnm == 1 and 'NM' or 'No NM'
+                MOB_infobar.isfishing = isfishing == 1 and 'F' or 'NF'
+                MOB_infobar.detect = (sight == 1 and 'S' or '')..
+                    (sight == 1 and sound == 1 and ',H' or sound == 1 and 'H' or '')..
+                    ((sight == 1 or sound == 1) and magic == 1 and ',M' or magic == 1 and 'M' or '')..
+                    ((sight == 1 or sound == 1 or magic == 1) and lowhp == 1 and ',HP' or lowhp == 1 and 'HP' or '')..
+                    ((sight == 1 or sound == 1 or magic == 1 and lowhp == 1) and healing == 1 and ',R' or healing == 1 and 'R' or '')..
+                    ((sight == 1 or sound == 1 or magic == 1 and lowhp == 1 and healing == 1) and ts == 1 and ',TS' or ts == 1 and 'TS' or '')..
+                    ((sight == 1 or sound == 1 or magic == 1 and lowhp == 1 and healing == 1 and ts == 1) and th == 1 and ',TH' or th == 1 and 'TH' or '')..
+                    ((sight == 1 or sound == 1 or magic == 1 and lowhp == 1 and healing == 1 and ts == 1 and th == 1) and scent == 1 and ',Sc' or scent == 1 and 'Sc' or '')
```
z16: Just saw this and it can be simplified as well:
```lua
                local detect = L{
                    sight == 1 and 'S' or '',
                    sound == 1 and 'H' or '',
                    magic == 1 and 'M' or '',
                    lowhp == 1 and 'HP' or '',
                    healing == 1 and 'R' or '',
                    ts == 1 and 'TS' or '',
                    th == 1 and 'TH' or '',
                    scent == 1 and 'Sc' or '',
                }
                MOB_infobar.detect = detect:filter(-''):concat(',')
```

z16 (approved):
Just left another minor note, but as before, it looks good. Let me know if you still wanna look into that, or to just pull it.

addons/InfoBar/InfoBar.lua
KenshiDRK: I'll update pr soon with this and a nil error fix for when a mob have no levelmin and levelmax on database.

PR #1558, opened 2017-08-20
closetCleaner lib

Fox214:
library addon to report gear usage across specified job files.


Fox214:
any idea when this might get looked at?

addons/GearSwap/libs/ccConfig.lua
```diff
@@ -0,0 +1,30 @@
+-- Jobs you want to execute with, recomment put all active jobs you have lua for will look for <job>.lua or <playername>_<job>.lua files
```
z16: I don't know how GS usually does this, but normally user files would go into the `GearSwap/data/` folder. That folder is special in that it won't be updated and overwritten. For example, if you ever wanted to update this file with new default values, the file would be overwritten with the new default file for all users who modified it.

Two options you can do then, either move this file to the aforementioned `data` folder, then it will be downloaded initially but ignored in any future updates (can make a subfolder for cc there as well), or forgo this file entirely. I know you want to show people how to use cc through this file, but I think that would work better if you added a blank file to the `data` folder, and keep this file here as an example, and name it as such (possibly refer to it in a readme or w/e). That way people trying it for the first time without looking at the config file will not experience weird behavior, which they will complain to you (and us) about in the end.

addons/GearSwap/libs/ccConfig.lua
```diff
@@ -0,0 +1,30 @@
+-- Jobs you want to execute with, recomment put all active jobs you have lua for will look for <job>.lua or <playername>_<job>.lua files
+ccjobs = { 'BLM', 'BLU', 'BRD', 'BST', 'DRG', 'GEO', 'MNK', 'NIN', 'PLD', 'RDM', 'RNG', 'RUN', 'SAM', 'SCH', 'SMN', 'THF', 'WAR', 'WHM' }
+-- Put any items in your inventory here you don't want to show up in the final report
+-- recommended for furniture, food, meds, pop items or any gear you know you want to keep for some reason
+-- supports lua regex syntax, put each separate one as its own entry in the table
+-- see: https://www.lua.org/pil/20.2.html
```
z16: I would suggest using wildcard matching instead of Lua's string pattern matching. You would need to use the `windower.wc_match` function instead of `string.match` below, but people are more familiar with it and it's generally simpler to use (you actually used it by accident on `"Deed of*"` below, which is not a valid Lua string pattern (or well, it is, but probably not the one you intended it to be)). The wildcard matching is in many cases also more powerful, since it allows alternation with `|`, which Lua string patterns do not.

addons/GearSwap/libs/ccConfig.lua
```diff
@@ -0,0 +1,30 @@
+-- Jobs you want to execute with, recomment put all active jobs you have lua for will look for <job>.lua or <playername>_<job>.lua files
+ccjobs = { 'BLM', 'BLU', 'BRD', 'BST', 'DRG', 'GEO', 'MNK', 'NIN', 'PLD', 'RDM', 'RNG', 'RUN', 'SAM', 'SCH', 'SMN', 'THF', 'WAR', 'WHM' }
+-- Put any items in your inventory here you don't want to show up in the final report
+-- recommended for furniture, food, meds, pop items or any gear you know you want to keep for some reason
+-- supports lua regex syntax, put each separate one as its own entry in the table
+-- see: https://www.lua.org/pil/20.2.html
+ccignore = { "Rem's Tale*", "Storage Slip %d+", "Deed of*", "%a+ Virtue", "Dragua's Scale",
+			"Glittering Yarn", "Dim. Ring*", "Cupboard", ".*VCS.*", ".*Abjuration.*", "%a+ Organ",
```
z16: Use spaces instead of tabs please. Superficially it's not too important, but we've been treating it as a policy. You can configure most editors to use spaces when you press the tab key.

The same goes for the other file.

addons/GearSwap/libs/closetCleaner.lua
```diff
@@ -0,0 +1,379 @@
+--Copyright (c) 2016~2017, Brimstone
```
z16:
```
--Copyright © 2016-2017, Brimstone
```

addons/GearSwap/libs/closetCleaner.lua
```diff
@@ -0,0 +1,379 @@
...
+
+--Redistribution and use in source and binary forms, with or without
+--modification, are permitted provided that the following conditions are met:
+
+--    * Redistributions of source code must retain the above copyright
+--      notice, this list of conditions and the following disclaimer.
+--    * Redistributions in binary form must reproduce the above copyright
+--      notice, this list of conditions and the following disclaimer in the
+--      documentation and/or other materials provided with the distribution.
+--    * Neither the name of <addon name> nor the
```
z16: Use `closetCleaner` or whatever you want this library's "official" name to be.

addons/GearSwap/libs/closetCleaner.lua
```diff
@@ -0,0 +1,379 @@
...
+--      names of its contributors may be used to endorse or promote products
+--      derived from this software without specific prior written permission.
+
+--THIS SOFTWARE IS PROVIDED BY THE COPYRIGHT HOLDERS AND CONTRIBUTORS "AS IS" AND
+--ANY EXPRESS OR IMPLIED WARRANTIES, INCLUDING, BUT NOT LIMITED TO, THE IMPLIED
+--WARRANTIES OF MERCHANTABILITY AND FITNESS FOR A PARTICULAR PURPOSE ARE
+--DISCLAIMED. IN NO EVENT SHALL <your name> BE LIABLE FOR ANY
```
z16: Use `Brimstone` here.

addons/GearSwap/libs/closetCleaner.lua
```diff
@@ -0,0 +1,379 @@
...
+cc.sandbox = {}
+cc.sandbox.windower = setmetatable({}, {__index = windower})
+cc.sandbox.windower.register_event = functions.empty()
+cc.sandbox.windower.raw_register_event = functions.empty()
+cc.sandbox.windower.register_unhandled_command = functions.empty()
+
+register_unhandled_command(function(...)
+    local cmds = {...}
+    for _,v in ipairs(cmds) do
+        if S{'closetCleaner','cc'}:contains(v:lower()) then
```
z16: If you do `v:lower()` it will never match `closetCleaner`. Should lowercase that as well.

addons/GearSwap/libs/closetCleaner.lua
```diff
@@ -0,0 +1,379 @@
...
+			setmetatable(cc.sandbox, {__index = gearswap.user_env})
+			cc.sandbox.itemsBylongName = T{}
+			cc.sandbox.itemsByName = T{}
+			cc.sandbox.inventoryGear = T{}
+			cc.sandbox.gsGear = T{}
+			for k,v in pairs(gearswap.res.items) do
+				cc.sandbox.itemsBylongName[gearswap.res.items[k].enl:lower()] = k
+				cc.sandbox.itemsByName[gearswap.res.items[k].en:lower()] = k
```
z16: A note on using our resources. Please use `english` and `english_log` instead of `en` and `enl`, despite what it shows in the file. We use the short version in the file primarily to save on file size since the resources can get huge in some cases.

Also, I don't know how well that works with other GearSwap features, but consider using `name` and `name_log` instead of either. That will make it so that Japanese players can write their item names in Japanese and it will still match them, since `name` expands to `english` for English clients, and `japanese` for Japanese ones.

addons/GearSwap/libs/closetCleaner.lua
```diff
@@ -0,0 +1,379 @@
...
+    return function()
+        i = i + 1
+        if keys[i] then
+            return keys[i], t[keys[i]]
+        end
+    end
+end
+
+-- recursively copies a table into a new table, this avoids dealing with references so proper tallies can be done
+function cc.deepcopy(orig)
```
z16: Fwiw there's already a deepcopy function in our tables library, which works on `T{}`-tables by just writing `orig:copy(true)` (`true` indicated a deep instead of shallow copy). For non-`T{}`-tables you can do `table.copy(orig, true)`.

addons/GearSwap/libs/closetCleaner.lua
```diff
@@ -0,0 +1,379 @@
...
+    local keys = {}
+    for k in pairs(t) do keys[#keys+1] = k end
+
+    -- if order function given, sort by it by passing the table and keys a, b,
+    -- otherwise just sort the keys 
+    if order then
+        table.sort(keys, function(a,b) return order(t, a, b) end)
+    else
+        table.sort(keys)
+    end
```
z16: We do not have a sorted iterator, but the above few lines could be replaced by our library functions:
```lua
local keys = t:keyset():sort()
```

Or, if `t` is not a `T{}`-table:
```lua
local keys = table.sort(table.keyset(t))
```

z16 (changes requested):
I added a review, but keep in mind that I haven't been playing since GS was a thing and never used it or any libraries for it, so my knowledge on it is very limited and I didn't actually test your code. I basically only commented on Lua-specific things I noticed or things regarding the Windower API or overall design. I'll trust that you know and use GS well enough yourself to know if this works or doesn't :)

Most of the things I said were just suggestions, the only changes we need to have are the changes to the copyright notice I mentioned and the tabs thing. I'll leave the rest up to your discretion.

addons/GearSwap/libs/closetCleaner.lua
z16: Actually I just realized the copy is deep by default (i.e. you have to explicitly specify `false` to make it shallow). So just writing `orig:copy()` should do (or `table.copy(orig)` for non-`T{}`-tables).

Fox214:
Apologies for taking a month to get back to this, I've taken your first few recommendations, I was not able to get the windower deep copy or table sorts to work so I left it alone.

I thought I had np++ configed for space tabs, but Apparently not, I've fixed it, but will only slowly change the code. Guess that's what happens when I code on 3-4 computers. 

Also I've noticed a bug that the Mote-Treasurehunter seems to try to do stuff after the files are executed. (its at the bottom of the file windoer.raw_register_event(action, on_action_for_th) where there is uninitialized values, just doing //lua r gearswap fixes it, but it echos errata to the console otherwise. Not sure how to fix it.

trv6:
You can replace all tabs with spaces in notepad++ pretty quickly.
Press `<ctrl h>` to open the replace menu. Select the extended search mode button, then replace all instances of `\t` with `    ` (four spaces).

z16:
Unfortunately I'm not proficient enough with GS to answer that. A bit curious why our `copy`/`sort` isn't working for you, would be cool if you could give an example,

And feel free to commit the new changes you made and push them to your remote branch, they will automatically appear in this pull request, then we can keep reviewing it.

Fox214:
I got the deep copy to work, I didn't bother with the other for now at 
least because it looked more complex and my brain hurts.

I fixed the tabs (can't believe I didn't think to use a regex).

Also fixed a functional bug in the code, and recommited to dev let me 
know if you can't see it.

------ Original Message ------

Fox214:
Does this look ok now?

z16:
Oh sorry, I had the tab open but forgot to look at it... I'll check it out when I wake up.

z16:
I looked through it and it seems largely the same as it was before. Are you sure you committed everything? It needs to be committed and pushed to the same branch from which you opened the pull request (`Fox214:dev`).

Fox214:
I see them here: 
https://github.com/Windower/Lua/pull/1558/commits/27c468ff47437915179b255ab475fe0d6ec9f599

Says this:
          closetCleaner lib #1558
Open
Fox214 <https://github.com/Fox214> wants to merge 3 commits into 
Windower:dev from Fox214:dev

------ Original Message ------

addons/GearSwap/libs/closetCleaner.lua
```diff
@@ -265,7 +265,6 @@ function cc.list_sets(t, f)
 							if write_sets[itemid] == nil then
```
cairthenn: tabs :cry:

Fox214:
my bad, i think I referenced the wrong commit, I had a small bug fix and 
all the tabs fixed here:
https://github.com/Windower/Lua/pull/1558/commits/b042ea247e5ac7d50fdcc10704aad78e27a6e149


------ Original Message ------

addons/GearSwap/data/ccConfig.lua
```diff
@@ -0,0 +1,14 @@
...
+-- recommended for furniture, food, meds, pop items or any gear you know you want to keep for some reason
+-- use * for wildcard matching. see ccConfig_example.lua for examples
+ccignore = { }
+-- This is the most use of an item you want to show up in the report
+-- Set to nil or delete for unlimited
+ccmaxuse = nil
+-- List bags you want to not check against, needs to match "Location" column in <player>_report.txt
+skipBags = S{ 'Temporary' }
+-- this prints out the _sets _ignored and _inventory files
+ccDebug = false
```
z16: This file shouldn't be there. I'd trust people to be competent enough to look at the example config and figure it out, although I would also move that out of the data folder. The problem with the data folder is that anything in there will never be updated. So if you ever adjust or expand the mechanism people's files will not be overwritten (because they're meant to be *user* created/edited files). This includes the example file.

addons/GearSwap/data/ccConfig.lua
```diff
@@ -0,0 +1,14 @@
...
+ccjobs = { 'BLM', 'BLU', 'BRD', 'BST', 'COR', 'DRG', 'DNC', 'DRK', 'GEO', 'MNK', 'NIN', 'PLD', 'PUP', 'RDM', 'RNG', 'RUN', 'SAM', 'SCH', 'SMN', 'THF', 'WAR', 'WHM' }
+-- Put any items in your inventory here you don't want to show up in the final report
+-- recommended for furniture, food, meds, pop items or any gear you know you want to keep for some reason
+-- use * for wildcard matching. see ccConfig_example.lua for examples
+ccignore = { }
+-- This is the most use of an item you want to show up in the report
+-- Set to nil or delete for unlimited
+ccmaxuse = nil
+-- List bags you want to not check against, needs to match "Location" column in <player>_report.txt
+skipBags = S{ 'Temporary' }
```
z16: All the other values in here have `cc` prefixed (why?), but this one doesn't. Is this an oversight?

addons/GearSwap/data/ccConfig_example.lua
```diff
@@ -0,0 +1,30 @@
...
+            ".*Signet Staff", "Capacity Ring", "Facility Ring", "Trizek Ring", "Bam. Grass Basket", "Portafurnace",
+            "%a+'s Sign", "%a+'s Apron", "Toolbag .*", "%a+ Crepe", "%a+ Sushi.*", ".* Stable Collar", "Plovid Effluvium",
+            "Gem of the %a+", "Inoshishinofuda", "Ichigohitofuri", "Chonofuda", "Shikanofuda", "Shihei", "Midras.s Helm .1", "Cobra Staff", "Ram Staff", "Fourth Staff",
+            "Warp Ring", "Chocobo Whistle", "Warp Cudgel", "%a+ Broth", "Tavnazian Ring", "Official Reflector",
+            "Pashhow Ring", "Dredger Hose", "Trench Tunic", "Sanjaku.Tenugui", "Katana.kazari", "Sanjaku.Tenugui",
+            "Carver's Torque", "Kabuto.kazari", "Etched Memory", "Trbl. Hatchet", "Maat's Cap", "Trbl. Pickaxe",
+            "Olduum Ring", "Linkpearl", "Caliber Ring", "Vana'clock", "Signal Pearl", "Windy Greens", "Nexus Cape",
+            "Wastebasket", "Sickle", "Carpenter's Gloves", "Shinobi.Tabi", "Guardian Board", "%a+ Pie", "Pet Food %a+",
+            "Reraise %a+", "P. %a+ Card", "Diorite", "Abdhaljs Seal", "Kin's Scale", "*Curry*", "Crepe B. Helene",
```
z16: Personally I would use one matching mechanism, i.e. `windower.wc_match` or `string.match`, not mix both. I'd recommend wildcard matching over Lua patterns when dealing with users.

addons/GearSwap/libs/closetCleaner.lua
```diff
@@ -0,0 +1,367 @@
+--Copyright (c) 2016-2017, Brimstone
```
z16: Please adjust `(c)` to `©`. Are you `Brimstone`, or is that a copy/paste oversight?

addons/GearSwap/libs/closetCleaner.lua
```diff
@@ -0,0 +1,367 @@
...
+--(INCLUDING, BUT NOT LIMITED TO, PROCUREMENT OF SUBSTITUTE GOODS OR SERVICES;
+--LOSS OF USE, DATA, OR PROFITS; OR BUSINESS INTERRUPTION) HOWEVER CAUSED AND
+--ON ANY THEORY OF LIABILITY, WHETHER IN CONTRACT, STRICT LIABILITY, OR TORT
+--(INCLUDING NEGLIGENCE OR OTHERWISE) ARISING IN ANY WAY OUT OF THE USE OF THIS
+--SOFTWARE, EVEN IF ADVISED OF THE POSSIBILITY OF SUCH DAMAGE.
+
+-- _addon.name = 'closetCleaner'
+-- _addon.version = '0.9'
+-- _addon.author = 'Brimstone'
+-- _addon.commands = {'cc','closetCleaner'}
```
z16: Was this originally its own addon, and that's why it's commented out? Any reason to keep it around?

addons/GearSwap/libs/closetCleaner.lua
```diff
@@ -0,0 +1,367 @@
...
+
+local cc = {}
+cc.sandbox = {}
+cc.sandbox.windower = setmetatable({}, {__index = windower})
+cc.sandbox.windower.register_event = functions.empty()
```
z16: Are you trying to disable `register_event`? If so, you'd want `functions.empty`, not `functions.empty()`. The latter is the same as using `nil`.

Any particular reason to do that? I would probably throw an error instead, so people actually know not to use it, instead of nothing happening:
```lua
cc.sandbox.windower.register_event = error:apply('Registering events disabled for CC')
```

addons/GearSwap/libs/closetCleaner.lua
```diff
@@ -0,0 +1,367 @@
...
+cc.sandbox.windower.raw_register_event = functions.empty()
+cc.sandbox.windower.register_unhandled_command = functions.empty()
+
+register_unhandled_command(function(...)
+    local cmds = {...}
+    for _,v in ipairs(cmds) do
```
z16: Are you sure this is what you want? This would trigger even if `cc` was a third parameter in an unhandled command that was meant to be handled by something else.

I'd change the function signature and setup to this:
```lua
register_unhandled_command(function(command)
    command = command and command:lower() or nil
    if command ~= 'cc and command ~= 'closetcleaner' then
        return
    end

    --[[ rest of the function ]]
end)
```

addons/GearSwap/libs/closetCleaner.lua
```diff
@@ -0,0 +1,367 @@
...
+            cc.sandbox.itemsByName = T{}
+            cc.sandbox.inventoryGear = T{}
+            cc.sandbox.gsGear = T{}
+            for k,v in pairs(gearswap.res.items) do
+                cc.sandbox.itemsBylongName[gearswap.res.items[k].name_log:lower()] = k
+                cc.sandbox.itemsByName[gearswap.res.items[k].name:lower()] = k
+            end
+            cc.sandbox.jobs = {}
+            for k,v in pairs(gearswap.res.jobs) do
+                cc.sandbox.jobs[gearswap.res.jobs[k].ens] = k
```
z16: Please use `english_short` instead of `en`.

addons/GearSwap/libs/closetCleaner.lua
```diff
@@ -0,0 +1,367 @@
...
+            end
+            if not windower.dir_exists(windower.addon_path..'report') then
+                windower.create_dir(windower.addon_path..'report')
+            end
+            local path = windower.addon_path:gsub('\\','/')
```
z16: Our of curiosity, did you have issues using `\` over `/`?

addons/GearSwap/libs/closetCleaner.lua
```diff
@@ -0,0 +1,367 @@
...
+            require 'ccConfig'
+            cc.run_report(path)
+            cc.sandbox = nil
+        end
+    end
+    cc.sandbox = {}
+    cc.sandbox.windower = setmetatable({}, {__index = windower})
+    cc.sandbox.windower.register_event = functions.empty()
+    cc.sandbox.windower.raw_register_event = functions.empty()
+    cc.sandbox.windower.register_unhandled_command = functions.empty()
```
z16: See above about `function.empty`.

addons/GearSwap/libs/closetCleaner.lua
```diff
@@ -0,0 +1,367 @@
...
+    return true
```
z16: Return values for this function are ignored, safe to remove.

addons/GearSwap/libs/closetCleaner.lua
```diff
@@ -0,0 +1,367 @@
...
+        cc.print_break(f2, form)
+    end
+    for k,v in cc.spairs(cc.sandbox.gsGear, function(t,a,b) return t[b] > t[a] end) do
+        if ccmaxuse == nil or v <= ccmaxuse then
+            printthis = 1
+            if not cc.job_used[k] then
+                cc.job_used[k] = " "
+            end
+            for i,s in ipairs(ccignore) do
+                if windower.wc_match(gearswap.res.items[k].en, s) or string.match(gearswap.res.items[k].en, s) then
```
z16: At this and other points you use `en` instead of `english` or `name` (same with `english_log` and `name_log` instead of `enl`). And see the comment above about recommending only one matching algorithm. If you go for `name` over `english`, try to do so consistently.

z16 (changes requested):
Looks like some of your changes didn't make it in or they got lost somewhere along the way. I reviewed it again and mentioned a few things. As before, most of what I said is optional, but I'd like you to adjust the copyright notice and resource usage before I accept this. And I will also highly recommend to remove the default `ccConfig.lua` from the `data` repository, see the specific comment for details.

Fox214:
Ok, I think I missed 1 or two updates from your original note (en vs 
english) not sure how. I have them fixed now.

What should I do with the config file, remove ccConfig.lua and move the 
example one to what directory? (once this is answered i'll recommit, i 
haven't pushed any changes)

I am Brimstone (well it was my original character's name).

The cc.sandbox.windower.register_event= error:apply('Registering events 
disabled for CC') did not work (it crashes on lua with register events 
in them) but removing the () worked and I believe fixed a bug so I think 
that is good.

I'm a unix guy backslash is an escape character and not a path delimiter 
:). Yeah I know the game is in windows, I don't recall if i had issues 
or I just did that to appease my sensibilities (obviously its working 
though).

------ Original Message ------

z16:
Yeah, both `\` and `/` work, was just curious.

I would recommend using the `config` library for configurations in the data folder. It will automatically create the file from default settings you pass into it, if it's missing, and can even amend single settings if you add new ones in the future.

addons/GearSwap/libs/closetCleaner.lua
trv6: The error might lead to confusion when people run the script for the first time. 

The library only loads the user files to get to the gear sets, but `register_event` (and a few other functions, I think?) would make persistent changes within the environment of the user's current gearswap file. If they weren't blocked, you'd have to reload gearswap after running the script, but there's not much need to know that the library is disabling them within its own sandbox.

Fox214:
Ok 1 question about the config library, since its in gearswap/data 
folder and gets named settings.xml what happens if gearswap or someone 
else creates a similar library for gearswap that uses the config library 
will they conflict?

------ Original Message ------

z16:
You can specify a filename:
```lua
settings = config.load('ccConfig.xml', defaults) -- or whatever
```

Otherwise yes, it would conflict.

Fox214:
Well I updated to the xml file, but i really hate the way the ccignore 
list works, because this is likely to be a long list for a lot of folks.

<ccignore>
     <1>someth*ng</1>
     <2>foo*bar</2>
</ccignore>

Is a lot more verbose to edit than:
  ccignore = {"someth*ng", "foo*bar"}

(7:2 character ratio not to mention matching/non-repeat issue, I 
wouldn't want to type out 50 items that way)

I tried using it as S{} table that fixes the Xml but just doesn't work 
in the lua then.

Any ideas?

------ Original Message ------

z16:
It should work fine with `S{}` tables (and `L{}`-tables for that matter), other addons use that too (like Treasury) and it results in a normal, comma-separated list of strings. I would try to figure out why that doesn't work for you. If you want you [gist](https://gist.github.com/) the code you have that isn't working for you, then I can have a look.

Fox214:
I was able to figure it out new commit here: 
https://github.com/Windower/Lua/pull/1558/commits/20fc7e50d6a42a21bb4fa3441ab714e9cd3e0d91

(had to change the iterator with S{}) I think its ok now.

------ Original Message ------

Fox214:
Awesome just out of curiosity any idea when it might be live?

------ Original Message ------

z16:
Now :)

PR #1560, opened 2017-08-28
Fixes for bluguide

Chiaia:


addons/bluguide/bluguide.lua
```diff
@@ -6,7 +6,7 @@ traitboxes = require("ui/traitboxes")
 spellboxes = require("ui/spellboxes")
 setspells = require("masterlist")
 
-_addon.version = '1.1'
+_addon.version = '1.2 Aug 28th, 2017'
```
z16: Would appreciate it if you could keep the versioning scheme of the original author.

z16 (changes requested):

addons/bluguide/bluguide.lua
Chiaia: Ah ok I was trying to change it so it more noticeable. All these addons update and most people don't notice because its just a bump from 1.x to 1.y but will do

PR #1564, opened 2017-09-20
Support for subjob macro sets

eeach:
Add detection of subjob and selection of macro sets based on
  both main job and subjob.

Change configuration to use the windower config library.

Reduce the size of the default configuration. Addon no longer
  crashes when searching for an absent key.

On branch macrochanger-update
Changes to be committed:
	deleted:    addons/macrochanger/data/settings.txt
	new file:   addons/macrochanger/data/settings.xml
	modified:   addons/macrochanger/macrochanger.lua


addons/macrochanger/data/settings.xml
```diff
@@ -0,0 +1,21 @@
+<?xml version="1.1" ?>
```
z16: You should remove this file, it will be auto-generated on the first run with the default settings. If you ever update the default settings, this file will not be updated and any new settings will not be copied over. Instead the config library's own mechanism will add new default keys automatically, so you should just use the existing mechanism.

addons/macrochanger/macrochanger.lua
```diff
@@ -26,143 +26,85 @@
...
-        SCH = {Book = '', Page = ''},
-        GEO = {Book = '', Page = ''},
-        RUN = {Book = '', Page = ''},
-        }
-	options_load()
-end)
+function sleep (n)
+  local t0 = os.clock()
+  while os.clock() - t0 <= n do end
+end
```
z16: We have a sleep function you can use: `coroutine.sleep(2)`. That way you don't have to busy wait.

addons/macrochanger/macrochanger.lua
```diff
@@ -26,143 +26,85 @@
...
+     windower.add_to_chat(17, 'disableall [on|off]')
+     windower.add_to_chat(17, '   on - Disables all automated macro switching')
+     windower.add_to_chat(17, '   off - Enables all automated macro switching not disabled individually')
+     windower.add_to_chat(17, '   Resets to what is stored in settings upon unloading of addon.  To Permanently change, please change the option in the settings file.')
+   end
 end
 
-windower.register_event('job change',function ()
+function job_change ()
 -- Could use the job ID passed into this function, but the addon would have to include the resources library
```
z16: This is not your fault, since the comment indicates it has been here for a while, but you should definitely use the provided IDs in combination with our resources. It's very light if you only use jobs and much more efficient than calling `get_player`, especially twice. The results of this function should be cached when possible, so if you do decide to do it this way, save it in a local variable first. But like I said, best to use the provided arguments.

addons/macrochanger/macrochanger.lua
```diff
@@ -26,143 +26,85 @@
...
-        if job and macros[job] then
-			book = macros[job].Book
-			page = macros[job].Page
-		end
+  local mjob = windower.ffxi.get_player().main_job
+  local sjob = windower.ffxi.get_player().sub_job
+  local job  = ''
+  local book = ''
+  local page = ''
+  if settings.globaldisable == 0 then
```
z16: This looks like a boolean setting, so you can use a boolean variable. Will also make it easier on users, since this is written to their user files, so they can use `true` and `false` instead of `1` and `0`.

addons/macrochanger/macrochanger.lua
```diff
@@ -26,143 +26,85 @@
...
-	if globaldisable == 0 then
-        if job and macros[job] then
-			book = macros[job].Book
-			page = macros[job].Page
-		end
+  local mjob = windower.ffxi.get_player().main_job
+  local sjob = windower.ffxi.get_player().sub_job
+  local job  = ''
+  local book = ''
+  local page = ''
```
z16: In Lua it's enough to just declare local variables, you don't need to provide values that will always be overwritten. Also, all of these could be declared in inner scope, book and page even when they're used:
```lua
    local page = --[[...]]
    local book = --[[...]]
```

addons/macrochanger/macrochanger.lua
```diff
@@ -26,143 +26,85 @@
...
+  if settings.globaldisable == 0 then
+    if mjob and sjob and settings.macros[(mjob..sjob):lower()] then
```
z16: I would suggest using an underscore to divide the jobs in the settings file, i.e. `mjob .. '_' .. sjob`. That will make the settings file look better and job combinations easier to find, and we use similar syntax for other heavily used addons like *GearSwap*.

addons/macrochanger/macrochanger.lua
```diff
@@ -26,143 +26,85 @@
...
-	elseif globaldisable == 1 then
+function login (name)
+  local defaults = {
+    character = '',
+    globaldisable = 0,
+    macros = {
+      war = {book = 1, page = 1},
+    },
+  }
+  settings = config.load(defaults)
```
z16: Setting the default settings should happen on file scope, not on the login event. It will automatically reparse the settings when a new character logs in, so you don't need to do it manually.

addons/macrochanger/macrochanger.lua
```diff
@@ -26,143 +26,85 @@
...
+  if name then
```
z16: Is there ever a situation when `name` is `nil`? If so that's likely a bug, let us know if you've encountered that and under which circumstances, if possible.

z16 (changes requested):
Added a few remarks, most of it looks ok, but please use 4 space indentation throughout the file. Most of what I wrote are just suggestions to leverage existing functions which you may have not been aware of, but it's up to you.

There is one issue that's a bit unsettling, namely that this update is incompatible with previous configurations. That means as soon as we roll it out to users it will stop working for them until they copy the settings over. It's not a huge drawback, but something you should take into account. It can be as simple as informing them: If you don't read any settings other than the defaults, print a line to the log saying that they need to set it up in the settings file (preferably with a path to the file).

In the same vein, I would remove the default settings for `WAR`. I assume you added them to show an example of how it works. While that works, I'd suggest writing a short `ReadMe.md`, which you can then refer to. This will also be displayed when users click on "About" under the addon in the launcher menu.

One final note, this is also not your error, but since you're already changing it could you please adjust the copyright to conform to our standards? Replace `(c)` with `©` and add his (or his and your) name where it says `<your name>` and so on. Would appreciate it, but you don't have to, otherwise I'll just do it myself after we merge this.

z16:
I've been on vacation for a while, but I'll return home at the end of the weekend, then I'll look over it again. Sorry it's taking so long!

eeach:
Oh, no worries. I thought you were waiting for me.

I had a bunch of IRL work stuff going on, so I'm sorry it took me so long to get back to this.

z16:
Sorry, same here, this completely slipped my mind. I'll try to give it a rundown in a few hours.

addons/macrochanger/options.lua
```diff
@@ -0,0 +1,47 @@
+options = {}
```
z16, line 1: It's cleaner to define a local variable `options` and then return it at the end of the module. That way you don't have to modify global state inside a module and can reuse it easier. (You'd then have to include it by assigning it.)

addons/macrochanger/macrochanger.lua
```diff
@@ -26,143 +27,67 @@
...
+			return
+        elseif options[args[i]] then
+            options[args[i]](args[i+1])
+            i = i + 2
+        else
+            windower.add_to_chat(17, 'Unknown parameter: '..args[i])
+            options.help()
+            return
+        end
+    end
```
z16, line 53: Is this mechanism meant to loop over multiple options in the same line? While that wouldn't really be an issue, it prevents passing an arbitrary number of arguments to the respective functions (although it seems that this addon in particular doesn't need that). And you don't really gain much from it, since you can easily chain commands with `;`.

addons/macrochanger/README.md
```diff
@@ -0,0 +1,47 @@
...
+Enables all automated macro switching (not disabled individually).
+
+```
+//macrochanger enabled false
+```
+
+Disables all automated macro switching.
+
+### Configuration Example: ###
+```
```
z16, line 21: If you add `xml` to this line you'll get syntax coloring.

z16 (approved):
I added some comments, it looks good to me. The few things I mentioned are very much optional. Let me know if you still want to adjust something or if you want me to merge it as it is.

PR #1585, opened 2017-10-30
Records of eminence addon

cairthenn:


addons/roe/roe.lua
```diff
@@ -0,0 +1,280 @@
...
+    packets.inject(p)
+end
+
+local function eval(...)
+    assert(loadstring(table.concat({...}, ' ')))()
+end
+
+local function save(name)
+    if not type(name) == "string" then
+        error('`save` : specify a profile name')
```
z16, line 84: Markdown doesn't actually work here ;D If that was the intent behind using ``.

Since you're using `logger` it will redirect `error` to the FFXI log, so you can use FFXI coloring functions to highlight terms.

addons/roe/roe.lua
```diff
@@ -0,0 +1,280 @@
...
+        return
+    end
+    
+    name = name:lower()
+
+    settings.profiles[name] = S(_roe.active:keyset())
+    settings:save('global')
+    notice('saved %d objectives to the profile %s':format(_roe.active:length(), name))
```
z16, line 92: Sometimes you start sentences with upper case and sometimes with lower case!

addons/roe/roe.lua
```diff
@@ -0,0 +1,280 @@
...
+    end
+    
+    if type(settings.profiles[name]) == "string" then
+        settings.profiles[name] = S(settings.profiles[name]:split(','):map(tonumber))
+    end
+    
+
+    
+    local needed_quests = settings.profiles[name]:diff(_roe.active:keyset())   
+    local available_slots = 30 - _roe.active:length()
```
z16: Magic number! Is that a fixed constant? Maybe put it somewhere near the top for easy adjustment, if that can be ever adjusted.

I just saw below it was the length of the packet payload, so we could derive the number from the `fields.lua` file. That info should be available and exposed through the `packets` API.

addons/roe/roe.lua
```diff
@@ -0,0 +1,280 @@
...
+    for id,progress in pairs(_roe.active:copy()) do
+        if progress == 0 or settings.clearprogress then
+            cancel_roe(id)
+            coroutine.sleep(.5)
+        end
+    end
+end
+
+local true_strings = S{'true','t','y','yes','on'}
+local false_strings = S{'false','f','n','off'}
```
z16: `'no'` is missing here, unless that was intentional.

addons/roe/roe.lua
```diff
@@ -0,0 +1,280 @@
...
+    setting = handle_setting,
+    blacklist = blacklist,
+    help = help,
+}
+
+
+local function inc_chunk_handler(id,data)
+    if id == 0x111 then
+        _roe.active:clear()
+        for i = 1,30 do
```
z16: The same `30` as above? I assume that's packet payload length, as indicated in `fields.lua`. Any reason not to use `packet.parse` here?

addons/roe/roe.lua
```diff
@@ -0,0 +1,280 @@
...
+                _roe.active[id] = progress
+            end
+        end
+    elseif id == 0x112 then
+        local complete = T{data:unpack('b1':rep(1024),4)}:key_map(
+            function(k) 
+                return (k + 1024*data:unpack('H', 133) - 1) 
+            end):map(
+            function(v) 
+                return (v == 1 and true) or false
```
z16: This line should be equivalent to `return v == 1`, no?

Since they use this type of bit field quite a bit (quests, missions, key items, maps, etc.) we should maybe introduce a specific type for it so we don't need to do manual parsing...

z16 (approved):
👍

addons/roe/roe.lua
cairthenn: It was not!

addons/roe/roe.lua
cairthenn: The only reason was so I didn't have to put together a string with index i :( I hate strings :(

addons/roe/roe.lua
cairthenn: Whoops. And yeah, I think it would be a good idea.

addons/roe/roe.lua
cairthenn: I think I've avoided color codes forever because they have the rare chance to bug out the chat and I couldn't really pinpoint why

PR #1587, opened 2017-11-04
Fix bug in gearswap filter_pretarget function that can conflate job ability and weapon skill IDs

### LINKED ISSUE #1586: Gearswap filter_pretarget can conflate job ability and weapon skill IDs
Example:

BLU/THF not wearing a sword or unable to use Savage Blade while attempting to use Savage Blade shouldn't be able to finish the function "filter_pretarget" in helper_functions.lua without setting err to true, but since "Flee" job ability has the same id in resources, it proceeds due to this block:

    elseif category == 7 or category == 9 then
        local available = windower.ffxi.get_abilities()
        available = S(available.job_abilities):union(S(available.weapon_skills))
        if not available[action.id] then
            bool,err = false,"Unable to excute command. You do not have access to that ability."
        end
        ...

ghost:
Fixes #1586

### DIFF
--- a/addons/GearSwap/helper_functions.lua
+++ b/addons/GearSwap/helper_functions.lua
@@ -642,11 +642,11 @@ function filter_pretarget(action)
         local available_spells = windower.ffxi.get_spells()
         bool,err = check_spell(available_spells,action)
     elseif category == 7 then
-		local available = windower.ffxi.get_abilities().weapon_skills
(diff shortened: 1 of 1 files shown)


cairthenn (changes requested):
Looks good to me! Change your tabs to spaces, though, please!

ghost:
Oops! Fixed that.

PR #1588, opened 2017-11-04
Fix filter_pretarget function to have proper table checking syntax instead of always returning nil

ghost:
This fixes a regression in the #1586 PR.


GenericHero:
This fixes a regression in the #1586 PR.

PR #1607, opened 2017-12-09
Fix Config lib to handle numeric keys properly

Nifim:
config.lua will convert numeric keys to string after reading from the xml. these changes check if the value is a number.

### DIFF
--- a/addons/libs/config.lua
+++ b/addons/libs/config.lua
@@ -153,7 +153,7 @@ function merge(t, t_merge, path)
     end
     
     for lkey, val in pairs(t_merge) do
-        local key = keys[lkey:lower()]
+        local key = keys[tonumber(lkey) or lkey:lower()]
         if not key then
             if type(val) == 'table' then
                 t[lkey] = setmetatable(table.copy(val), getmetatable(val) or _meta.T)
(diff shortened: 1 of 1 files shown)


cairthenn:
@z16 Think we just wanted to know if there's any reason we can't do this/if config already had a built in option for it nobody was aware of

z16:
While me coding it that way was deliberate to avoid a potential issue, the reality is that it's an issue unlikely to ever occur, so I wouldn't mind this PR. On the other hand, I believe the original issue has now been resolved (by using either sets or lists, see our Discord discussion about it). So I'll leave it up to @Nifim to decide. Feel free to pull it if you want.

Nifim:
I dont think its necessary if we update the documentation to state that all keys from a standard table will be returned as strings, to return numeric keys use the list structures.

By default i think people expect to get what they put in. I believe we could make it little smarter and check if the numeric value that tonumber() got is = to or greater then the length of the input string. this would prevent converting a mixed string and numeric key.

Edit: Something like this?.
function key_check(k)
               local key
               if tonumber(k) and tosting(tonumber(k)) == k then
                              key = tonumber(k)
               else
                              key = k:lower()
               end
               return key
end

PR #1613, opened 2018-01-16
Giltracker addon

azamorapl:
Submitting a new addon that displays the current gil, similar to the FFXIV Gil HUD widget:
https://i.imgur.com/vZ8NkDr.png


addons/giltracker/giltracker.lua
```diff
@@ -0,0 +1,182 @@
...
+    local yRes = windower.get_windower_settings().ui_y_res
+    update_gil()
+    gil_image:pos(xRes + settings.GilText.pos.x + 1 ,
+      yRes + settings.GilText.pos.y - (settings.GilImage.size.height/6))
+    show()
+end
+
+function update_gil()
+  local gil = windower.ffxi.get_items().gil
+  gil_text:text(''..comma_value(gil))
```
z16: `tostring(...)` indicates what you wanna do more than relying on Lua's awful string conversion behavior.

However, I'd skip it entirely and adjust the `comma_value` function to convert the argument to a string directly. Currently it only works because of another one of Lua's stupid quirks: `string.gsub` also works on numbers, which is not really intuitive. So I'd do this below:
```lua
    local formatted = tostring(amount)
```

addons/giltracker/giltracker.lua
```diff
@@ -0,0 +1,182 @@
...
+end)
+
+windower.register_event('keyboard', function(dik, flags, blocked)
+  if (dik == hideKey) and (flags == true) and (is_hidden_by_key == true) and (is_hidden_by_cutscene == false) then
+    is_hidden_by_key = false
+    show()
+  elseif (dik == hideKey) and (flags == true) and (is_hidden_by_key == false) and (is_hidden_by_cutscene == false) then
+    is_hidden_by_key = true
+    hide()
+  end
```
z16: You usually use four space indentation, but occasionally switch to two spaces. Please adjust it to all four space indentation.

addons/giltracker/giltracker.lua
```diff
@@ -0,0 +1,182 @@
...
+
+
+windower.register_event('incoming text', function(...)
+    update_gil()
+end)
+
+function initialize()
+    hideKey = settings.HideKey
+    local xRes = windower.get_windower_settings().ui_x_res
+    local yRes = windower.get_windower_settings().ui_y_res
```
z16: You acquite `windower.get_windower_settings` twice here. It's not a tremendously expensive function, but I would still avoid calling it twice:
```lua
    local windower_settings = windower.get_windower_settings()
```

Then access that for x/y resolution.

addons/giltracker/giltracker.lua
```diff
@@ -0,0 +1,182 @@
...
+    update_gil()
+    gil_image:pos(xRes + settings.GilText.pos.x + 1 ,
+      yRes + settings.GilText.pos.y - (settings.GilImage.size.height/6))
+    show()
+end
+
+function update_gil()
+  local gil = windower.ffxi.get_items().gil
```
z16: `windower.ffxi.get_items()` is one of the most expensive functions in our API, and you potentially call it quite often here. What this does is go through the entire inventory, all slots in every bag, convert all the item data and push it to the Lua stack. Since this was so awful we added some overloads to the function to make it more efficient if you don't need the entire inventory (which is the case for you, since you only need gil).

So now you can do these, for example, and it will only go through the specified bags:
```lua
winodwer.ffxi.get_items('inventory')
winodwer.ffxi.get_items('safe')
winodwer.ffxi.get_items('treasure')
winodwer.ffxi.get_items('equipment')
```

You can even specify individual slots, if you know exactly what you want:
```lua
winodwer.ffxi.get_items('inventory', 12)
```

In your case, you'll want this:
```lua
winodwer.ffxi.get_items('gil')
```

addons/giltracker/giltracker.lua
```diff
@@ -0,0 +1,182 @@
...
+
+windower.register_event('logout', function(...)
+    hide()
+end)
+
+windower.register_event('add item', function(...)
+    update_gil()
+end)
+
+windower.register_event('remove item', function(original, modified, original_mode, modified_mode, blocked)
```
z16: The function signature is wrong, it looks like the one for the `incoming text` event. Not that Lua cares ¯\ (ツ) /¯

`remove item` and `add item` have the same signature: `function(bag, slot, id, count) --[[ ... ]] end`

Which is why it surprises me that the below code works the way you did it... are you sure it does. Again, because Lua stupidly doesn't error when calling string functions on numbers, this silently passes and returns `nil`. You would need to convert the item ID to an item using the resources first to compare the name:
```lua
windower.register_event('remove item', function(bag, slot, id, count)
    local item = res.items[id]
    if item.english:match('Gil') then -- it's also spelled with a capital G
        update_gil()
    end
end)
```

However, this would be true on items like `Gilt tapestry` as well, since it contains `gil`. You'd be best to either check for the whole name (item.english =='Gil'), or save yourself extra effort and just go by ID and save the entire resource lookup:
```lua
    if id == 0xFFFF then
        update_gil()
    end
```

addons/giltracker/giltracker.lua
```diff
@@ -0,0 +1,182 @@
...
+windower.register_event('login', function()
+    initialize()
+end)
+
+windower.register_event('logout', function(...)
+    hide()
+end)
+
+windower.register_event('add item', function(...)
+    update_gil()
```
z16: This updates gil every time any item is added. See comments below for `remove item`, the same applies here.

addons/giltracker/giltracker.lua
```diff
@@ -0,0 +1,182 @@
...
+    if (new_status_id == 4) and (is_hidden_by_key == false) then
+        is_hidden_by_cutscene = true
+        hide()
+    elseif (new_status_id ~= 4) and (is_hidden_by_key == false) then
+        is_hidden_by_cutscene = false
+        show()
+    end
+end)
+
+windower.register_event('keyboard', function(dik, flags, blocked)
```
z16: `flags` is an odd name for this. This parameter indicates only if the button was pressed down or released. So a name like `down` would be more appropriate, I guess.

z16 (changes requested):
I added some comments, most of the addon is fine, all of the comments are merely suggestions (although I'd recommend still going through them), except for the four spaces thing, as that's our policy for accepting addons.

There is one thing I'd investigate though, if I were you, namely how gil is actually updated. I assume it happens with the typical item packets. In fact, if you asked me before I read through this addon, I would have said that changing gil does not generate any `add item` or `remove item` events. If it does, great, this will do fine. Otherwise I'd look into getting the right packet and update info via that.

addons/giltracker/giltracker.lua
z16: Actually, I just realized that this only happens on load or login, so it's not too bad...

addons/giltracker/giltracker.lua
trv6: There's a fourth parameter for the `keyboard` event. I think the order is `function(dik, down, flags, blocked)`. There must have been an example or an early addon where someone forgot the `down` parameter, because I remember copy/pasting the same a few times.

It won't cause you trouble unless you check the `blocked` parameter expecting to see if another addon has handled the event. Currently `blocked` is being passed the `flags` data, which will always short-circuit to `true`.

addons/giltracker/giltracker.lua
azamorapl: I messed up here. That was supposed to go inside the 'incoming text' event: for some reason, when you receive gil when you defeat a mob the 'add item' event wasn't getting triggered, hence the need to read the 'gil' part of an incoming text to trigger the update of the gil counter.

azamorapl:
This was my first lua effort ever so the suggestions are much appreciated. I've fixed all of them, let me know if something else can be improved.

addons/giltracker/giltracker.lua
```diff
@@ -0,0 +1,212 @@
...
+
+windower.register_event('login', function()
+    initialize()
+end)
+
+windower.register_event('logout', function(...)
+    hide()
+end)
+
+windower.register_event('add item', function(...)
```
z16: The function signature needs to be adjusted:
```lua
windower.register_event(function(bag, slot, id, count)
    update_gil_if_item_id_matches(id)
end)
```

Otherwise the `id` parameter will always be `nil` and the check in the following function will always fail.

And ftr, there's something some people do in Lua, if a parameter is unused (since this function has four parameters, but you only use one of them), which is to name all other parameters `_`, to indicate that they are unused. Don't have to do that, some people also prefer not to do it to keep the signature accurate. Just letting you know.

addons/giltracker/giltracker.lua
```diff
@@ -0,0 +1,212 @@
...
+    update_gil_if_item_id_matches(id)
+end)
+
+windower.register_event('remove item', function(bag, slot, id, count)
+    update_gil_if_item_id_matches(id)
+end)
+
+windower.register_event('incoming text', function(original, ...)
```
z16: Why are you checking for this event? Are there situations where the first two events (add/remove item) don't cover all the cases? Even if so, I would update the matching function, yours updates every time the characters `g`, `i` and `l` appear next to each other. So any time someone mentions "agility", "gilded", "fragile", etc. it will update the gil count. Which isn't *bad*, just unnecessary.

z16 (commented):
Looking good, I added two more comments, but no glaring issues anymore at this time. Let me know when you want me to pull it.

addons/giltracker/giltracker.lua
z16: You're right, and they already adjusted it accordingly. The confusion is actually kind warranted, because whether the key is pressed or released is normally part of the flags, but since it was so commonly used (in comparison to the other flags) we decided to extract it into its own parameter on the LuaCore side.

addons/giltracker/giltracker.lua
azamorapl: For some reason, when you receive gil by defeating a mob the 'add item' event doesn't get triggered, hence the need to read the 'gil' part of an incoming text to trigger the update of the gil counter. I can't think of a better solution to this...

It is strange that the add item event doesn't fire, though.

EDIT: I can't seem to replicate the issue anymore, which leads me to believe it may have been something with my code (can't be the add item signature, I only messed that up during the last commit).

EDIT2: Replicated it. Seems to be random, I was killing quadavs in gustaberg when the event didn't fire.

addons/giltracker/giltracker.lua
cairthenn:
```lua
windower.register_event('add item', function(bag, slot, id, count)
    update_gil_if_item_id_matches(id)
end)
```

:D

addons/giltracker/giltracker.lua
cairthenn: Gil should *always* be updated by one of the incoming item packets, which should appropriately call the `add item` or `remove item` events. If you do ever notice a discrepancy, you can make an issue though, because that would mean there's a problem!

azamorapl: I'll make an issue, I definitely have been able to reproduce the problem.

There is another case when gil doesn't get updated by incoming packets: during login. I traced it with logs and item 0xFFFF is never added.

windower.ffxi.get_items('gil') returns 0 during login too, so, I'll handle it with the "Stop downloading" packet. I suppose that there's another packet we are not aware of that updates currencies or something?

azamorapl:
Pushed some fixes. See https://github.com/Windower/Lua/pull/1613#discussion_r162234425

addons/giltracker/giltracker.lua
z16: If you use PacketViewer you can check your current gil amount, log out, then type this into the console:
```
packetviewer scan incoming 123456
```

Then log in, it will scan all incoming packets for this value. That should tell you which packet carries that information. It's possible (and wouldn't surprise me at all) if the gil value is also transmitted separately from other items.

However, it's also possible that it just isn't transmitted at all - after all you don't need to know it until you open the menu, which sends a packet that requests some info from the server (like player stats). If that turns out to be the case, you also force that info by injecting that outgoing packet on login.

azamorapl:
I've updated the PR to include the changes requested, mainly using the packets library to get the necessary events when gil is updated instead of parsing incoming text.

By the way, I had to leave the 'add/remove item' events even though i'm already parsing the ITEM_UPDATE packet because there were some situations where it would not be received (for example while bidding on the auction house).

PR #1614, opened 2018-01-16
Invtracker addon

azamorapl:
Submitting a new addon that displays a grid detailing empty and filled inventory slots, similar to the FFXIV Inventory Grid HUD widget:
https://i.imgur.com/PgiMxRZ.png


addons/invtracker/README.md
```diff
@@ -0,0 +1,17 @@
...
+## How to edit the settings
+1. Login to your character in FFXI
+2. Edit the addon settings file: **_Windower4\addons\invtracker\data\settings.xml_**
+3. Save the file
+4. Press Insert in FFXI to access the windower console
+5. Type ``` lua r invtracker ``` to reload the addon
+6. Press Insert in FFXI again to close the windower console
+
+## Issues:
+1. You may need to log out from you character and enter again in order to have the addon display properly if loaded manually.
```
z16: Zoning should also work, I am not aware of anything that a login does that a zone doesn't do as well, at least packet wise, which is (almost) all we use with Windower. Although I don't think there's anything you should need to load for - or is there something about the item structure we don't expose via `windower.ffxi.get_items`?

addons/invtracker/invtracker.lua
```diff
@@ -0,0 +1,460 @@
...
+defaults.SlotImage.status.empty.color.alpha = 150
+defaults.SlotImage.status.empty.color.red = 0
+defaults.SlotImage.status.empty.color.green = 0
+defaults.SlotImage.status.empty.color.blue = 0
+defaults.SlotImage.status.empty.background = {}
+defaults.SlotImage.status.empty.background.color = {}
+defaults.SlotImage.status.empty.background.color.alpha = 150
+defaults.SlotImage.status.empty.background.color.red = 0
+defaults.SlotImage.status.empty.background.color.green = 0
+defaults.SlotImage.status.empty.background.color.blue = 0
```
z16: You're mixing `PascalCase` with `camelCase` and `snake_case` throughout the settings, I would suggest adopting one style (and I would go with `PascalCase` as it's obviously the best).

addons/invtracker/invtracker.lua
```diff
@@ -0,0 +1,460 @@
...
+settings.SlotImage.background.texture.fit = true
+settings.SlotImage.background.size = {}
+settings.SlotImage.background.size.height = 3
+settings.SlotImage.background.size.width = 3
+settings.SlotImage.background.repeatable = {}
+settings.SlotImage.background.repeatable.x = 1
+settings.SlotImage.background.repeatable.y = 1
+
+local yRes = windower.get_windower_settings().ui_y_res
+local xRes = windower.get_windower_settings().ui_x_res
```
z16: See my comments on `giltracker`, and again, it only happens on load, so it's not too bad.

addons/invtracker/invtracker.lua
```diff
@@ -0,0 +1,460 @@
...
+end)
+
+windower.register_event('remove item', function(...)
+    if (os.time() - update_time > 30) then
+        update()
+    end
+end)
+
+function initialize()
+    hideKey = settings.HideKey
```
z16: Is this configurable? It could be, I'm not very familiar with the mechanism. If so, I would suggest adding this to *giltracker* as well.

But in that case, I would refactor this. The config library has a reloading mechanism which can execute a function every time it reloads the settings:
```lua
config.register(function(settings)
    hideKey = settings.HideKey
end)
```

This will automatically be executed on load, login, and any other time the settings are reloaded (such as doing it manually).

addons/invtracker/README.md
```diff
@@ -0,0 +1,17 @@
...
+2. There is no way to get the inventory sort order, so all items in the grid will be ordered by status and item count.
+3. Since there is no way to know many items determine a full stack, all regular items are painted blue. Instead, bazaar equipped items are painted orange, linkshell and equipped items are painted green, and temporary items are painted red.
```
z16: You can actually know the max stack size, if you use the resources. In there items have a `stack` property, which indicates the max stack size, so you could color incomplete stacks differently based on it, for example.

addons/invtracker/invtracker.lua
```diff
@@ -0,0 +1,460 @@
...
+    update_bag(settings.SlotImage.mogLocker,items.locker,items.max_locker,items.enabled_locker)
+    update_bag(settings.SlotImage.mogSatchel,items.satchel,items.max_satchel,items.enabled_satchel)
+    update_bag(settings.SlotImage.mogSack,items.sack,items.max_sack,items.enabled_sack)
+    update_bag(settings.SlotImage.mogCase,items.case,items.max_case,items.enabled_case)
+    update_bag(settings.SlotImage.mogWardrobe,items.wardrobe,items.max_wardrobe,items.enabled_wardrobe)
+    update_bag(settings.SlotImage.mogWardrobe,items.wardrobe2,items.max_wardrobe2,items.enabled_wardrobe2)
+    update_bag(settings.SlotImage.mogWardrobe,items.wardrobe3,items.max_wardrobe3,items.enabled_wardrobe3)
+    update_bag(settings.SlotImage.mogWardrobe,items.wardrobe4,items.max_wardrobe4,items.enabled_wardrobe4)
+    update_temp_bag(settings.SlotImage.tempInventory,items.treasure,#items.treasure,items.enabled_treasure)
+    --update_temp_bag(settings.SlotImage.tempInventory,items.temporary,items.max_temporary,items.enabled_temporary)
```
z16: Forgot to rempove that? Or forgot to uncomment it? Or WIP?

addons/invtracker/invtracker.lua
```diff
@@ -0,0 +1,460 @@
...
+    update_coordinates()
+    if slot_images[current_block][current_slot] == nil then
+        slot_images[current_block][current_slot] = {}
+    end
+    print_slot_background(status.background.color)
+    print_slot_box(status.color)
+    update_indexes(max_columns,last_index)
+end
+
+function print_slot_background(slot_color, max_columns, last_index)
```
z16, line 597: Might be worth introducing a local variable for `slot_images[current_block][current_slot]` to reduce the clutter. Same below in `print_slot_box`.

addons/invtracker/invtracker.lua
```diff
@@ -0,0 +1,460 @@
...
+    update_time = os.time()
+end)
+
+windower.register_event('logout', function(...)
+    hide()
+    clear()
+end)
+
+windower.register_event('incoming chunk',function(id,org,modi,is_injected,is_blocked)
+    if (os.time() - update_time > 30 and id == 0x50 or id == 0xE0 or id == 0x1D) then
```
z16: The `update_time` variable is never updated. I assume the idea is to prevent spamming, so you only update once every x ms?

z16 (approved):
Added my review, all of it looks good. Made some suggestions, but they're just that. Let me know if you want to work on them or if you want me to pull it.

If you're on the edge about changing setting names, you should consider it now, because those are hard to change later on.

addons/invtracker/README.md
azamorapl: For some reason, the images are not loaded properly when loading the addon from within the game, if you look closely at the picture they look like small dots instead of squares: https://imgur.com/iT5iQqP

Zoning solves the issue, so i've updated the readme with your suggestion.

addons/invtracker/invtracker.lua
azamorapl: It's only intented to work when logging in, since you get the incoming chunk event several times per second and really lags the game. With this I validate that 30 seconds have elapsed before actually drawing the grid.

I've changed the name to start_time instead of update_time to reflect this.

addons/invtracker/invtracker.lua
azamorapl: My bad, this got commented out while testing and I've added it back with a small refactor.

addons/invtracker/invtracker.lua
azamorapl: I settled with camelCase because I come from the Java world :(

azamorapl:
Great review, I've taken all suggestions and implemented them. The code could be better, but I think I can settle with this unless you have other suggestions.

addons/invtracker/invtracker.lua
```diff
@@ -0,0 +1,525 @@
...
+local current_slot = 1
+local current_row = 1
+local current_column = 1
+local last_column = 1
+local items = {}
+local slot_images = {}
+local start_time = 0
+
+config.register(settings, function(settings)
+    hideKey = settings.HideKey
```
z16: You should consider adding `xBase` and `yBase` in here as well, since at the moment they will only be set on load, if a user changes settings later and relogs for example, it will be ignored.

z16 (approved):
All looking good, added just a comment about two variables which could be added to the config event registry.

Also, you can obviously keep working on it, so everything doesn't have to be done right now. And here's a suggestion for the next version:
Currently, you update the entire thing every time an item-related packet comes in (although you're not considering all item-related packets yet, and `0x01C` and the treasure-pool packets might also be worth considering). When you zone you get spammed by hundreds of item packets, and after every single one you acquire all items from the LuaCore API and construct the display anew. You can consider waiting until all the zoning-related packets are done. Currently you do that on login with the 30s timer, but there's actually a packet that comes in which indicates that the inventory is fully loaded, and I believe it was `0x01D`. You can experiment with that, and set a global variable that indicates you're ready as soon as you receive that, and a set it to not ready when the zone packet comes in `0x00A`. The latter also comes on login before any other packet, so you can use it to cover both cases.

Another improvement would be to only adjust the specific bag that was affected. For example, if you get an inventory related item only adjust the equipment and the bag you equipped that item from (that info is in the packet). If an item changed packet comes in from a certain bag, only adjust that bag.

But anyway, that's something you can consider for the next version of the addon, as they're only improvements on the current functionality. Let me know if you want to fix the `xBase`/`yBase` thing or if you want me to pull it now.

addons/invtracker/invtracker.lua
z16: You get the same packet every time you zone - see my suggestion below about packet `0x00A`. As far as FFXI is concerned, zoning is no different from logging in, from a packet standpoint. The fix here would be easy, just reset `start_zone` to `os.time()` on incoming packet `0x00A`.

addons/invtracker/README.md
z16: This is something that should be looked into, that really shouldn't happen. If you can narrow down the issue or create a small test case that doesn't work right, I can look into it.

addons/invtracker/invtracker.lua
z16: I figured, sadly many people do :(

azamorapl:
I've reworked the addon with your suggestions. I now read the appropiate packets and according to that wait for INVENTORY_FINISH to update the graphics, so no more unecessary updates are performed. I tested this over the weekend and it's working fine.

The only thing left to do would be updating only the item or bag that received the event, but for now it's working fine so i'll defer that for a later version.

PR #1615, opened 2018-01-18
Boxdestroyer: Fix zone id not updating

Ivaar:


z16:
It looks like you weren't working off of the latest version of the code base. Please do a local pull to get the latest changes, then commit a new update from that base version.

PR #1625, opened 2018-03-01
Lookup addon

Karuberu:
A simple addon that looks up search terms through in-game commands.

### DIFF
--- a/addons/Lookup/Lookup.lua
+++ b/addons/Lookup/Lookup.lua
@@ -1,5 +1,5 @@
 --[[
-    Copyright (c) 2018, Karuberu
+    Copyright © 2018, Karuberu
     All rights reserved.
 
     Redistribution and use in source and binary forms, with or without
@@ -28,8 +28,7 @@
 _addon.name = 'Lookup'
 _addon.author = 'Karuberu'
 _addon.version = '1.0'
-_addon.language = 'english'
-_addon.command = 'lookup'
+_addon.commands = {'lookup', 'lu'}
 
 config = require('config')
 res = require('resources')
@@ -38,32 +37,44 @@ settings = nil
 ids = nil
 last_item = nil
 
-function initialize()
-    load_settings()
-    initialize_ids()
-end
-
 function load_settings()
     settings = config.load({
         default = 'ffxiclopedia';
-        url = {
-            ffxiclopedia = 'http://ffxiclopedia.wikia.com/wiki/Special:Search?go=Go&search=%s';
-            bgwiki = 'https://www.bg-wiki.com/bg/Special:Search?go=Go&search=%s';
+        sites = {
+            ffxiclopedia = {
+                search = 'http://ffxiclopedia.wikia.com/wiki/Special:Search?go=Go&search=${term}';
+            };
+            ['bg-wiki'] = {
+                search = 'https://www.bg-wiki.com/bg/Special:Search?go=Go&search=${term}';
+            };
             ffxidb = {
-                item = 'http://www.ffxidb.com/items/%s';
-                zone = 'http://www.ffxidb.com/zones/%s';
-                search = 'http://www.ffxidb.com/search?q=%s';
+                item = 'http://www.ffxidb.com/items/${term}';
+                zone = 'http://www.ffxidb.com/zones/${term}';
+                search = 'http://www.ffxidb.com/search?q=${term}';
             };
             ffxiah = {
-                item = 'http://www.ffxiah.com/item/%s';
-                search = 'http://www.ffxiah.com/search/item?q=%s';
-                player = 'https://www.ffxiah.com/search/player?name=%s';
+                item = 'http://www.ffxiah.com/item/${term}';
+                search = 'http://www.ffxiah.com/search/item?q=${term}';
             };
-            google = 'https://www.google.com/search?q=%s';
+            ffxiahplayer = {
+                search = 'https://www.ffxiah.com/search/player?name=${term}';
+            };
+            google = {
+                search = 'https://www.google.com/search?q=${term}';
+            };
+            ffxi = { redirect = 'ffxiclopedia'; };
+            wikia = { redirect = 'ffxiclopedia'; };
+            bgwiki = { redirect = 'bg-wiki'; };
+            bg = { redirect = 'bg-wiki'; };
+            db = { redirect = 'ffxidb'; };
+            ah = { redirect = 'ffxiah'; };
+            ffxiahp = { redirect = 'ffxiahplayer'; };
+            ahp = { redirect = 'ffxiahplayer'; };
         };
     })
 end
 
+-- Creates a list of item and zone ids by name for quicker lookup by name
 function initialize_ids()
     ids = {
         items = {};
@@ -74,11 +85,14 @@ function initialize_ids()
         ids.items[item.name_log] = item.id
     end
     for zone in res.zones:it() do
-            ids.zones[zone.name] = zone.id
+        ids.zones[zone.name] = zone.id
     end
 end
 
-function find_id(name)
+function get_id(name)
+    if id == nil then
+        return {}
+    end
     return {
         item = ids.items[name];
         zone = ids.zones[name];
@@ -86,46 +100,60 @@ function find_id(name)
 end
 
 function get_name(id, list)
-    for key, value in next, list do
-        if value == id then
-            return key
-        end
+    if id == nil then
+        return nil
     end
-    return nil
+    return (list[id] or {}).name
 end
 
+-- Converts auto-translate strings to plain text.
+-- If the string is not an auto-translate string, the original string is returned.
 function translate(str)
     return windower.convert_auto_trans(str)
 end
 
-function get_selection(str)
+-- Checks to see if the string is a selector (enclosed by <>) and returns a replacement.
+-- If the string is not a selector, the original string is returned.
+function parse_selection(str)
     local target = str:match('<(.+)>')
-    if target == 'job' then
-        str = windower.ffxi.get_player().main_job_full
-    elseif target == 'subjob' or target == 'sj' then
-        str = windower.ffxi.get_player().sub_job_full
-    elseif target == 'zone' or target == 'area' then
-        if windower.ffxi.get_info().mog_house then
-            str = 'Mog House'
-        else
-            str = get_name(windower.ffxi.get_info().zone, ids.zones)
-        end
-    elseif target == 'lastitem' or target == 'item' then
-        str = get_name(last_item, ids.items)
-    elseif target ~= nil then
-        local mob = windower.ffxi.get_mob_by_target(str:match('<(.+)>'))
-        if mob ~= nil then
-            str = mob.name
-        else
-            str = nil
-        end
+    if target == nil then
+        return str
     end
-    return str
+
+    -- custom selection handlers
+    if target == 'job' or target == 'mjob' then
+        return windower.ffxi.get_player().main_job_full
+    elseif target == 'sjob' then
+        return windower.ffxi.get_player().sub_job_full
+    elseif target == 'zone' then
+        if windower.ffxi.get_info().mog_house then
+            return 'Mog House'
+        else
+            return get_name(windower.ffxi.get_info().zone, res.zones)
+        end
+    elseif target == 'item' then
+        return get_name(last_item, res.items)
+    end
+    -- default to windower's selection handlers
+    return (windower.ffxi.get_mob_by_target(str) or {}).name
 end
 
-function set_default(term)
-    settings.default = term
-    config.save(settings, 'all')
+function set_default_site(command_modifier, site)
+    settings.default = site
+    if command_modifier == 'player' or command_modifier == 'p' then
+        -- save only for the current character
+        settings:save()
+    else
+        -- save for all characters
+        settings:save('all')
+    end
+end
+
+function modify_site_settings(site, type, url)
+    if url == 'remove' then
+        url = nil
+    end
+    settings.sites[site][type] = url
 end
 
 function set_last_item(bag, index, id, count)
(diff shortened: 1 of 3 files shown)


addons/Lookup/Lookup.lua
```diff
@@ -0,0 +1,190 @@
+--[[
+	Copyright (c) 2018, Karuberu
```
z16: Please change this to a © sign. If you copied this from somewhere could you tell us where, so we can update it?

z16: Actually upon reading it, it seems very slightly (only superficially) modified from the text we use as a template. Now I want to know where you have it from even more.

z16:
This addon seems to serve the same purpose as [Linker](https://github.com/Windower/Lua/tree/dev/addons/linker). Does it do anything Linker doesn't? If so, I would probably suggest amending Linker instead. If you feel it's separate enough from Linker to coexist let me know, then I'll start a review.

addons/Lookup/Lookup.lua
```diff
@@ -0,0 +1,190 @@
...
+res = require('resources')
+
+settings = nil
+ids = nil
+last_item = nil
+
+function initialize()
+	load_settings()
+	initialize_ids()
+end
```
z16: Please changes tabs for four spaces throughout the file. Most editors can be configured to work this way by default.

Karuberu:
Lookup supports auto-translate text, has special selectors like `<t>`, `<zone>`, and `<lastitem>`, and allows a default search site to be set, among other things - very large usability improvements over Linker.

It was written from scratch, so it shares very little in common with Linker aside from the same general purpose. If I did update Linker, it would be almost a complete overwrite.

addons/Lookup/Lookup.lua
```diff
@@ -0,0 +1,190 @@
...
+
+function set_last_item(bag, index, id, count)
+	if bag == 0 then
+		last_item = id
+	end
+end
+
+function process_command(command, term)
+	if term == nil or term == '' then
```
z16: `term` cannot ever be `''`, so this check can be removed.

addons/Lookup/Lookup.lua
```diff
@@ -0,0 +1,190 @@
...
+				search = 'http://www.ffxidb.com/search?q=%s';
+			};
+			ffxiah = {
+				item = 'http://www.ffxiah.com/item/%s';
+				search = 'http://www.ffxiah.com/search/item?q=%s';
+				player = 'https://www.ffxiah.com/search/player?name=%s';
+			};
+			google = 'https://www.google.com/search?q=%s';
+		};
+	})
```
z16: I would adjust the URLs to use something other than `%s`. For users without a technical background that's hard to read, and they don't know what it does.

addons/Lookup/Lookup.lua
```diff
@@ -0,0 +1,190 @@
...
+	local target = str:match('<(.+)>')
+	if target == 'job' then
+		str = windower.ffxi.get_player().main_job_full
+	elseif target == 'subjob' or target == 'sj' then
+		str = windower.ffxi.get_player().sub_job_full
+	elseif target == 'zone' or target == 'area' then
+		if windower.ffxi.get_info().mog_house then
+			str = 'Mog House'
+		else
+			str = get_name(windower.ffxi.get_info().zone, ids.zones)
```
z16: The `get_name` function goes through the entire list and sees which one has this ID. Instead you can just use this:
```lua
            str = res.zones[windower.ffxi.get_info().zone].name
```

The same would apply to the items below, I would just get rid of the `get_name` function, since it effectively just does a resource lookup, only more expensive.

addons/Lookup/Lookup.lua
```diff
@@ -0,0 +1,190 @@
...
+	return windower.convert_auto_trans(str)
+end
+
+function get_selection(str)
+	local target = str:match('<(.+)>')
+	if target == 'job' then
+		str = windower.ffxi.get_player().main_job_full
+	elseif target == 'subjob' or target == 'sj' then
+		str = windower.ffxi.get_player().sub_job_full
+	elseif target == 'zone' or target == 'area' then
```
z16: There's a philosophical debate about allowing multiple search terms. I would argue that allowing multiple ways to specify the same thing breaks uniformity and results in a messier API, and may confuse end users as to the difference. Personally I would only use `zone` as it matches both FFXI's vernacular as well as our internal API.

addons/Lookup/Lookup.lua
```diff
@@ -0,0 +1,190 @@
...
+		if windower.ffxi.get_info().mog_house then
+			str = 'Mog House'
+		else
+			str = get_name(windower.ffxi.get_info().zone, ids.zones)
+		end
+	elseif target == 'lastitem' or target == 'item' then
```
z16: One could make the same argument as above here, but I would say that `item` is just an abbreviation of `lastitem`, so I don't think there's as much of an issue.

addons/Lookup/Lookup.lua
```diff
@@ -0,0 +1,190 @@
...
+		str = get_name(last_item, ids.items)
+	elseif target ~= nil then
+		local mob = windower.ffxi.get_mob_by_target(str:match('<(.+)>'))
+		if mob ~= nil then
+			str = mob.name
+		else
+			str = nil
+		end
+	end
```
z16: The entire last `elseif` case could be removed and the final part of the function written as:
```lua
    return (windower.ffxi.get_mob_by_target(str) or {}).name
```

`windower.ffxi.get_mob_by_name` also accepts enclosing `<>`, so you don't need to filter those out.

addons/Lookup/Lookup.lua
```diff
@@ -0,0 +1,190 @@
...
+	return str
+end
+
+function set_default(term)
+	settings.default = term
+	config.save(settings, 'all')
```
z16: While it would probably make sense to save for all users by default, I could imagine some people wanting different default sites for different accounts. A crafting mule for example could have a different preference from the main character.

addons/Lookup/Lookup.lua
```diff
@@ -0,0 +1,190 @@
...
+		if id.item ~= nil then
+			url = settings.url.ffxiah.item:format(id.item)
+		else
+			url = settings.url.ffxiah.search:format(term)
+		end
+	elseif command == 'ahp' or command == 'ffxiahp' or command == 'ffxiahplayer' then
+		url = settings.url.ffxiah.player:format(term)
+	elseif command == 'google' then
+		url = settings.url.google:format(term)
+	end
```
z16: You do a hard-coded URL lookup. I would design it in such a way that lookup goes automatically by what is specified in the settings file. That way users can add their own sites as well. The way it is now I don't see why it needs to be in user-facing settings at all, since the structure is hard-coded, and it makes little sense to adjust the URLs.

This would remove some more ambiguity as for what to specify as a command, the same thing I mentioned above.

addons/Lookup/Lookup.lua
```diff
@@ -0,0 +1,190 @@
...
+	(INCLUDING, BUT NOT LIMITED TO, PROCUREMENT OF SUBSTITUTE GOODS OR SERVICES;
+	LOSS OF USE, DATA, OR PROFITS; OR BUSINESS INTERRUPTION) HOWEVER CAUSED AND
+	ON ANY THEORY OF LIABILITY, WHETHER IN CONTRACT, STRICT LIABILITY, OR TORT
+	(INCLUDING NEGLIGENCE OR OTHERWISE) ARISING IN ANY WAY OUT OF THE USE OF THIS
+	SOFTWARE, EVEN IF ADVISED OF THE POSSIBILITY OF SUCH DAMAGE.
+]]
+_addon.name = 'Lookup'
+_addon.author = 'Karuberu'
+_addon.version = '1.0'
+_addon.language = 'english'
```
z16: This line is bad, and you're not at fault for writing this, because many people do (including the tool who wrote Linker). This effectively forces lookup to be made with English names throughout your addon, which would suck for Japanese players.

addons/Lookup/Lookup.lua
```diff
@@ -0,0 +1,190 @@
...
+_addon.command = 'lookup'
+
+config = require('config')
+res = require('resources')
+
+settings = nil
+ids = nil
+last_item = nil
+
+function initialize()
```
z16: I would remove this code from the function and write it at the global addon namespace. This has a number of possible advantages if you ever decide to do it multi-user based, and at least it's slightly more efficient than doing it in the load event.

addons/Lookup/README.md
```diff
@@ -0,0 +1,100 @@
...
+| `<t>` | The current target's name. |
+| `<bt>` | The current battle target'name . |
+| `<pet>` | The name of the current player's pet. |
+| `<me>` | The current player's name. |
+| `<r>` | The name of the player that last sent a tell to you. |
+| `<job>` | The current player's main job. |
+| `<subjob>`<br>`<sj>` | The current player's subjob. |
+| `<area>`<br>`<zone>` | The current area/zone. |
+| `<item>`<br>`<lastitem>` | The last item placed in the player's inventory, including items moved from other bags. |
+
```
z16: Since this uses `windower.ffxi.get_mob_by_target` it actually supports a lot more than these, although I'm not sure if it's worth it to mention... technically all immediate target identifiers that FFXI allows in `<>` can also be used here, like `<pX>`, `<aX>`, `<ht>` and so on. But again, not sure the use case is big enough to warrant mentioning here...

addons/Lookup/Lookup.lua
```diff
@@ -0,0 +1,190 @@
...
+	LOSS OF USE, DATA, OR PROFITS; OR BUSINESS INTERRUPTION) HOWEVER CAUSED AND
+	ON ANY THEORY OF LIABILITY, WHETHER IN CONTRACT, STRICT LIABILITY, OR TORT
+	(INCLUDING NEGLIGENCE OR OTHERWISE) ARISING IN ANY WAY OUT OF THE USE OF THIS
+	SOFTWARE, EVEN IF ADVISED OF THE POSSIBILITY OF SUCH DAMAGE.
+]]
+_addon.name = 'Lookup'
+_addon.author = 'Karuberu'
+_addon.version = '1.0'
+_addon.language = 'english'
+_addon.command = 'lookup'
```
z16: I would suggest a shorthand command for it as well, as this will rarely be macrod and users always have to type it out.

addons/Lookup/Lookup.lua
```diff
@@ -0,0 +1,190 @@
...
+
+function find_id(name)
+	return {
+		item = ids.items[name];
+		zone = ids.zones[name];
+	}
+end
+
+function get_name(id, list)
+	for key, value in next, list do
```
z16: That's a really strange way to iterate, did you copy that from another addon? It does work, but it's very anti-idiomatic for Lua, as it exploits `for`'s underlying mechanism and Lua's multiple return values rather than use the intended behavior with `pairs`.

z16 (changes requested):
I added some review comments, most of the addon is fine the way it is, but I would like to know why the copyright notice differs, and the tabbing and addon language should be adjusted before this can be accepted.

All other comments are just remarks on my side and won't stop me from merging this PR. Let me know if you want to work on some of those or if you want to leave them as it is.

addons/Lookup/Lookup.lua
Karuberu: I copied it from the [Licensing section of the Writing Addons wiki page](http://github.com/Windower/Lua/wiki/Writing-Addons#licensing) and opted to use "THE COPYRIGHT HOLDERS OR CONTRIBUTORS" instead of my name in the last paragraph so that any other contributors will also be covered.

I'll change the copyright symbol and can change the last paragraph if required.

addons/Lookup/Lookup.lua
Karuberu: Oh, interesting... So I should just remove that `_addon.language = 'english'` line, then?

It may be helpful to make some note about that technicality to the [Information section of the Writing Addons wiki page](https://github.com/Windower/Lua/wiki/Writing-Addons#information), which seems to recommend adding it.

addons/Lookup/Lookup.lua
Karuberu: Whoops. Sorry about that. I forgot to check what the standard was. I prefer tabs over spaces, but uniformity is more important.

z16: I feel you on this :(

addons/Lookup/Lookup.lua
z16: Thanks for pointing it out, I adjusted the notice there. @alphaONE2 How do you feel about the change to the copyright notice?

addons/Lookup/Lookup.lua
z16: Yes, please, I also removed it from that guide, thanks for pointing that out. I implemented the `_addon.language` field with some grand ideas in mind that turned out to be drunken misconceptions.

addons/Lookup/Lookup.lua
Karuberu: I believe I copied it from an example online somewhere, but it's also in the [lua documentation](https://www.lua.org/pil/7.3.html), interestingly enough. You're probably right that I should change it to use pairs, though, so that it's easier to understand.

addons/Lookup/Lookup.lua
Karuberu: As far as I know, they're only officially referred to as "areas" in FFXI, with "zone" being a community-generated term. I was going to just use "area", but since "zone" is more commonly used, I added both options.

I think I agree with you on selectors/search terms, though: FFXI itself does not have alternate selectors, only alternate commands. Maybe I'll use `<pos>` and sidestep the whole issue. Thinking about it now, I should probably change `<subjob>` to `<sjob>` also.

addons/Lookup/Lookup.lua
Karuberu: I blame my poor documentation/naming here, but the `get_selection` function is actually more of a pass-through parsing function. The last elseif is there so that it will return the original text if it doesn't have anything to work with (in this case, if the text doesn't have enclosing `<>`).

Thanks for the tip on `get_mob_by_target` - I wasn't sure since the [documentation](https://github.com/Windower/Lua/wiki/FFXI-Functions#windowerffxiget_mob_by_targettarget) is a little ambiguous.

addons/Lookup/README.md
Karuberu: Yeah, I picked a few that would be most useful, since the total list looked too cluttered. Maybe I should add a link to the rest, though.

addons/Lookup/Lookup.lua
Karuberu: Wow, yeah, I don't know why I opted to do it that way. Thanks for that catch. I think I'll keep the function, but change it to be a simple lookup inside.

addons/Lookup/Lookup.lua
Karuberu: That's a good point. Do you know if `lu` used by any other addon?

z16: I don't, I'd just change it, if it clashes we'll find out soon by user backlash ;)

z16:
I just noticed that you forgot to modify the `addons.xml` file. Please include that in the PR, otherwise this addon won't show up in the Launcher.

addons/Lookup/Lookup.lua
Karuberu: Hah. That works.

Karuberu:
I believe I've fixed most of the issues you brought up. Let me know if I need to change anything else.

Karuberu:
@z16 Is this still being reviewed or should I submit a new, separate pull request for the changed version?

z16:
No, this is fine, I just didn't get around to reviewing it again. I'll do it this week, I'll try to sneak it into a slow work day.

Karuberu:
Ok, no rush needed - thanks for the status update.

PR #1630, opened 2018-03-24
New addon: Debuffed

aukon:


addons/Debuffed/Debuffed.lua
```diff
@@ -0,0 +1,250 @@
+--[[
+Copyright (c) 2018, Auk
```
z16: Please use the proper copyright sign `©`.

addons/Debuffed/Debuffed.lua
```diff
@@ -0,0 +1,250 @@
...
+
+Redistribution and use in source and binary forms, with or without
+modification, are permitted provided that the following conditions are met:
+
+    * Redistributions of source code must retain the above copyright
+    notice, this list of conditions and the following disclaimer.
+    * Redistributions in binary form must reproduce the above copyright
+    notice, this list of conditions and the following disclaimer in the
+    documentation and/or other materials provided with the distribution.
+    * Neither the name of <addon name> nor the
```
z16: Add your addon name here.

addons/Debuffed/Debuffed.lua
```diff
@@ -0,0 +1,250 @@
...
+    names of its contributors may be used to endorse or promote products
+    derived from this software without specific prior written permission.
+
+THIS SOFTWARE IS PROVIDED BY THE COPYRIGHT HOLDERS AND CONTRIBUTORS "AS IS" AND
+ANY EXPRESS OR IMPLIED WARRANTIES, INCLUDING, BUT NOT LIMITED TO, THE IMPLIED
+WARRANTIES OF MERCHANTABILITY AND FITNESS FOR A PARTICULAR PURPOSE ARE
+DISCLAIMED. IN NO EVENT SHALL <your name> BE LIABLE FOR ANY
```
z16: And your name again here.

addons/Debuffed/debuffs.lua
```diff
@@ -0,0 +1,54 @@
+    return {
```
z16: You know, I think this would be a useful addition to Windower resources. Feel free to add this to the [ResourceExtractor](https://github.com/Windower/ResourceExtractor/) repository (in the [fixes.xml](https://github.com/Windower/ResourceExtractor/blob/master/fixes.xml) file). That way you could access that then.

I would also leave out the name field (both if you go the ResourceExtractor route or if you leave it here). You can use the ID to access the name from the buff resources, and then you can use the `.name` property to localize it based on client language (so it shows the japanese name on the japanese client). You can still use the english name in a comment at the end of the line for easier maintenance (both here and in the `fixes.xml` file).

addons/Debuffed/Debuffed.lua
```diff
@@ -0,0 +1,250 @@
...
+                        current_string = current_string..': '..string.format('%.0f', math.max(0, spell[2] - os.clock()))
+                    else
+                        current_string = current_string..'\n'..name
+                    end
+                    count = count + 1
+                end
+            end
+        end
+    end
+    box.current_string = count > 0 and current_string or ''
```
z16: This is minor suggestion, but repeated expanding string concatenation can be "very" expensive (for certain definitions of "very"). I would probably use a list to append lines to and later join them all by `\n`.

Has the even minorer additional benefit, that you won't need the count variable - at the end of the function you could just check if the table is empty. If so - return. If not, update the `box.current_string` with `Debuffed [' .. target.name ..']\n' .. list:concat('\n')`, where `lines` is the table used to hold all the lines.

That makes it a tad more efficient, but I doubt it would make much of a difference unless it's on the lowest of low boxes.

addons/Debuffed/Debuffed.lua
```diff
@@ -0,0 +1,250 @@
...
+end
+
+function handle_overwrites(target, new, t)
+    if debuffed_mobs[target] then
```
z16: Can invert the condition to reduce nesting:
```lua
    if not debuffed_mobs[target] then
        return true
    end
```

addons/Debuffed/Debuffed.lua
```diff
@@ -0,0 +1,250 @@
...
+    if not handle_overwrites(target, spell, overwrites) then
+        return
+    end
+    
+    -- Create timer
+    debuffed_mobs[target][effect] = {spell, os.clock() + debuffs[spell].duration}
+end
+
+function inc_action(act)
+    if act.category == 4 then
```
z16: Same as above, can invert the condition:
```lua
    if act.category ~= 4 then
        return
    end
```

addons/Debuffed/Debuffed.lua
```diff
@@ -0,0 +1,250 @@
...
+            end
+            settings:save()
+        end
+    else
+        print('%s (v%s)':format(_addon.name, _addon.version))
+        print('    \\cs(255,255,255)mode\\cr - Switches between blacklist and whitelist mode (default: blacklist)')
+        print('    \\cs(255,255,255)interval <value>\\cr - Allows you to change the refresh interval (default: 0.1)')
+        print('    \\cs(255,255,255)blacklist|whitelist add|remove <name>\\cr - Adds or removes the spell <name> to the specified list')
+    end
+end)
```
z16: No new line at the end of the file. This is critical.

z16 (changes requested):
Overall it looks pretty good, you just need to fix the copyright notice. Aside from that all the things I commented on are minor suggestions, although I would like to see the debuff file committed to the RE repository. But even if you decide to do that, it wouldn't have to be now, can do it at any other point in time.

So once you fix the copyright notice let me know if you want me to merge it as is, or if you want to implement any of the other suggestions.

aukon:
Implemented all of those suggestions. I'll definitely look into the resources but I still need to add a few other spells like songs yet.

aukon:
Switched to using resources now and I've made a pull request with an updated fixes.xml. https://github.com/Windower/ResourceExtractor/pull/52/commits/39b324adcef04f5acd7c5e928d32eee2b77afaaf

z16:
Just so you know, I want to refactor RE first to handle tables within the fixes file and interpret it as an array. As it is now we end up with the string `'{123,456,789}'` instead of an actual table in Lua. After that you'll need to adjust your code as well at the appropriate place. Since I'm currently on vacation this is taking a bit, sorry for the delay...

aukon:
No worries, I did take a look at RE myself but it's a bit out of my expertise. Should be a lot cleaner than the hacky workaround I used for the string issue.

z16:
Sorry this took so long, we went over a few things and resources were adjusted again. Now the overwrites should be an actual table instead of a string containing a table definition.

Also adjusted some duration values, but not many. Some seem to be off still, but it should be usable. Let me know when you've upgraded the addon to handle the new resources.

aukon:
Fixed for the new resources now but I also added a check for Light Shot to upgrade Dia. I've made two more pull requests to add the spell attributes for Dia IV so overwrites work properly for it.

https://github.com/Windower/ResourceExtractor/pull/54
https://github.com/Windower/Resources/pull/28

I'll look into fixing the durations if I ever get chance, haven't been actively playing recently.

PR #1633, opened 2018-04-11
New addon: Spellbook

sigilbaram:
Search for missing spells by job/level/jp, or by category. Category search is also limited to spells for which the player has a job that is high enough level to learn the spell, unless the player specifically asks for `all` spells in the category. There is also an `all` category so that the player can quickly list all spells they are missing for all jobs.

I do not have any jobs that are level 99, so I have not tested if the addon detects the player's spent JP correctly, but in theory it should&trade;. This detection happens whenever a search by job command (`current`, `main`, `whm`, etc.) doesn't specify a level or jp cap and the player has 100+ spent JP in that job. Level detection does work fine for jobs below level 99, I'm just not sure that the jp part is working. The `sub` command does also search by job, but it uses the current sub job level by default, rather than the job's actual level, so it won't ever be detecting level 99 and hence it doesn't check the player's jp.


addons/SpellBook/SpellBook.lua
```diff
@@ -0,0 +1,382 @@
...
+      notice, this list of conditions and the following disclaimer in the
+      documentation and/or other materials provided with the distribution.
+    * Neither the name of SpellBook nor the
+      names of its contributors may be used to endorse or promote products
+      derived from this software without specific prior written permission.
+
+THIS SOFTWARE IS PROVIDED BY THE COPYRIGHT HOLDERS AND CONTRIBUTORS "AS IS" AND
+ANY EXPRESS OR IMPLIED WARRANTIES, INCLUDING, BUT NOT LIMITED TO, THE IMPLIED
+WARRANTIES OF MERCHANTABILITY AND FITNESS FOR A PARTICULAR PURPOSE ARE
+DISCLAIMED. IN NO EVENT SHALL <your name> BE LIABLE FOR ANY
```
z16: Need to insert `<your name>` here.

addons/SpellBook/readme.md
```diff
@@ -0,0 +1,20 @@
...
+
+```
+//spbk help -- show this help text
+//spbk [current] -- Shows learnable spells based on current main and sub job and level.
+//spbk <main|sub> [<level|all>] -- Show missing spells for current main or sub job. Defaults to the job's current level.
+//spbk <job> [<level|all>] -- Show missings spells for specified job and level. Defaults to the job's level.
+//spbk <category> [all]-- Show learnable spells by category. Limited to spells which are learnable, unless all is added after the category.
+Categories: whitemagic, blackmagic, songs, ninjustu, summoning, bluemagic, geomancy, trusts, all (Trusts are not included in all)
+Note: A limit on spent job points can be specified for spells learned from Gifts by entering a level of 100-1500. Spells are never given as Gifts for less than 100 jp.
+```
```
z16: A bit of a nitpick, but some consistency in the comments would be neat. Some sentences start with lowercase, some with uppercase. Some end with a period, some don't. Some start with a singular `shows`, others with a declarative form `show`. Might wanna tidy that up :)

Personally I'd probably make them singular code blocks for each command, with a description below, to match a more common format among addons, but that's just a preference,

addons/SpellBook/SpellBook.lua
```diff
@@ -0,0 +1,382 @@
+_addon.name     = 'SpellBook'
+_addon.author   = 'SigilBaram'
+_addon.version  = '0.0.1'
+_addon.commands  = {'spellbook','spbk'}
```
z16: `spbk` feels a bit long for a "short" command. Maybe just `sb`? Would conflict with *Scoreboard*, but from what I hear it's not used much anymore due to *Parse*.

addons/SpellBook/SpellBook.lua
```diff
@@ -0,0 +1,382 @@
...
+    whitemagic= { type = 'WhiteMagic', readable = 'White Magic spells' },
+    blackmagic= { type = 'BlackMagic', readable = 'Black Magic spells' },
+    songs=      { type = 'BardSong', readable = 'Bard songs' },
+    ninjutsu=   { type = 'Ninjutsu', readable = 'Ninjutsu' },
+    summoning=  { type = 'SummonerPact', readable = 'Summoning spells' },
+    bluemagic=  { type = 'BlueMagic',  readable = 'Blue Magic spells' },
+    geomancy=   { type = 'Geomancy', readable = 'Geomancy spells' },
+    trusts=     { type = 'Trust', readable = 'Trusts'},
+    all=        { type = 'all', readable = 'spells of all types'}
+}
```
z16: Drives me nuts that you aligned the start of the table but not the individual entries and that you didn't insert a space before the first `=` :D

addons/SpellBook/SpellBook.lua
```diff
@@ -0,0 +1,382 @@
...
+        if args[1] == 'all' then
+            spells_by_type(spell_types[command], false)
+        elseif args[1] == nil then
+            spells_by_type(spell_types[command], true)
+        else
+            invalid_input()
+        end
+    elseif jobs[command] then
+        local job = jobs[command]
+        local level = args[1] or player.jobs[res.jobs[job].ens]
```
z16: Use `english_short` over `ens`.

addons/SpellBook/SpellBook.lua
```diff
@@ -0,0 +1,382 @@
...
+end)
+
+--[[
+Builds a list of jobs with short name as the key and id as the value, for
+reading in user input.
+--]]
+function build_job_list()
+    local jobs = {}
+    for id,val in pairs(res.jobs) do
+        jobs[val.ens:lower()] = id
```
z16: Same as above `s/ens/english_short/`.

addons/SpellBook/SpellBook.lua
```diff
@@ -0,0 +1,382 @@
...
+
+        if spell_level and spell.type ~= 'Trust' and
+            not player_spells[spell_id] and not spell.unlearnable then
+
+            missing_spells:append(format_spell(spell))
+            spell_count = spell_count + 1
+        end
+    end
+
+    if next(missing_spells) ~= nil then
```
z16: See above.

addons/SpellBook/SpellBook.lua
```diff
@@ -0,0 +1,382 @@
...
+            spell.type ~= 'Trust' and not player_spells[spell_id] and
+            not spell.unlearnable then
+
+            missing_spells[spell_level] = missing_spells[spell_level] or T{}
+            missing_spells[spell_level]:append(spell.en)
+            spell_count = spell_count + 1
+        end
+    end
+
+    if next(missing_spells) ~= nil then
```
z16: See above.

addons/SpellBook/SpellBook.lua
```diff
@@ -0,0 +1,382 @@
...
+
+    if learnable_only then
+        windower.add_to_chat(7, string.format('Showing learnable %s.',
+            spell_type.readable))
+    else
+        windower.add_to_chat(7, string.format('Showing all missing %s.',
+            spell_type.readable))
+    end
+
+    if next(missing_spells) ~= nil then
```
z16: You can use `if missing_spells:empty() then`, it evaluates to the same thing but looks more intuitive. It only works on `S{}`, `T{}` and `L{}`, not on regular Lua tables (`{}`). For those you can use `if table.empty({}) then`. You could use this above instead of `next(spell.levels) ~= nil`.

addons/SpellBook/SpellBook.lua
```diff
@@ -0,0 +1,382 @@
...
+            else
+                levels:append(res.jobs[job_id].ens .. ' Jp.' ..
+                    tostring(spell.levels[job_id]))
+            end
+        end
+        format = levels:concat(' / ')
+    else
+        format = ' ( Trust )'
+    end
+    return string.format('%-20s %s', spell.en, format)
```
z16: Use `spell.english` over `spell.en`.

addons/SpellBook/SpellBook.lua
```diff
@@ -0,0 +1,382 @@
...
+end
+
+--[[
+Returns true if the player has any jobs which is high enough level to learn
+the given spell.
+--]]
+function is_learnable(spell)
+    local player_levels = windower.ffxi.get_player().jobs
+    for job,level in pairs(spell.levels) do
+        if player_levels[res.jobs[job].ens] >= level then
```
z16: Same as above `s/ens/english_short/`.

addons/SpellBook/SpellBook.lua
```diff
@@ -0,0 +1,382 @@
...
+    if spell.type ~= 'Trust' then
+        local jobs = T{}
+        local levels = T{}
+        for job_id,_ in pairs(spell.levels) do
+            jobs:append(job_id)
+        end
+        jobs:sort()
+        for _,job_id in ipairs(jobs) do
+            if spell.levels[job_id] <= 99 then
+                levels:append(res.jobs[job_id].ens .. ' Lv.' ..
```
z16: Same as above `s/ens/english_short/`.

addons/SpellBook/SpellBook.lua
```diff
@@ -0,0 +1,382 @@
...
+                    tostring(spell.levels[job_id]))
+            else
+                levels:append(res.jobs[job_id].ens .. ' Jp.' ..
```
z16: Same as above `s/ens/english_short/`.

addons/SpellBook/SpellBook.lua
```diff
@@ -0,0 +1,382 @@
...
+
+--[[
+Show missing spells of a given type. If learnable is true, then the
+results will be limited to spells for which the player has a job at a
+level required to learn the spell.
+--]]
+function spells_by_type(spell_type, learnable_only)
+    local missing_spells = T{}
+    local player_spells = windower.ffxi.get_spells()
+    local spell_count = 0
```
z16: You should not need this count, you can just use `#missing_spells`.

addons/SpellBook/SpellBook.lua
```diff
@@ -0,0 +1,382 @@
...
+        end
+    end
+end
+
+-- Show missing spells for the current main and sub jobs.
+function spells_by_current()
+    local missing_spells = T{}
+    local player = windower.ffxi.get_player()
+    local player_spells = windower.ffxi.get_spells()
+    local spell_count = 0
```
z16: See above.

addons/SpellBook/SpellBook.lua
```diff
@@ -0,0 +1,382 @@
...
+    summoning=  { type = 'SummonerPact', readable = 'Summoning spells' },
+    bluemagic=  { type = 'BlueMagic',  readable = 'Blue Magic spells' },
+    geomancy=   { type = 'Geomancy', readable = 'Geomancy spells' },
+    trusts=     { type = 'Trust', readable = 'Trusts'},
+    all=        { type = 'all', readable = 'spells of all types'}
+}
+
+windower.register_event('addon command', function (command, ...)
+    local args = L{...}
+    local player = windower.ffxi.get_player()
```
z16: It's not commonly known, but this function (`windower.ffxi.get_player`) is pretty expensive. You use it once here and then again several times later, including in every `is_learnable` call, which is a lot. So much that it could lag out older PCs all on its own. Two ways around that, either pass that variable you get here to the nested function calls, or create a global variable for the player and update it on every login/logout.

z16 (changes requested):
Sorry it took a while, I left a few notes on changes I'd like to see, as well as some opinions which you may or may not share. The only things I would require you to change are the missing replacement in the copyright notice and the resource usage (use proper resource accessors). The rest I'll leave up to you.

But there's one more thing you need to do for your addon to even show up in the Windower launcher: you need to update the `addons.xml` file in the root addon directory. Without that the code will be in our repository but our launcher won't offer the addon to users.

addons/SpellBook/SpellBook.lua
```diff
@@ -0,0 +1,444 @@
...
+                player.sub_job, player.sub_job_level))
+        else
+            windower.add_to_chat(7,  string.format('Congratulations! You know all learnable spells for %s%d!',
+                player.main_job, player.main_job_level))
+        end
+    end
+end
+
+--[[
+Copyright © 2018, John S Hobart
```
z16, line 419: I assume that is you? Just checking, not that you forgot to change the name, it happens...

addons/SpellBook/SpellBook.lua
```diff
@@ -0,0 +1,444 @@
+_addon.name     = 'SpellBook'
+_addon.author   = 'SigilBaram'
+_addon.version  = '0.0.1'
```
z16: Is it really that pre-alpha? :D

z16 (approved):
Sorry for the long delay, I have to admit I forgot about it since I went on vacation and didn't think about it anymore when I came back <_< But this looks pretty good to me, if you're ready now I would merge, just drop me a quick line.

addons/SpellBook/SpellBook.lua
sigilbaram: Yeah that is me. heh

addons/SpellBook/SpellBook.lua
sigilbaram: Yeah... Made it version 1.0.0 since it's ready for release or whatnot. xD

PR #1634, opened 2018-04-13
New addon : checkparam

from20020516:
- Displays the total of property of target player's current equipments.

img_20180530_041408

# checkparam

- `/check` OR `/c` (in-game command)
  - Whenever you `/check` any player, displays the total of property of that players current equipments.(defined in `settings.xml`)
  - プレイヤーを「調べる」したとき、そのプレイヤーが装備しているアイテムの任意のプロパティを合計して表示します。(`settings.xml`で定義)
- `//checkparam` OR `//cp` (addon command)
  - same as /check <me\>. you can use this command in equipment menu.
  - /check <me\> と同様ですが、装備変更画面でも使用できます。

## data/settings.xml (auto-generated)

- Define the properties you want to be displayed for each job.
  - `|` divide each property
  -  `pet: ` define properties for all pets, which means avatar: wyvern: automaton: luopan:
- 表示させたいプロパティをジョブごとに定義します。
  - `|` 区切り記号
  - `pet: ` 召喚獣: 飛竜: オートマトン: 羅盤: は代わりに`pet:`で指定します。

- If there’s something wrong,or something strange,  
please tell me on Twitter [@from20020516](https://twitter.com/from20020516) **with simple English**. Thank you!


addons/checkparam/checkparam.lua
```diff
@@ -0,0 +1,174 @@
+_addon.name = 'checkparam'
+_addon.author = 'from20020516'
+_addon.version = '1.0'
+_addon.commands = {'cp','checkparam'}
+res,extdata,config = require('resources'),require('extdata'),require('config')
```
z16: Feel free to use some extra space, don't need to cram everything into one line, makes it easier to read.

addons/checkparam/checkparam.lua
```diff
@@ -0,0 +1,174 @@
...
+settings = config.load('data/settings.xml')
```
z16: You don't need to specify `'data/settings.xml'`, that is the default path. It's the same as just this:
```lua
settings = config.load()
```

I *would* however specify default values. Currently you provide a `settings.xml`. I would *remove* that file completely, and move all of its entries in here:

```lua
defaults = {
    war = 'store tp|double attack|triple attack|quadruple attack||weapon skill damage',
    mnk = 'store tp|double attack|triple attack|quadruple attack||martial arts|subtle blow',
    -- etc.
}
```

That way the `settings.xml` file will be auto-generated. Also has the advantage that if you update any defaults they will also be updated in the file, if it was missing before. That will *not* happen if you use a pre-defined file, as it will only be downloaded once.

addons/checkparam/checkparam.lua
```diff
@@ -0,0 +1,174 @@
...
+['Stone Gorget']='stoneskin+30',
+['Suppanomimi']='dual wield+5',
+['Swith Cape +1']='fast cast+4',
+['Swith Cape']='fast cast+3',
+['Veela Cape']='fast cast+1',
+['Witful Belt']='fast cast+3',
+['Yagrush']='divine benison+3',
+}
+windower.register_event('addon command', checkparam)
+windower.send_command('cp')
```
z16: Just call:
```lua
checkparam()
```

addons/checkparam/checkparam.lua
```diff
@@ -0,0 +1,174 @@
...
+        tbl[key] = tonumber(value) + (tbl[key] or 0)
+        if debug then
+            local color = windower.regex.match(key,'enhanced') and 170
+            windower.add_to_chat(color or 72,'['..id..']'..res.items[id].en..'['..key..'] '..value..' '..tbl[key])
+        end
+    end
+end
+function checkparam(arg1)
+    debug = arg1=='debug'
+    tbl,txt={},
```
z16: `debug`, `tbl` and `txt` are all globally used but you never declare them. Which is not a big issue, but it's more understandable for people reading code if you do. So I would either add these to the top:

```lua
debug = false
tbl = {}
txt = ''
```

... or I would jut make them local and pass them to the called functions (which seems to make a more sense in this case).

addons/checkparam/checkparam.lua
```diff
@@ -0,0 +1,174 @@
...
+    windower.send_command('input /checkparam <me>')
```
z16: What does this line do? Won't this loop endlessly?

addons/checkparam/checkparam.lua
```diff
@@ -0,0 +1,174 @@
...
+    for slot,bag in pairs(windower.ffxi.get_items().equipment) do
+        gtbl = windower.ffxi.get_items().equipment
```
z16: In contrast you can make this one `local`.

addons/checkparam/checkparam.lua
```diff
@@ -0,0 +1,174 @@
...
+        if not string.find(slot,'_bag') then
+            gear = windower.ffxi.get_items(gtbl[slot..'_bag'],gtbl[slot])
```
z16: Another `local`.

addons/checkparam/checkparam.lua
```diff
@@ -0,0 +1,174 @@
...
+            if gtbl[slot] > 0 and gear.id < 65535 then
+                if res.item_descriptions[gear.id] then
+                    local txt = windower.regex.split(res.item_descriptions[gear.id].en,"(Pet|Avatar|Automaton|Wyvern|Luopan): ")
```
z16: This `txt` hides the previous `txt`, and it's not even a string. I would strongly suggest naming your variables appropriately, like call this one `gear_stats` or something to describe the information it actually holds.

z16: Also please don't use `.en`, use `.english` instead, all throughout this file.

addons/checkparam/checkparam.lua
```diff
@@ -0,0 +1,174 @@
...
+                        split_text(txt,gear.id,'pet: ')
+                    end
+                end
+                if gear.extdata and extdata.decode(gear).type == 'Augmented Equipment' then
+                    for slot,augment in pairs(extdata.decode(gear).augments) do
+                        split_text(augment,gear.id)
+                    end
+                end
+                if enhanced[res.items[gear.id].en] then
+                    local txt = string.split(string.gsub(enhanced[res.items[gear.id].en], "([+-:][0-9]+)",",%1"),",")
```
z16: You should make more use of local variables to hold values, you perform a lot of work twice. For example `extdata.decode(gear)` above can be stored in a variable, same with `local name = res.items[gear.id].english`. Then this line could be rewritten to a much more readable:

```lua
                    local stats = enhanced[name]:gsub('([+-:][0-9]+)', ',%1'):split(',')
```

That said, I'm not sure why you index by `.english`, why not by `.id`?

addons/checkparam/checkparam.lua
```diff
@@ -0,0 +1,174 @@
...
+    end
+    windower.add_to_chat(69, '//checkparam')
+    for index,key in pairs(key_define()) do
+        txt = key == '' and txt..'\n' or (txt or '')..'['..key..'] '..(tbl[string.lower(key)] or 0)..' '
+    end
+    windower.add_to_chat(70, txt)
+    collectgarbage()
+end
+integrate={
+--[[integrate same property.information needed for development. @from20020516]]
```
z16: I would also indent these to show their nested level.

addons/checkparam/checkparam.lua
```diff
@@ -0,0 +1,174 @@
...
+                end
+            end
+        end
+    end
+    windower.add_to_chat(69, '//checkparam')
+    for index,key in pairs(key_define()) do
+        txt = key == '' and txt..'\n' or (txt or '')..'['..key..'] '..(tbl[string.lower(key)] or 0)..' '
+    end
+    windower.add_to_chat(70, txt)
+    collectgarbage()
```
z16: Any particular reason why you manually collect garbage here?

addons/checkparam/data/settings.xml
```diff
@@ -0,0 +1,26 @@
+<?xml version="1.1" ?>
```
z16: Should delete this file, see my previous comment on `config.load`

z16 (changes requested):
I added some comments on the code you provided, most of it is just a suggestion. The only two things that you must change are:
1. Change `.en` to `.english`
2. Add an entry in the `addons.xml` file for your addon

Also, please undo the second commit. If that is a new addon please commit it separately to this one into a new branch, and open a new pull request from there.

addons/checkparam/checkparam.lua
from20020516: uses in-game /checkparam <me> command for physical accuracy. not checkparam() :)

addons/checkparam/checkparam.lua
from20020516: thanks for reading my weird code. XD
I currently working on correcting them.

addons/checkparam/checkparam.lua
from20020516: wow, nice knowledge!
fix it~

addons/checkparam/checkparam.lua
from20020516: and thanks too reading my weired english :)

addons/checkparam/checkparam.lua
from20020516: for easier to read and legendary weapons. (e.g. Yagrush has 11 type of .id)
but fixed.

addons/checkparam/checkparam.lua
from20020516: done.

addons/checkparam/checkparam.lua
from20020516: done.

addons/checkparam/data/settings.xml
from20020516: done.

addons/checkparam/checkparam.lua
from20020516: because the 'lua memory' after checkparam() was small. is that meaningless?

addons/checkparam/checkparam.lua
from20020516: done.

addons/checkparam/checkparam.lua
from20020516: done.

addons/checkparam/checkparam.lua
from20020516: replaced to `stats `. sorry ><

addons/checkparam/checkparam.lua
from20020516: done.

addons/checkparam/checkparam.lua
from20020516: delete send_command.

addons/checkparam/checkparam.lua
```diff
@@ -0,0 +1,305 @@
...
+  SMN = 'blood pact delay|blood pact delay ii|blood pact damage|avatar perpetuation cost|pet: magic attack bonus|pet: str|pet: double attack',
+  BLU = 'store tp|double attack|triple attack|quadruple attack|critical hit rate|critical hit damage|weapon skill damage|fast cast|magic attack bonus|magic accuracy|cure potency',
+  COR = 'store tp|snapshot|rapid shot|fast cast|cure potency|magic accuracy|magic attack bonus|magic damage|weapon skill damage',
+  PUP = 'pet: hp|pet: damage taken|pet: regen|martial arts|store tp|double attack|triple attack|quadruple attack',
+  DNC = 'store tp|double attack|triple attack|quadruple attack',
+  SCH = 'magic attack bonus|magic burst damage|magic burst damage ii|magic accuracy|magic damage|fast cast|elemental magic casting time|cure potency|enh mag eff dur|enhancing magic effect duration',
+  GEO = 'pet: regen|pet: damage taken|indicolure effect duration|indi eff dur|fast cast|magic evasion',
+  RUN = 'enmity|damage taken|physical damage taken|magic damage taken|spell inturruption rate|pharanx|inquartata|fastcast',
+  levelfilter = 99,
+  debugmode = false,
```
z16: Do you intend for everyone to be able to debug? If not, I would remove this from settings and just keep it as an internal variable.

addons/checkparam/checkparam.lua
```diff
@@ -0,0 +1,305 @@
...
+tbl = {}
+
+windower.register_event('addon command',function(arg)
+  if arg == 'debugmode' then
+    settings.debugmode = not settings.debugmode
+    settings:save('all')
+  else
+    for i=0,#res.slots do
+      local slot = windower.regex.replace(string.lower(res.slots[i].en),' ','_')
+      local gear_set = windower.ffxi.get_items().equipment
```
z16: `windower.ffxi.get_items` is a *very* slow function and you call it 32 times here. You should define `local items = windower.ffxi.get_items()` once outside of the loop, then use that variable inside the loop.

addons/checkparam/checkparam.lua
```diff
@@ -0,0 +1,305 @@
...
+      local count = p['Count']
+      if count == 1 then
+        get_text(p['Item'],p['ExtData'])
+      else
+        for i=1,count do
+          get_text(p['Item '..i],p['ExtData '..i])
+        end
+      end
+    elseif p['Type'] == 1 then
+      local t = windower.ffxi.get_mob_by_id(p['Target ID'])
```
z16: `windower.ffxi.get_mob_by_index(p['Target Index'])` is faster than `get_mob_by_id`.

addons/checkparam/checkparam.lua
```diff
@@ -0,0 +1,305 @@
...
+
+tbl = {}
+
+windower.register_event('addon command',function(arg)
+  if arg == 'debugmode' then
+    settings.debugmode = not settings.debugmode
+    settings:save('all')
+  else
+    for i=0,#res.slots do
+      local slot = windower.regex.replace(string.lower(res.slots[i].en),' ','_')
```
z16: Throughout the file you use `en` and `ens`, try to use `english` and `english_short`.

z16 (changes requested):
Sorry for the slow response, finally had some time to review this. Note the changes I mentioned to resource names, i.e. use `english` over `en` and `english_short` over `ens`. The spacing is also something you should adjust to 4 spaces instead of 2. Aside from that this looks ready to merge.

PR #1638, opened 2018-04-20
Use prepared statements in get_notes

cameronhimself:
This is not just a better practice than concatenating strings for queries, but it also fixes a runtime error caused by the shop door in Upper Jeuno called `Door:"Durable Shields"`. There may be others targets with quotes as well.

All of the queries should be replaced with prepared statements, but I'm starting small.


z16:
Yes Sir, this is indeed better practice! ありがとうございます

PR #1642, opened 2018-04-25
New addon : Bonanza

from20020516:
# Bonanza

- Judge your Bonanza Marbles. activate with when you recive system message (/smes) with winning numbers.
- Purchase Bonanza Marbles with any combination of numbers.

#### Japanese

- モグボナンザの当せんを判定。当せん番号を含むシステムメッセージ (/smes) を受信するとボナンザマーブルを鑑定し、結果を出力します。
- 任意の数字の組み合わせでボナンザマーブルを購入します。

## Commands

- //bonanza judge
  - same as in-game /smes.

### Purchase marble (inject packets)

- //bonanza `<number>` `[number]`...
  - purchase specified marble(s).
- //bonanza random
  - purchase up to 10 marbles with at random.
- //bonanza sequence `<number>`
  - purchase up to 10 marbles with consecutive number.
  - e.g. `15250` then buying 15250 to 15259.
- //bonanza last `<number>`
  - purchase up to 10 random marbles with tail of specified `0-9`.
  - e.g. `0` then 2224**0**, 6231**0**, 4586**0**, 9078**0**...


addons/Bonanza/Bonanza.lua
```diff
@@ -0,0 +1,195 @@
...
+windower.add_to_chat(200,windower.to_shift_jis('2等 下4けた「8940」 当せん個数「71」個'))
+windower.add_to_chat(200,windower.to_shift_jis('1等 全5けた「66621」 当せん個数「7」個'))
+]]
+--[[
+windower.add_to_chat(200,'Rank 5 prize: "3" (last digit)-- 83,350 winners.')
+windower.add_to_chat(200,'Rank 4 prize: "08" (last two digits)-- 8,287 winners.')
+windower.add_to_chat(200,'Rank 3 prize: "723" (last three digits)-- 914 winners.')
+windower.add_to_chat(200,'Rank 2 prize: "8940" (last four digits)-- 71 winners.')
+windower.add_to_chat(200,'Rank 1 prize: "66621" (all five digits)-- 7 winners.')
+]]
```
z16: Do you still need this? If not, just remove it.

addons/Bonanza/Bonanza.lua
```diff
@@ -0,0 +1,195 @@
...
+
+Redistribution and use in source and binary forms, with or without
+modification, are permitted provided that the following conditions are met:
+
+    * Redistributions of source code must retain the above copyright
+      notice, this list of conditions and the following disclaimer.
+    * Redistributions in binary form must reproduce the above copyright
+      notice, this list of conditions and the following disclaimer in the
+      documentation and/or other materials provided with the distribution.
+    * Neither the name of checkparam nor the
```
z16: Make sure to replace `checkparam` with `Bonanza`.

addons/Bonanza/Bonanza.lua
```diff
@@ -0,0 +1,195 @@
...
+function talk_moogle()
+  --print(unpack(decides))
+  if #decides > 0 then
+    hide_ui = true
+    local packet = packets.new('outgoing',0x01A,{
+      ["Target"]=moogle.id,
+      ["Target Index"]=moogle.index,
+      ["Category"]=0,
+      ["Param"]=0,
+      ["_unknown1"]=0})
```
z16: You don't need to define `= 0` as that is the default value.

addons/Bonanza/Bonanza.lua
```diff
@@ -0,0 +1,195 @@
...
+    packets.inject(packet)
+  end
+end
+
+windower.register_event('incoming chunk',function(id,data)
+  --got a marble.
+  if id == 0x020 and hide_ui and data:unpack('H',0x0D) == marble.id and data:unpack('C',0x0F) == 0 then
```
z16: You can use the `packets` library here as well to not manually unpack the data:
```lua
    local packet = packets.parse('incoming', data)
    if packet.id == 0x020 and hide_ui and packet['Item'] == marble.id and packet['Bag'] == 0 then
        -- And so on
```

That said, your numbers look off to me. `0x0D` is the high order byte of the item ID, if I'm reading this right. That should be `0x0C`. And `0x0F` is the inventory index... you probably meant `0x0E` here? I would double check these numbers if I were you.

addons/Bonanza/Bonanza.lua
```diff
@@ -0,0 +1,195 @@
...
+require('logger')
+
+marble = {id=2559,price=2000,limit=10} --general settings
+
+win_numbers = {}
+math.randomseed(os.time())
+
+--search your bags for bonanza marble and get number then add prefix 0
+function search_marbles()
+  local marbles = {}
```
z16: Use 4 space indentation please.

z16 (changes requested):
This looks good, please adjust the copyright notice then this will be ready to merge. Let me know when you're ready and I'll merge it.

addons/Bonanza/Bonanza.lua
```diff
@@ -0,0 +1,195 @@
...
+    cmd[i] = tonumber(cmd[i]) and cmd[i]*1 or cmd[i]
+  end
+
+  if cmd[1] == 'judge' then
+    windower.send_command('input /smes')
+  elseif count == 0 then
+    error('you have already '..marble.limit..' marbles.')
+  elseif cmd[1] == 'random' then
+    for i=1,count do
+      table.insert(decides,math.random(0,99999))
```
alphaONE2: You should be aware that `math.random` relies on the C `rand` function which can only generate integers between 0 and ‭32767‬ (inclusive). Lua effectively does `rand() / 32767.0 * (upper - lower) + lower`. What this means is that some values, e.g. 32 and 48831, will never be possible.

I suggest instead calling `math.random` multiple times and combining the results.
```lua
table.insert(decides,math.random(0,99) + math.random(0,99)*100 + math.random(0,9)*10000)
```

alphaONE2 (changes requested):

alphaONE2 (approved):

addons/Bonanza/Bonanza.lua
from20020516: umm, but it works. It's probably because the .lua table starts with 1. (not 0)

from20020516:
modified all issue and ready to merge.
thank you for reviewing, sir :)

addons/Bonanza/Bonanza.lua
z16: You're right, that explains it :)

z16:
Something in the addons.xml file is broken now, merging this would remove entries for MountMuzzle and Checkparam. Can you review that file please?

from20020516:
updated. is this correct?

PR #1647, opened 2018-05-03
gearswap_disabled additions

Nekseus:
added gearswap_disabled for pet_status_change and pet_change calling equip_sets

### DIFF
--- a/addons/GearSwap/gearswap.lua
+++ b/addons/GearSwap/gearswap.lua
@@ -285,15 +285,15 @@ function incoming_chunk(id,data,modified,injected,blocked)
             refresh_globals()
             next_packet_events.globals_update = data:unpack('H',3)
         end
-        if next_packet_events.pet_status_change then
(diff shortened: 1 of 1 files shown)


addons/GearSwap/gearswap.lua
```diff
@@ -285,15 +285,15 @@ function incoming_chunk(id,data,modified,injected,blocked)
             refresh_globals()
             next_packet_events.globals_update = data:unpack('H',3)
         end
-        if next_packet_events.pet_status_change then
+        if next_packet_events.pet_status_change and not gearswap_disable then
```
z16, line 288: It's called `gearswap_disable` in this line, but `gearswap_disabled` in the other two changed lines. Is that intended?

Byrth: oops, good catch

Nekseus: Oh wow, I didn't even see those two lines at all. Good eye, and sorry for my mix up.

PR #1653, opened 2018-05-09
New Addon Request: SkillUp

Iminiillusions:
SkillUp is a very easy to use skill up bot for self targeted magic types, such as enhancing magic and healing magic.


z16:
We do not take botting addons into our repository.

PR #1655, opened 2018-05-10
Updated craft addon

Snapster-amb:
Updated craft addon create_recipes script to parse new bg-wiki
Escutcheons page format.
Updated craft addon recipes list with new recipes.
Updated craft addon to handle spirit sphere recipes using
emblem cards.

### DIFF
--- a/addons/craft/create_recipes.py
+++ b/addons/craft/create_recipes.py
@@ -2,7 +2,19 @@ import urllib2
 from bs4 import BeautifulSoup
 from slpp import slpp as lua
 import os
+import platform
 
+
+hdr = {
+    'User-Agent': 'Mozilla/5.0 (X11; Linux x86_64) AppleWebKit/537.11 (KHTML, like Gecko) Chrome/23.0.1271.64 Safari/537.11',
+    'Accept': 'text/html,application/xhtml+xml,application/xml;q=0.9,*/*;q=0.8',
+    'Accept-Charset': 'ISO-8859-1,utf-8;q=0.7,*;q=0.3',
+    'Accept-Encoding': 'none',
+    'Accept-Language': 'en-US,en;q=0.8',
+    'Connection': 'keep-alive',
+}
+
+sphere = None
 
 def get_recipe(row):
     crystals = [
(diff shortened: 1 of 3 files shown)


Snapster-amb:
Note - I didn't test any of this on windows using the XI client.

addons/craft/create_recipes.py
```diff
@@ -80,13 +111,16 @@ def get_recipes_from_soup(soup, spheres=False):
...
         ]
         rows.extend(children)
     return get_recipes_from_rows(rows, spheres)
 
 def get_items_dictionary():
-    path = 'C:\\Program Files (x86)\\Windower4\\res\\items.lua'
+    if platform.system() == 'Windows':
+        path = 'C:\\Program Files (x86)\\Windower4\\res\\items.lua'
+    else:
+        path = os.path.join(os.path.expanduser("~"), 'Resources/lua/items.lua')
```
z16, line 123: Is the working directory this is exceuted in the addon directory? If so, you can work backwards from here and just go to `..\..\res\items.lua`. Then you need no distinction by OS, nor an assumption about the installation path.

z16 (approved):
Just a minor comment. If you're fine with it, I can pull it as it is.

addons/craft/create_recipes.py
Snapster-amb: There are two ways to run the script.

1 - The way you've suggested, which is more of a FFXI user way.  If there are new recipes posted on bg-wiki and people don't want to wait for an update, they can run the script themselves from the addon directory.

2 - Running the script from a forked repository.  This is more of the developer way and the way I usually do it.  On windows, I have git repository cloned in C:/Lua and just use the windower resource files.  This lets me change the addon source code without having to worry about the launcher overwriting (or not overwriting) whatever I'm working on.  On Linux I don't have the client or windower installed, I rely on ~/Lua to run the script and ~/Resources for the resources (both cloned repositories.)

Snapster-amb: The overall point being is that if I wish to be able to run the script out of a cloned Lua repository, I can't use a relative path because the resources are not included with the Lua repository.

PR #1656, opened 2018-05-22
New addon: MountMuzzle

Sjshovan:
Allows players to change or remove the default mount music.


addons/MountMuzzle/data/settings.xml
```diff
@@ -0,0 +1,6 @@
+<?xml version="1.1" ?>
```
z16: Please remove this file, it will be auto-generated. You should never check anything into the `data` sub-folder of an addon, as that is an entirely client-side managed folder.

addons/MountMuzzle/LICENSE
```diff
@@ -0,0 +1,25 @@
...
+      notice, this list of conditions and the following disclaimer in the
+      documentation and/or other materials provided with the distribution.
+    * Neither the name of MountMuzzle nor the
+      names of its contributors may be used to endorse or promote products
+      derived from this software without specific prior written permission.
+
+THIS SOFTWARE IS PROVIDED BY THE COPYRIGHT HOLDERS AND CONTRIBUTORS "AS IS" AND
+ANY EXPRESS OR IMPLIED WARRANTIES, INCLUDING, BUT NOT LIMITED TO, THE IMPLIED
+WARRANTIES OF MERCHANTABILITY AND FITNESS FOR A PARTICULAR PURPOSE ARE
+DISCLAIMED. IN NO EVENT SHALL Patrick Finnigan BE LIABLE FOR ANY
```
z16: Not sure if that is you, but the person or organization in this place should match the copyright holder at the top of the notice.

addons/MountMuzzle/README.md
```diff
@@ -0,0 +1,155 @@
...
+ 
+___
+### Usage
+
+**Manually load** the addon by using one of the following commands:
+    
+    //lua load mountmuzzle  
+    //lua l mountmuzzle
+    
+**Automatically load** this addon by adding `lua load mountmuzzle` or `lua l mountmuzzle` to the bottom of your `Windower4/scripts/init.txt` file.
```
z16: This is actually discouraged. Automatic loading should be handled either through the Windower launcher or through the `plugin_manager` addon, which offers finer grained control over character-based plugin and addon usage.

addons/MountMuzzle/README.md
```diff
@@ -0,0 +1,155 @@
...
+Reloads the Mount Muzzle addon. Below are the equivalent ways of calling the command:
+	
+    //mountmuzzle reload
+    //muzzle reload
+    //mm reload
+
+___
+### Support
+**Having Issues with this addon?**
+* Please let me know [here.](https://github.com/Ap0gee/MountMuzzle/issues/new)
```
z16: It bothers me more than it should that the period is part of the link :<

addons/MountMuzzle/constants.lua
```diff
@@ -0,0 +1,94 @@
...
+      notice, this list of conditions and the following disclaimer in the
+      documentation and/or other materials provided with the distribution.
+    * Neither the name of MountMuzzle nor the
+      names of its contributors may be used to endorse or promote products
+      derived from this software without specific prior written permission.
+
+THIS SOFTWARE IS PROVIDED BY THE COPYRIGHT HOLDERS AND CONTRIBUTORS "AS IS" AND
+ANY EXPRESS OR IMPLIED WARRANTIES, INCLUDING, BUT NOT LIMITED TO, THE IMPLIED
+WARRANTIES OF MERCHANTABILITY AND FITNESS FOR A PARTICULAR PURPOSE ARE
+DISCLAIMED. IN NO EVENT SHALL Patrick Finnigan BE LIABLE FOR ANY
```
z16: See above.

addons/MountMuzzle/helpers.lua
```diff
@@ -0,0 +1,80 @@
...
+      notice, this list of conditions and the following disclaimer in the
+      documentation and/or other materials provided with the distribution.
+    * Neither the name of MountMuzzle nor the
+      names of its contributors may be used to endorse or promote products
+      derived from this software without specific prior written permission.
+
+THIS SOFTWARE IS PROVIDED BY THE COPYRIGHT HOLDERS AND CONTRIBUTORS "AS IS" AND
+ANY EXPRESS OR IMPLIED WARRANTIES, INCLUDING, BUT NOT LIMITED TO, THE IMPLIED
+WARRANTIES OF MERCHANTABILITY AND FITNESS FOR A PARTICULAR PURPOSE ARE
+DISCLAIMED. IN NO EVENT SHALL Patrick Finnigan BE LIABLE FOR ANY
```
z16: See above.

addons/MountMuzzle/mountmuzzle.lua
```diff
@@ -0,0 +1,222 @@
...
+      notice, this list of conditions and the following disclaimer in the
+      documentation and/or other materials provided with the distribution.
+    * Neither the name of MountMuzzle nor the
+      names of its contributors may be used to endorse or promote products
+      derived from this software without specific prior written permission.
+
+THIS SOFTWARE IS PROVIDED BY THE COPYRIGHT HOLDERS AND CONTRIBUTORS "AS IS" AND
+ANY EXPRESS OR IMPLIED WARRANTIES, INCLUDING, BUT NOT LIMITED TO, THE IMPLIED
+WARRANTIES OF MERCHANTABILITY AND FITNESS FOR A PARTICULAR PURPOSE ARE
+DISCLAIMED. IN NO EVENT SHALL Patrick Finnigan BE LIABLE FOR ANY
```
z16: See above.

addons/MountMuzzle/helpers.lua
```diff
@@ -0,0 +1,80 @@
...
+ON ANY THEORY OF LIABILITY, WHETHER IN CONTRACT, STRICT LIABILITY, OR TORT
+(INCLUDING NEGLIGENCE OR OTHERWISE) ARISING IN ANY WAY OUT OF THE USE OF THIS
+SOFTWARE, EVEN IF ADVISED OF THE POSSIBILITY OF SUCH DAMAGE.
+]]
+
+require("constants")
+
+function ucFirst(str)
+    return (str:gsub("^%l", string.upper))
+end
```
z16: There's already ` str:ucfirst()` function in our `strings` library.

addons/MountMuzzle/helpers.lua
```diff
@@ -0,0 +1,80 @@
...
+    end
+    return response_template:format(response_type:color(response_color), message)
+end
+
+function displayResponse(response, color)
+    color = color or colors.info
+    windower.add_to_chat(color, response)
+    windower.console.write(response:strip_colors())
+end
+
```
z16: While it's not terribly important, nor am I not gonna merge this because of it, your modular file design can be improved. Currently you use some functions in here that you don't declare in here. Thus you rely on this file being `require`d after some others. Normally you'd `require` everything you use in this file directly in this file.

And the `constants` file would not just write global values, but return them:
```lua
return packets, player, music, colors, muzzle
```

And in this file you'd set them like this:
```lua
local packets, player, music, colors, muzzle = require('constants')`
```

Again, don't need to really bother with it, but that's what a really modular design would look like, as now you rely on the order of inclusion too much.

addons/MountMuzzle/mountmuzzle.lua
```diff
@@ -0,0 +1,222 @@
...
+    return false 
+end
+
+windower.register_event('login', requestInject)
+windower.register_event('load', requestInject)
+windower.register_event('zone change', requestInject)
+
+windower.register_event('addon command', function(command, ...)
+    if command then
+        local command = command:lower()
```
z16: You probably want to remove the `local`. Otherwise this statement does nothing. It creates a local variable `command` which *hides* the original `command` variable, and which is immediately discarded, as the inner `command` dies at the `else` keyword.

addons/MountMuzzle/mountmuzzle.lua
```diff
@@ -0,0 +1,222 @@
...
+    end
+    return muzzles[current_muzzle]
+end
+
+function requestInject()
+    needs_inject = true
+end
+
+function injectMuzzleMusic()
+    windower.packets.inject_incoming(
```
z16: You can use the packets library to do the packing and conversions for you:
```lua
    packets.inject(packets.new('incoming', packets.inbound.music_change.id, {
        ['BGM Type'] = music.types.mount,
        ['Song ID'] = resolveCurrentMuzzle().song,
    }))
```

z16 (changes requested):
Most of it looks good, but please look through the copyright comments I made. If that is correct then this is ready to merge (the other things I wrote were just suggestions). Let me know when you're ready to have this merged.

addons/MountMuzzle/data/settings.xml
Sjshovan: Done.

addons/MountMuzzle/LICENSE
Sjshovan: Done.

addons/MountMuzzle/README.md
Sjshovan: Done.

addons/MountMuzzle/README.md
Sjshovan: Done.

addons/MountMuzzle/constants.lua
Sjshovan: Done.

addons/MountMuzzle/mountmuzzle.lua
Sjshovan: Done.

addons/MountMuzzle/helpers.lua
Sjshovan: Done.

addons/MountMuzzle/helpers.lua
Sjshovan: Instead of returning everything like: `return function() return packets, player, music, colors, muzzles end` I return a table so I can get specific objects from the file like: `local colors = require(constants').colors` in `helpers.lua` and still require everything within main.  Is the approach ok?

addons/MountMuzzle/mountmuzzle.lua
Sjshovan: Done.

addons/MountMuzzle/mountmuzzle.lua
Sjshovan: Yes! Have been using this for next mod, thank you! Done.

addons/MountMuzzle/helpers.lua
z16: Yes, that's fine, and like I said it wasn't *bad* before either, was just a suggestion to keep the logic more separated and modular. This is totally fine.

PR #1673, opened 2018-06-11
New addon: Tab

from20020516:
# Tab

### English
- Simple addon Replace Tab key input to `<stnpc>` or `<stpc>`. (X+Tab)
- **Tips:** if you won't to include pet or trust in `<stnpc>` by other players, use `/ignorepet on` and `/ignoretrust on` (in-game commands)

### 日本語
- Tabキーの動作を`<stnpc>`または`<stpc>`(X+Tab)に置き換えます。
- **Tips:** `<stnpc>`に他プレイヤーのペットやフェイスを含めたくない場合、`/ignorepet on` または `/ignorefaith on`（ゲーム内コマンド）をご利用ください。

### DIFF
--- a/addons/Tab/tab.lua
+++ b/addons/Tab/tab.lua
@@ -37,7 +37,7 @@ x_pressed = false
 windower.register_event('keyboard',function(dik,pressed,flags,blocked)
     if dik == 45 then
         x_pressed = pressed
-    elseif not windower.chat.is_open(true) then
+    elseif not windower.chat.is_open() then
         if dik == 15 and pressed and not st then --Tab
             st = x_pressed and '<stpc>' or '<stnpc>'
             windower.chat.input('/ta '..st)
--- a/addons/Tab/tab.lua
+++ b/addons/Tab/tab.lua
@@ -35,10 +35,10 @@ x_pressed = false
 
 --replace input tab. validated in Japanese keyboard. please let me know if problems occur with the English keyboard.
 windower.register_event('keyboard',function(dik,pressed,flags,blocked)
-    if not windower.chat.is_open() then
-        if dik == 45 then
-            x_pressed = pressed
-        elseif dik == 15 and pressed and not st then --Tab
+    if dik == 45 then
+        x_pressed = pressed
+    elseif not windower.chat.is_open(true) then
+        if dik == 15 and pressed and not st then --Tab
             st = x_pressed and '<stpc>' or '<stnpc>'
             windower.chat.input('/ta '..st)
             return true; --tab input blocking. it's probably broken..
--- a/addons/Tab/tab.lua
+++ b/addons/Tab/tab.lua
@@ -29,25 +29,20 @@ _addon.name = 'Tab'
 _addon.author = 'from20020516'
 _addon.version = '1.0'
 
-key_flags={}
+st = false
+x_pressed = false
 
 --replace input tab. validated in Japanese keyboard. please let me know if problems occur with the English keyboard.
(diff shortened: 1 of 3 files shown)


addons/Tab/tab.lua
```diff
@@ -0,0 +1,53 @@
...
+SOFTWARE, EVEN IF ADVISED OF THE POSSIBILITY OF SUCH DAMAGE.
+]]
+_addon.name = 'Tab'
+_addon.author = 'from20020516'
+_addon.version = '1.0'
+
+key_flags={}
+
+--replace input tab. validated in Japanese keyboard. please let me know if problems occur with the English keyboard.
+windower.register_event('keyboard',function(dik,flags,blocked)
```
z16: This is not the right signature. It should be:
```lua
windower.register_event('keyboard',function(dik, pressed, flags, blocked)
```
Where `pressed` is `true` if the key was pressed down and `false` if the key was released. Which also works better with what you're doing than calling it `flags`.

addons/Tab/tab.lua
```diff
@@ -0,0 +1,53 @@
...
+    key_flags[dik] = flags
+    if flags and not windower.chat.is_open(true) then
```
z16: This is my fault, `is_open` should not require an argument, but currently it does... I'll fix it later to not require the argument, so feel free to adjust it in the pull request already.

addons/Tab/tab.lua
```diff
@@ -0,0 +1,53 @@
...
+            return true; --tab input blocking. it's probably broken..
+        elseif dik == 1 or dik == 28 then --Esc or Enter
+            st = false
+        end
+    end
+end)
+
+--block error message.
+windower.register_event('incoming text',function(original,modified,original_mode)
+    if original_mode == 123 and st then
```
z16: When do you get this error? And why block it?

addons/Tab/tab.lua
```diff
@@ -0,0 +1,53 @@
...
+LOSS OF USE, DATA, OR PROFITS; OR BUSINESS INTERRUPTION) HOWEVER CAUSED AND
+ON ANY THEORY OF LIABILITY, WHETHER IN CONTRACT, STRICT LIABILITY, OR TORT
+(INCLUDING NEGLIGENCE OR OTHERWISE) ARISING IN ANY WAY OUT OF THE USE OF THIS
+SOFTWARE, EVEN IF ADVISED OF THE POSSIBILITY OF SUCH DAMAGE.
+]]
+_addon.name = 'Tab'
+_addon.author = 'from20020516'
+_addon.version = '1.0'
+
+key_flags={}
```
z16: I would define the `st` variable here and default it to `false`.

z16 (changes requested):
Mentioned a few minor things as well as a problem on my part which I'll fix later today. I'd like for both changes to go live at the same time.

addons/Tab/tab.lua
from20020516: When there is no <stnpc\> target in the field of view of the player.
It was simply annoying for me :P

addons/Tab/tab.lua
from20020516: wiki fixed :O
https://github.com/Windower/Lua/wiki/Events

from20020516: X key does not appear to be supported by `flags`. (I want to emulate the target filter of ffxiv)

from20020516:
changes done. thank you :)

PR #1674, opened 2018-06-12
[bugfix] Gearswap/triggers.lua

from20020516:
for work 'Area Display' effect (config Misc. 3) while <st\>, <stnpc\> and <stpc\> in JP client.
I'd like you to merge If there is no problem in English environment.


Byrth:
I suspect that this causes precast to not work on for \<st\>, \<stnpc\>, and \<stpc\> spells.

from20020516:
ummm...
In my environment both Japanese and English seem to be working without problems.

PR #1675, opened 2018-06-12
Add list command to Trusts addon.

fiveisprime:
Added `//tru list` command to show saved sets. Tested on the latest
client.

I'm not familiar with how you all do versioning so I didn't bump the
version.


from20020516:
Thank you for your help in improving my addon. I'm very happy! :)

PR #1682, opened 2018-06-25
New Addon: EasyNuke

deadman80:
Unified commands for single target and area of effect elemental magic and cures, plus Drain/Aspir and DRK Absorbs.


addons/EasyNuke/EasyNuke.lua
```diff
@@ -0,0 +1,207 @@
...
+
+Redistribution and use in source and binary forms, with or without
+modification, are permitted provided that the following conditions are met:
+
+* Redistributions of source code must retain the above copyright
+notice, this list of conditions and the following disclaimer.
+* Redistributions in binary form must reproduce the above copyright
+notice, this list of conditions and the following disclaimer in the
+documentation and/or other materials provided with the distribution.
+* Neither the name of findAll nor the
```
z16: Change `findAll` here to `EasyNuke`.

addons/EasyNuke/EasyNuke.lua
```diff
@@ -0,0 +1,207 @@
...
+spell_tables["dark"] = T{"Impact"}
+spell_tables["dark"]["ga"] = T{"Comet"}
+spell_tables["cure"] = T{"Cure","Cure II","Cure III","Cure IV","Cure V","Cure VI"}
+spell_tables["cure"]["ga"] = T{"Curaga","Curaga II","Curaga III","Curaga IV","Curaga V",}
+spell_tables["cure"]["ra"] = T{"Cura","Cura II","Cura III"} 
+spell_tables["drain"] = T{"Aspir","Aspir II","Aspir III","Drain","Drain II","Drain III"}
+spell_tables["absorb"] = T{"Absorb-Acc","Absorb-TP","Absorb-Attri","Absorb-STR","Absorb-DEX","Absorb-VIT","Absorb-AGI","Absorb-INT","Absorb-MND","Absorb-CHR"}
+
+function find_index ()
+		if current_element == "fire" then elements_index = 1 end
```
z16: Please use 4 spaces as indentation instead of hard tabs.

addons/EasyNuke/EasyNuke.lua
```diff
@@ -0,0 +1,207 @@
...
+spell_tables["light"]["ga"] = T{"Banishga","Banishga II"}
+spell_tables["dark"] = T{"Impact"}
+spell_tables["dark"]["ga"] = T{"Comet"}
+spell_tables["cure"] = T{"Cure","Cure II","Cure III","Cure IV","Cure V","Cure VI"}
+spell_tables["cure"]["ga"] = T{"Curaga","Curaga II","Curaga III","Curaga IV","Curaga V",}
+spell_tables["cure"]["ra"] = T{"Cura","Cura II","Cura III"} 
+spell_tables["drain"] = T{"Aspir","Aspir II","Aspir III","Drain","Drain II","Drain III"}
+spell_tables["absorb"] = T{"Absorb-Acc","Absorb-TP","Absorb-Attri","Absorb-STR","Absorb-DEX","Absorb-VIT","Absorb-AGI","Absorb-INT","Absorb-MND","Absorb-CHR"}
+
+function find_index ()
```
z16: This function could be replaced with a lookup table:
```lua
local indices = {
    fire = 1,
    wind = 2,
    thunder = 3,
    light = 4,
    ice = 5,
    water = 6,
    earth = 7,
    dark = 8,
}
```

Then you can just use `indices[current_element]` later.

addons/EasyNuke/EasyNuke.lua
```diff
@@ -0,0 +1,207 @@
...
+		local current_spell_table = spell_tables[current_element]["ra"]
+		windower.send_command("@input /ma "..current_spell_table[arg].." <"..target_mode..">")
+	end
+	
+	if command == "target" then
+		if arg then
+			arg = string.lower(arg)
+			target_mode = arg
+			windower.send_command("@input /echo Target mode is now: "..target_mode)
+		else targets_index = targets_index + 1
```
z16: Please properly indent this. Don't write after the `else` clause, but write it one below, and then indent it one level further than the `else` itself. I'd also appreciate splitting all `if` constructs up into multiple lines, it's generally much easier to assert what the code does that way.

addons/EasyNuke/EasyNuke.lua
```diff
@@ -0,0 +1,207 @@
...
+		if targets_index > 3 then targets_index = 1 end
```
z16: This process, increasing by one and then wrapping the number after a certain limit is very commonly used in this addon and can be simplified:
```lua
targets_index = targets_index % 3 + 1
```
`% 3` tells it to "wrap" around `3`. So if the value is `3` it would wrap to `0` and then add `1`, resulting in `1`. You can use it similarly in other places below, replacing `3` by whatever limit you use there.

addons/EasyNuke/EasyNuke.lua
```diff
@@ -0,0 +1,207 @@
...
+		windower.send_command("@input /echo Target Mode is now: "..target_mode)
+		end
+	end
+	
+	if command == "element" then
+		arg = string.lower(arg)
+		if elements:contains(arg) then
+			current_element = arg
+			windower.send_command("@input /echo Element Mode is now: "..string.ucfirst(current_element))
+		elseif other_modes:contains(arg) then
```
z16: These two branches (`if` and `elseif`) share the exact same code and could be combined into one branch:
```lua
if elements:contains(arg) or other_modes:contains(arg) then
```

addons/EasyNuke/EasyNuke.lua
```diff
@@ -0,0 +1,207 @@
...
+		else windower.send_command("@input /echo Invalid element")
+		end
+	end
+	
+	if command == "cycle" then
+		arg = string.lower(arg)
+		if arg == nil then
+			find_index()
+			elements_index = elements_index + 1
+			if elements_index > 8 then elements_index = 1 end
```
z16: If you took my suggestion from before these three lines could be combined into one:
```lua
            elements_index = indices[current_element] % 8 + 1
```
Doing this for the other places too will save you quite a few lines.

addons/EasyNuke/EasyNuke.lua
```diff
@@ -0,0 +1,207 @@
...
+			current_element = tostring(elements[elements_index])
```
z16: The more I think about it the more you don't even need the `current_element` variable. I would instead just store an index. You can then do `elements[element_index]` in the places where you need it. That lookup is cheap and it saves you from having to synchronize the `element_index` and the `current_element` everywhere, reducing the boilerplate code by quite a bit.

addons/EasyNuke/EasyNuke.lua
```diff
@@ -0,0 +1,207 @@
...
+
+	if command == 'boom' then
+		if arg == nil then arg = 1 end
+		arg = tonumber(arg)
+		if arg > #spell_tables[current_element] then windower.send_command("@input /echo Invalid Spell. Try again.") return end
+		local current_spell_table = spell_tables[current_element]
+		windower.send_command("@input /ma "..current_spell_table[arg].." <"..target_mode..">")
+	end
+	
+	if command == "boomga" then
```
z16: You should swap all the command-level `if` clauses with `elseif`. Will save needless comparisons.

addons/EasyNuke/EasyNuke.lua
```diff
@@ -0,0 +1,207 @@
...
+		elseif arg == "fragmentation" or arg == "frag" or arg == "f" then
+			if current_element ~= "thunder" and current_element ~= "wind" then current_element = "thunder"
+			elseif current_element == "thunder" then current_element = "wind"
+			elseif current_element == "wind" then current_element = "thunder"
+			end
+		end
+	end
+		
+	if command == "showcurrent" then
+		windower.send_command("@input /echo ----- Element Mode: "..string.ucfirst(current_element).." --- Target Mode: < "..target_mode.." > -----")
```
z16: `windower.chat.input` saves you from having to preface every command with `@input ` and will hopefully soon also be faster than `windower.send_command`.

addons/EasyNuke/EasyNuke.lua
```diff
@@ -0,0 +1,207 @@
...
+spell_tables["thunder"]["ga"] = T{"Thundaga","Thundaga II","Thundaga III","Thundaja",}
+spell_tables["thunder"]["ra"] = T{"Thundara","Thundara II","Thundara III"}
+spell_tables["light"] = T{"Banish","Banish II","Holy","Banish III",}
+spell_tables["light"]["ga"] = T{"Banishga","Banishga II"}
+spell_tables["dark"] = T{"Impact"}
+spell_tables["dark"]["ga"] = T{"Comet"}
+spell_tables["cure"] = T{"Cure","Cure II","Cure III","Cure IV","Cure V","Cure VI"}
+spell_tables["cure"]["ga"] = T{"Curaga","Curaga II","Curaga III","Curaga IV","Curaga V",}
+spell_tables["cure"]["ra"] = T{"Cura","Cura II","Cura III"} 
+spell_tables["drain"] = T{"Aspir","Aspir II","Aspir III","Drain","Drain II","Drain III"}
```
z16: I noticed the `drain` entry has spells for both *Drain* and *Aspir* listed. Is that intended? There does not seem to be a table entry for `aspir`.

z16 (changes requested):
I left a few notes in the file. On the whole, there are three mandatory changes that need to be made before we can accept this:
1. The license needs to be adjusted to use the right name
2. The hard tabs need to be changed to a 4 space indentation
3. There's an entry in the `addons.xml` file missing. Make sure you add that, or this addon will not show up in the launcher.

The other things I mentioned are not mandatory, merely suggestions on my part that would improve the code in terms of performance or readability.

There are two things I didn't mention in the file that I would just like to point out (also not mandatory).

First, you use `T{}` a lot to define tables. In almost all cases it's not necessary here and sometimes it comes at the cost of performance. The `other_modes` table is an example of where you might want to use `S{}` instead. Most other tables can drop it entirely. It's really only useful if you want to use functions on those tables that are in the `table` namespace.

Second, I think the usage could be streamlined with different and shorter names. Since this seems to be all about efficiency at the keyboard I think choosing different commands could improve it.

I would probably use the `ez` prefix only for setting up modes, not for actual nuking. For that you could just use `/boom` or probably better `/nuke` (since nuke has no repeating letters, which is harder to type fast). Even `/n` would work, and I don't think it's used for anything else (although I'm not sure).

I'd also allow the user to set up a default nuke level instead of defaulting to one. Maybe even per type of spell (default levels for single, -ga and -ra).

But again, both of those are just observations. The three mandatory issues will need to be addressed, I'll leave everything else up to your judgment.

addons/EasyNuke/EasyNuke.lua
deadman80: It is intended.  There's only three of each type, and I prefer it this way for practical use. The missing Aspir table was most likely an error from the last batch of change attempts to prevent the errors that get thrown out when a non-existent table gets called (ex: There's no spell_tables["absorb"]["ra"], so boomra command would spit out an error in the console in addition to the embedded error text.  Those errors are still there, but will continue working on eliminating them. )

addons/EasyNuke/EasyNuke.lua
deadman80: Is there any benefit to changing this?  Does it actually function differently or is this an alternative method to do the exact same thing?

addons/EasyNuke/EasyNuke.lua
deadman80: Done.

addons/EasyNuke/EasyNuke.lua
deadman80: How would I go about wrapping around a descending order, like I do for the "cycle back" command/argument?

addons/EasyNuke/EasyNuke.lua
deadman80: Done.

addons/EasyNuke/EasyNuke.lua
deadman80: Done.  Guess it's easy to figure out where I copy/pasted it from. XD
Is it kosher to use "FaceDesk Linkshell" as the name, or am I required to include individual names?

deadman80:
I greatly appreciate the feedback.  I basically started totally from scratch and learned what little LUA I know now in order to make this.  I do have a few questions about the streamlining suggestions made:

* I also tried S{} and {} tables, but kept breaking them.  T{} actually works everywhere it's used currently.  Not sure which ones I could/should change to S{}, or even truly understand what the differences are between them.  I verified all of the T{} and as of right now, changing any of them breaks functionality.

* I wanted to have //boom //boomga //boomra as direct commands, but adding them to either _addon.command or _addon.commands did not work for me. How would I go about adding them?

* Command names: I guess I could add some additional shortcut names, like what I did for the T2 modes.  I actually envisioned this being used primarily via macros and not typed out live, but don't mind making adjustments to facilitate doing so. 

* Requires: Do I have any superfluous requires?  I don't remember at this point why I have so many...

addons/EasyNuke/EasyNuke.lua
deadman80: Working on this now.  Since it's going to change the "cycle" logic, might take a little bit to wrap my head around what needs to change first tho.

As an aside, the current janky function is what I came up with after struggling with how to make a reverse lookup work and only finding "LUA doesn't do that" as an answer.. There was one suggestion to create a reversed table using "v, i", but I don't understand that yet.

addons/EasyNuke/EasyNuke.lua
deadman80: Done.  Shoulda caught that one myself tbh.

addons/EasyNuke/EasyNuke.lua
alphaONE2:
```lua
windower.add_to_chat(206, "----- Element Mode: "..string.ucfirst(current_element).." --- Target Mode: < "..target_mode.." > -----")
```
This is faster, less code, and lets you change the color of the text (the `206`) easily.

For the cases where you're trying to send a `/command` to the game,
```lua
windower.chat.input('/command')
```
This is much simpler, and faster, since it skips the Windower command handler.

addons/EasyNuke/EasyNuke.lua
alphaONE2:
```lua
targets_index = (targets_index + 1) % 3 + 1
```

More generally:
```lua
x = (x + N - 2) % N + 1
```
will cycle `x` backwards in the interval `[1, N]`

deadman80:
Small aside: This is the first time I've done anything but lurk on GitHub, so if I goofed something up with submitting the 2nd/corrected versions of the files, someone please let me know what I need to fix.   Thank you. ^^;;

addons/EasyNuke/EasyNuke.lua
z16: @alphaONE2 What do you think? I don't know :D

alphaONE2: IANAL, but should be ok.

addons/EasyNuke/EasyNuke.lua
```diff
@@ -0,0 +1,204 @@
...
+        
+    elseif command == "target" then
+        if arg then
+            arg = string.lower(arg)
+            target_mode = arg
+        else
+            targets_index = targets_index % 3 + 1
+            target_mode = targets[targets_index]    
+        end
+        windower.add_to_chat(206,"/echo Target Mode is now: "..target_mode)
```
alphaONE2: This prints `/echo` to the chat log too. I assume you didn't intend that?

addons/EasyNuke/EasyNuke.lua
```diff
@@ -0,0 +1,204 @@
...
+        
+    elseif command == "element" or command == "mode" then
+        arg = string.lower(arg)
+        if elements:contains(arg) or other_modes:contains(arg) then
+            current_element = arg
+            windower.add_to_chat(206,"/echo Element Mode is now: "..string.ucfirst(current_element))
```
alphaONE2: Ditto here

addons/EasyNuke/EasyNuke.lua
```diff
@@ -0,0 +1,204 @@
...
+        else
+            windower.add_to_chat(206,"/echo Invalid element")
```
alphaONE2: And here

alphaONE2 (changes requested):

deadman80:
Those /echos were not intentional and have been removed.

alphaONE2:
Looks like you had some merge conflicts on that last commit.

z16:
### About `T{}` vs `S{}` vs `L{}` vs `{}`

`{}` is a default Lua table and should work in every place you currently use `T{}` *except* for `elements` and `other_modes`, because you use `:contains` on both of them. You cannot do that with regular Lua tables, but can do it with both `S{}` and `T{}`
`T{}` is a wrapper around basic Lua tables and enables them to be accessed as if they were objects with methods. The `elements:contains(arg)` you do for example would not work on regular tables, but `T{}` tables translate that to `table.contains(elements, arg)`, which *does* work on regular Lua tables (although it would still require you to `require('tables')` since `table.contains` isn't a regular table feature provided by Lua alone).
`S{}` is a wrapper for sets and it supports almost all methods that `T{}` tables do, but you cannot index them by number, which is what you do a lot. Hence, you cannot use `S{}` in most cases. The only place you can use those in is for the `other_modes` table, since you never access it with an index, but only call `:contains` on it. The advantage of `S{}` tables is that lookup is much faster than in all other kinds of tables.
`L{}` is a wrapper for lists, ordered sets of elements. It's very rarely used, but provides a slightly faster alternative to regular Lua tables in certain cases.

Quick guidelines for when to use what:
Use `{}` for all tables that you don't really use outside of looking up values. This is what I would have used for all tables in your addon except `other_modes`.
Use `S{}` for a collection of values where **order does not matter** and if you need to look up values. (`require('sets')`) This is what I would have used for `other_modes`.
Use `L{}` for a collection of values where **order does matter**. (`require('lists')`) I wouldn't have used this in your addon.
Use `T{}` for key => value mappings which you intend to modify and work with a lot (search, filter, map, etc.). (`require('tables')`) I wouldn't have used this in your addon.

That said, as long as it works no one is gonna pay close attention to what you should use. Like I said in my review yesterday it was a very minor suggestion. These guidelines are more useful when you're writing performance critical code, but otherwise you can use whatever you're comfortable with. But you might still keep these notes in mind for the future.

### About how to handle multiple commands differently

This doesn't work all that well in v4 as you can only register commands to the same function. So you can only have commands that are aliases of each other and do the same thing. To get what you want you have to use the `unhandled command` event.

```lua
windower.register_event('unhandled command', function(command, ...)
    if command == 'boom' then
        -- stuff
    elseif command == 'boomga' then
        -- stuff
    elseif command == 'boomra' then
        -- stuff
    end
end)
```

That way you can avoid the `//ez` for those commands.

### Requires

Unused requires:
```lua
-- require('chat')
-- require('lists')
-- require('logger')
require('sets')
require('tables')
require('strings')
-- require('pack')

-- file  = require('files')
config = require('config')
-- texts = require('texts')
-- res = require('resources')
```

z16:
There's still a merge conflict in the readme file :)

deadman80:
The formatting that I used for the readme apparently looks like junk when viewed here. ^^;; Is there a way to do nested bullet points, or should I just remove/reduce the formatting and double space the file?

deadman80:
> The formatting that I used for the readme apparently looks like junk when viewed here. ^^;; Is there a way to do nested bullet points, or should I just remove/reduce the formatting and double space the file?

Google Search is a wonderful thing. >_>;;  Accidentally learned how to quote while learning how to do bullet points too.   \o/

PR #1683, opened 2018-06-27
Compensate for extra UI scaling in text functions

trv6:
Quick fix for texts library UI-scale related mouse issues.

### DIFF
--- a/addons/libs/texts.lua
+++ b/addons/libs/texts.lua
@@ -129,6 +129,29 @@ local apply_settings = function(_, t, settings)
     call_events(t, 'reload')
 end
 
+-- The numbers returned by the following two text functions 
+-- seem to be scaled twice by a factor equal to the UI scale. 
+-- We reverse the extra scaling.
+do
+    local windower_settings = windower.get_windower_settings()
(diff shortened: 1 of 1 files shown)


z16:
I'm not sure how I feel about changing those functions globally inside a library. If someone happens to use those functions along with the `texts` library the behavior will suddenly change for them. We should either compensate for it at the precise points where we use the function inside of the `texts` library, or fix it inside LuaCore, which I feel makes more sense. Would you mind if I adjusted LuaCore instead to return properly adjusted values for those two functions?

trv6:
Not at all.

qyyyyyyyyyyyyyyyyyyyy:
lua python java

Tny5989:
Any updates on this issue?  It would be nice to have this fix but I can see how it wouldn't really be a priority.

xurion:
+1
This affects everyone who wants to drag a text block on a 4k screen where Windows applies a 150% scale-up by default, making it near impossible to reorganise smaller text blocks.

Chiaia:
This has been fixed in the last few months. I don't have a pull request # this time but confirmed with Arcon who fixed it in Luacore back then.

PR #1690, opened 2018-07-20
Part of v2 but other's seem interested in updating also. Update rolltracker.lua

Chiaia:


addons/rolltracker/rolltracker.lua
```diff
@@ -207,13 +208,17 @@ windower.register_event('action', function(act)
             isLucky = false
             if rollNum == rollInfo[rollID][15] or rollNum == 11 then 
                 isLucky = true
+                log('Lucky roll')
```
z16: Is this meant to display `Lucky roll` with the `log` function? Shouldn't it go below into the `add_to_chat`?

PR #1703, opened 2018-09-07
Mount Muzzle update to v0.9.4

Sjshovan:


addons/MountMuzzle/mountmuzzle.lua
```diff
@@ -102,84 +121,136 @@ function resolveCurrentMuzzle()
...
+    tryInject()
+end)
+
+windower.register_event('unload', function() 
+    injectMusic(music.types.mount, muzzles.zone.song)
+end)
+
+windower.register_event('zone change', function() 
+    tryInject()
+end)
```
z16: Not too important, but you can chain events into a single `register_event` call:
```lua
windower.register_event('login', 'logout', 'zone change', function()
    tryInject()
end)
```

z16 (approved):

addons/MountMuzzle/mountmuzzle.lua
Sjshovan: oh, awesome, thanks!

PR #1705, opened 2018-09-11
Gracefully handle items missing from res/items

RubenatorX:
If an item is missing from res/items, rather than erroring, display item name as "NewItem#1234" where 1234 is the item's ID.
This allows the plugin to function when new items drop and have not yet been added to res/items.


z16:
This was a good idea, but I figured we could add a generic mechanism for it. I adjusted the resources library so when you try to access an unknown ID it returns a dummy item with only the name `<new item #id>`. That way it'll apply to all addons that use resources automatically and will hopefully save us some resource based error reports on updates. So I'm gonna close this, as the behavior is now standard in the resources library. Thanks for suggesting this.

PR #1708, opened 2018-10-19
Overhaul and upgrade of EnemyBar

AkadenTK:
I wanted more features than just a target HPP bar. New features include:
- casting/performing action tracking (graphical)
- crowd control display (if a target/subtarget/etc is bound, slept, petrified, terrorized)
- distance tracking
- rudimentary target-of-target display (essentially, hate tracking, but not as good)
- aggro bars (see: ffxiv) to display health bars and all of the above for enemies that have targeted you or party members.
- focus target bar (see: ffxiv) to display health bar and all of the above for a specified target (player or npc) whether they are targeted or not. 

There are all new commands to control all these settings, as well as specify a focus target. There's a setup mode that allows target bar and aggro frame dragging. Hold control while dragging to "snap" to a grid.

The default modes for all of the above are off. I don't want to disrupt what everyone when they log in for the first time after this change. Settings upgrade automatically, preserving their settings from the old version.


addons/enemybar/bars.lua
```diff
@@ -0,0 +1,284 @@
...
+	o.rcap:destroy()
+	o.ntext:destroy()
+	o.atext:destroy()
+	o.atar:destroy()
+	o.ttext:destroy()
+	o.dtext:destroy()
+	o.tstat:destroy()
+end
+
+-- Base class method printArea
```
z16: What is `printArea`?

addons/enemybar/bars.lua
```diff
@@ -0,0 +1,284 @@
+
+-- Meta class
+bars = {x_res = windower.get_windower_settings().x_res,y_res = windower.get_windower_settings().y_res}
+
+-- Base class method new
+
+function bars.new(o, bar_settings)
```
z16: You only ever call it with `nil` as `o`. If this is meant as an addon-specific library I'd remove the parameter completely, as it would also fit with the `new` moniker.

addons/enemybar/bars.lua
```diff
@@ -0,0 +1,284 @@
...
+	o.lcap:destroy()
+	o.bg_body:destroy()
+	o.fg_body:destroy()
+	o.rcap:destroy()
+	o.ntext:destroy()
+	o.atext:destroy()
+	o.atar:destroy()
+	o.ttext:destroy()
+	o.dtext:destroy()
+	o.tstat:destroy()
```
z16: Can't say I'm a big fan of these names :D Would appreciate if you would name them something more specific as it would ease the review process, since it would be more indicative of what each piece is meant to do.

addons/enemybar/actionTracking.lua
```diff
@@ -0,0 +1,299 @@
...
+
+tracked_actions = {}
+tracked_enmity = {}
+tracked_debuff = {}
+framerate = 75
+clean_actions_delay = framerate
+clean_actions_tick = clean_actions_delay
+
+function looking_at(a, b)
+	local h = a.facing % math.pi
```
z16: You're mixing different indentation styles throughout the file. Please commit to 4 spaces.

addons/enemybar/actionTracking.lua
```diff
@@ -0,0 +1,299 @@
...
+	local h2 = (math.atan2(a.x-b.x,a.y-b.y) + math.pi/2) % math.pi
+	return math.abs(h-h2) < 0.15
+end
+
+function is_party_member_or_pet(mob)
```
z16: This function is rather slow (through little fault of your own). Checking on every action for every target, having to fetch entity objects through the already slow API (and occasionally more than once) and comparing many values for each entity on every action packet may be fast on your PC, but I would not be surprised if this would slow down other player's computers significantly.

There are a number of ways around that, most helpfully though, caching and removing some of the unnecessary checks. You can assume that any function in `windower.ffxi.` runs very poorly, especially `windower.ffxi.get_items`, `windower.ffxi.get_party` and `windower.ffxi.get_player`, two of which you use here. The player ID could be cached throughout the addon for example, as it only changes during logout/login (at which point you can update it).

You could also do some pre-filtering of the ID. For it to be a player ID it must be `id < 0x01000000`, and for it to be a pet ID it must be `id > 0x01000000 && id % 0x1000 > 0x700`. If it is neither, the function may return `false` safely.

But here you can use caching as well, and it should be pretty simple, since you only need to capture the party update packet (0x0DD) as well as the pet info packet (0x067) and build a simple dictionary from there. That would probably be the best solution for the given effort (not much).

addons/enemybar/actionTracking.lua
```diff
@@ -0,0 +1,299 @@
...
+function handle_action_packet(id, data)
+	if 0x028 == id then
+		local ai = windower.packets.parse_action(data)
+
+		track_enmity(ai)
+		track_actions(ai)
+		track_debuffs(ai)
+	elseif 0x029 then
+    	local message_id = data:unpack('H',0x19)
+    	if not message_id then return end
```
z16: This check can be removed, it never fails (you have a habit of overchecking things throughout the addon, such as checking for `p` above, and a few more times below, I will try to mention it if I think of it).

addons/enemybar/actionTracking.lua
```diff
@@ -0,0 +1,299 @@
...
+    	message_id = message_id%32768
```
z16: I'd use `0x8000` here, seems like less of a magic number.

addons/enemybar/actionTracking.lua
```diff
@@ -0,0 +1,299 @@
...
+		local param_1 = data:unpack('I',0x0D)
+    	local target_id = data:unpack('I',0x09)
+		if S{204,206}:contains(message_id) then
+			-- wears off message.
+			local target = windower.ffxi.get_mob_by_id(target_id)
```
z16: You get the `target` object here, but you only use it for its `.id` field, which you already have (in the `target_id` variable). `windower.ffxi.get_mob_by_id` is not the slowest function in our API, but it certainly is not fast, it should be avoided if not necessary (so much so that I seriously recommend caching its value when it is not hard to implement).

addons/enemybar/actionTracking.lua
```diff
@@ -0,0 +1,299 @@
...
+			if target and tracked_debuff[target.id]  then
+				tracked_debuff[target.id][param_1] = nil
+			end
+		end
+	end
+end
+
+function track_enmity(ai)
+	local actor = windower.ffxi.get_mob_by_id(ai.actor_id)
```
z16: As I mentioned before, `windower.ffxi.get_mob_by_id` is a slow function, and you call it on the actor and targets in this function, and below in `track_action` and `track_debuffs`. I would call it once in the calling function, and then pass those mob objects as arguments. Also allows you to remove more `if` checks in here, if either of them is `nil` in the calling function you can just return.

addons/enemybar/actionTracking.lua
```diff
@@ -0,0 +1,299 @@
...
+	if not actor then return end
+
+	for i,t in ipairs(ai.targets) do
+		local target = windower.ffxi.get_mob_by_id(t.id)
+		if target then
+			local pc = nil
```
z16: Just fyi, the ` = nil` is not necessary here and can be removed. You can declare locals with just `local foo` and it will have `nil` as its default value.

addons/enemybar/actionTracking.lua
```diff
@@ -0,0 +1,299 @@
...
+	local actor = windower.ffxi.get_mob_by_id(ai.actor_id)
+	if not actor then return end
+
+	local target = windower.ffxi.get_mob_by_id(ai.targets[1].id)
+	if not target then
+		target = {name='Unknown (id:'..ai.targets[1].id, id=ai.targets[1].id}
+	end
+
+	-- if the category is not casting magic, jas, items or ws, don't bother.
+	if not S{8,4,7,11,3,6,9,5}:contains(ai.category) then return end
```
z16: This whole `S{1,2,3}:contains` thing has gotten a bit out of hand, and having started it I take full responsibility for it. This is also "slow", although I admit it is not enough to matter much in the end, it just bugs me personally (I would not mind it at all if you left it in). The problem with this is that it constructs the set each time the function is called. It constructs a new table with the values `{8, 4, 7, 11, 3, 6, 9, 5}` and then passes it to the `S` function, which copies all those values into a new table, which is then checked for with `:contains`.

Like I said, it's not a great optimization, but if you care for that kind of thing, you can define these fixed number sets outside of the function and just reuse them:
```lua
local ma_ja_item_ws = S{8, 4, 7, 11, 3, 6, 9, 5}
function track_actions(ai)
    --[[ ... ]]

    if not ma_ja_item_ws:contains(ai.category) then
        return
    end

    --[[ ... ]]
end
```

addons/enemybar/actionTracking.lua
```diff
@@ -0,0 +1,299 @@
...
+	end
+
+	-- find the action
+	local action_map = nil
+	if S{8,4}:contains(ai.category) then
+		action_map = res.spells[action_id]
+	elseif S{5,9}:contains(ai.category) then
+		action_map = res.items[action_id]
+	else
+		if actor.is_npc then
```
z16: You can combine this and the previous line into an `elseif`. Same for the check below.

addons/enemybar/actionTracking.lua
```diff
@@ -0,0 +1,299 @@
...
+    local overwrites = res.spells[spell].overwrites or {}
+    if not did_overwrite(target, spell, overwrites) then
+        return
+    end
+
+    tracked_debuff[target.id][effect] = {target=target,spell=spell,effect=effect,time=os.time(),duration=res.spells[spell].duration or 0,pos={x=target.x,y=target.y}}
+end
+
+function did_overwrite(target, new, t)
+    if not tracked_debuff[target.id] then
```
z16: This check can never be `true`. You check for that already in the calling function, and if `true` you set an empty table as the value, which will make this check always evaluate to `false`.

addons/enemybar/actionTracking.lua
```diff
@@ -0,0 +1,299 @@
...
+        return true
+    end
+    
+    for effect, tracked in pairs(tracked_debuff[target.id]) do
+        local old = res.spells[tracked.spell].overwrites or {}
+        
+        -- Check if there isn't a higher priority debuff active
+        if table.length(old) > 0 then
```
z16: Can remove this check. If the length is `0` the loop will immediately break and you get the same effect. Same for the `if` check below (line 213)

addons/enemybar/actionTracking.lua
```diff
@@ -0,0 +1,299 @@
...
+                    tracked_debuff[target.id][effect] = nil
+                end
+            end
+        end
+    end
+    return true
+end
+
+function check_conflicting_debuffs(ai)
+	local actor = windower.ffxi.get_mob_by_id(ai.actor_id)
```
z16: Another function that calls `windower.ffxi.get_mob_by_id` which could get the entity via an argument instead.

addons/enemybar/actionTracking.lua
```diff
@@ -0,0 +1,299 @@
...
+		-- the actor is an npc, let's check if they're supposed to be asleep/petrified/terror'd
+		local debuffs = tracked_debuff[actor.id]
+		if not debuffs then return end
+
+		for id,debuff in pairs(debuffs) do
+			if S{2,19,7,28}:contains(id) then
+				-- it was inactive, but now it's doing!
+				tracked_debuff[actor.id][id] = nil
+			end
+			if math.abs(debuff.pos.x-actor.x)>0.5 or math.abs(debuff.pos.y-actor.y)>0.5 then
```
z16: This and the previous `if` check could be joined with an `or`, but this may actually be better for readability... just pointing it out.

addons/enemybar/actionTracking.lua
```diff
@@ -0,0 +1,299 @@
...
+
+		-- for complete actions, timeout at 3s.
+		elseif action.complete and time - action.time > 3 then
+			tracked_actions[id] = nil
+		end
+	end
+
+	for id,enmity in pairs(tracked_enmity) do
+		if time - enmity.time > 3 then
+			local mob = windower.ffxi.get_mob_by_id(enmity.mob.id)
```
z16: You call that function a few more times in here. These may be new IDs, so not sure if there's a way around it, just pointing it out. I won't make a mention of this again, just search for those occurrences and see what you find that you can optimize.

addons/enemybar/bars.lua
```diff
@@ -0,0 +1,284 @@
...
+   o.show_target_icon = bar_settings.show_target_icon
+   o.show_action = bar_settings.show_action
+   o.show_debuff = bar_settings.show_debuff
+   bars.initialize(o)
+   bars.move(o, bar_settings.pos.x, bar_settings.pos.y)
+   return o
+end
+
+function bars.destroy(o)
+	if not o then return end
```
z16, line 24: You check for the existence of `o`, despite it never being `nil`, and also checking again at the call site. You can pretty much remove every one of these `if` checks in this module. It is good practice to do this for public facing libraries to validate unknown input (although I personally do not always agree with that either), but for internal libraries you always control the input, so it is not necessary.

addons/enemybar/enemybar.lua
```diff
@@ -24,31 +24,524 @@ SOFTWARE, EVEN IF ADVISED OF THE POSSIBILITY OF SUCH DAMAGE.
...
 
 config = require('config')
 images = require('images')
 texts = require('texts')
-require('gui_settings')
-require('targetBar')
-require('subtargetBar')
+table = require('table')
+require 'bars'
+require 'actionTracking'
```
z16: If you care for my feelings at all, please surround the required module names with parentheses :(

addons/enemybar/enemybar.lua
```diff
@@ -24,31 +24,524 @@ SOFTWARE, EVEN IF ADVISED OF THE POSSIBILITY OF SUCH DAMAGE.
...
+        -- settings.pos was the old position setting. if it's set that means we're upgrading from 1.0
+        defaults.target_bar.pos = settings_old.pos
+        defaults.target_bar.font = settings_old.font
+        defaults.subtarget_bar.font = settings_old.font
+        defaults.target_bar.font_size = settings_old.font_size
+        defaults.subtarget_bar.font_size = settings_old.font_size
+    end
+    settings = config.load(defaults)
+
+    initialize_bars()
```
z16: The `config` library has a feature you can use for this:
```lua
config.register(settings, initialize_bars)
```

This will execute `initialize_bars` every time the settings are reloaded (i.e. once initially at the start, but also when you relog into a different character, with different settings. You can also manually trigger it with `config.reload(settings)`, as you could do above in the `set_setting` function.

addons/enemybar/enemybar.lua
```diff
@@ -24,31 +24,524 @@ SOFTWARE, EVEN IF ADVISED OF THE POSSIBILITY OF SUCH DAMAGE.
...
+            if t and t.id == target.id then 
+                target_type = 1
+            elseif st and st.id == target.id then
+                target_type = 2
+            end
+            bars.update_target(bar, target.name, target.hpp, dist, target_type)
+
+            local action = tracked_actions[target.id]
+            if action and not action.complete then
+                    bars.update_action(bar, action.ability.en, '')
```
z16: Use `.name` instead of `.en` (and use `.english` over `.en` in any case). `.name` will resolve to `.english` on NA clients and to `.japanese` on JP clients.

addons/enemybar/enemybar.lua
```diff
@@ -24,31 +24,524 @@ SOFTWARE, EVEN IF ADVISED OF THE POSSIBILITY OF SUCH DAMAGE.
...
+        local mobname = mob.name:lower()
+        if n == mobname then
+            return mob
+        elseif not worse_match and mobname:sub(1, #n) == n then
+            worse_match = mob
+        elseif not worser_match and mobname:match(n) then
+            worser_match = mob
+        end
+    end
+    return worse_match and worse_match or worser_match
```
z16: This can be just `return worse_match or worser_match`.

addons/enemybar/enemybar.lua
```diff
@@ -24,31 +24,524 @@ SOFTWARE, EVEN IF ADVISED OF THE POSSIBILITY OF SUCH DAMAGE.
...
 end)
+windower.register_event('keyboard', function(dik, down) -- lol diks
+    if dik == 29 then -- if ctrl
+        drag_snap_key_down = down
+        --windower.add_to_chat(1,'ctrl: '..tostring(down))
+    end
+end)
+
+if windower.ffxi.get_info().logged_in then
+    player_id = windower.ffxi.get_player().id
```
z16, line 582: Here you already do cache the player ID as I suggested above. You can make use of that.

addons/enemybar/enemybar.lua
```diff
@@ -24,31 +24,524 @@ SOFTWARE, EVEN IF ADVISED OF THE POSSIBILITY OF SUCH DAMAGE.
...
+    else
+        settings[bar..'_bar'][setting] = v
+    end
+    windower.add_to_chat(207, 'EnemyBar: "'..setting..'" updated for "'..bar..'" bar')
+    settings:save()
+    initialize_bars()
+end
+
+function load_settings()
+    defaults = {}
```
z16: The `config` library has a feature where it reloads settings based on a new character itself. You do not need to define this function. Just define the defaults globally and do `local settings = config.load(defaults)`. After a logout/login this object will already contain updated values based on the settings of the new character. See also my note below about `config.register`.

addons/enemybar/enemybar.lua
```diff
@@ -24,31 +24,524 @@ SOFTWARE, EVEN IF ADVISED OF THE POSSIBILITY OF SUCH DAMAGE.
...
+windower.register_event('keyboard', function(dik, down) -- lol diks
+    if dik == 29 then -- if ctrl
+        drag_snap_key_down = down
+        --windower.add_to_chat(1,'ctrl: '..tostring(down))
+    end
+end)
+
+if windower.ffxi.get_info().logged_in then
+    player_id = windower.ffxi.get_player().id
+    load_settings()
```
z16: You should not need to call this, either here or in the `login`/`logout` events. See my comment above at `load_defaults`.

addons/enemybar/enemybar.lua
```diff
@@ -24,31 +24,524 @@ SOFTWARE, EVEN IF ADVISED OF THE POSSIBILITY OF SUCH DAMAGE.
...
+end
```
z16: There is a new line missing at the end of the file!

z16 (changes requested):
While none of the issues are crucial, I do suggest changes mostly with respect to the performance problem and the deeply nested, repeated calls of `windower.ffxi.get_mob_by_id`. Any `windower.ffxi.get_mob_by_X` function is slow because it has to copy the entire entity from its native data structure to a Lua table, and push dozens of values. Of all of them `windower.ffxi.get_mob_by_id` is often the worst (except for `windower.ffxi.get_mob_by_name`), since it may have to search the entire mob array for the ID (except in certain cases where it can infer the index from the ID, which leads to a direct access, but still makes it just as bad as all the other mob retrieval functions).

Also you should adjust the `.en` thing to `.name` and fix the indentation formatting. And you'd save yourself some effort if you used the `config` library to its full effect (i.e. with its auto-reloading and `config.register`).

addons/enemybar/actionTracking.lua
AkadenTK: Actually, I added this in testing. It was actually failing sometimes without that check. The packet library may have been returning a nil packet because it didn't read properly.

addons/enemybar/bars.lua
AkadenTK: It actually is called when o could be nil. it's an edge case though. I put it in here to clean up the code in enemybar.lua, since it's used in more places.

addons/enemybar/bars.lua
AkadenTK: I agree, I had them short for easy typing, but always planned on changing them. things slip the mind :P

addons/enemybar/bars.lua
AkadenTK: Something I copy-pasted. Ended up ditching the OOP style because I couldn't figure it out in lua.

PR #1709, opened 2018-11-01
SpellBook: Filter out UC trusts

sigilbaram:
The UC trusts are mutually exclusive and not learned in the the normal way, so they probably shouldn't be listed.


sigilbaram:
Something isn't right here...

PR #1724, opened 2018-12-22
fixes #1723 - correcting wrong "current_job" static variable use

### LINKED ISSUE #1723: [battlemod] Job filters are not reloaded
Actual result: When using `bm reload` command, job filters are not reloaded.

Expected result: `bm reload` should reload the current job's filters.

mharrington622:


addons/battlemod/battlemod.lua
```diff
@@ -143,6 +143,11 @@ function flip_block_cannot()
     block_cannot = not block_cannot
 end
 
+function reload()
```
z16: I would just inline that function while you fix the tabs :)

z16 (changes requested):

PR #1749, opened 2019-04-17
Refactoring TreasurePool

KenshiDRK:
Refactored to use packets instead of memory

### DIFF
--- a/addons/TreasurePool/TreasurePool.lua
+++ b/addons/TreasurePool/TreasurePool.lua
@@ -1,4 +1,4 @@
---[[Copyright © 2017, Kenshi
+--[[Copyright © 2019, Kenshi
 All rights reserved.
 
 Redistribution and use in source and binary forms, with or without
@@ -26,7 +26,7 @@ SOFTWARE, EVEN IF ADVISED OF THE POSSIBILITY OF SUCH DAMAGE.]]
 
 _addon.name = 'TreasurePool'
 _addon.author = 'Kenshi'
-_addon.version = '1.0'
+_addon.version = '2.0'
 
 require('luau')
 texts = require('texts')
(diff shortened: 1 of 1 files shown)


addons/TreasurePool/TreasurePool.lua
```diff
@@ -54,115 +54,113 @@ defaults.display.text.alpha = 255
...
 end)
 
-windower.register_event('prerender', function()
-    local treasure = T(windower.ffxi.get_items().treasure)
-    local remove = S{}
-    local info = S{}
-    if zoning_bool or treasure:empty() then
-        treasure_text:update(info)
-        treasure_text:hide()
+function Update()
```
z16: Why did you move this out from `prerender`? If you're worried about efficiency, I don't think you need to be, nothing very complicated happens here. You can still improve it a bit by precomputing as much as possible. I know `prerender` is often frowned upon, but adjusting things that are rendered on-screen is exactly what it was made for.

addons/TreasurePool/TreasurePool.lua
```diff
@@ -54,115 +54,113 @@ defaults.display.text.alpha = 255
...
+    local current_string = ''
```
z16: Don't need to define it here if you just return immediately. Can just do `local current_string = 'Treasure Pool:'` below.

addons/TreasurePool/TreasurePool.lua
```diff
@@ -54,115 +54,113 @@ defaults.display.text.alpha = 255
...
+    if items:empty() then
```
z16, line 130: One thing that isn't really related to this PR but which I would personally like is to *not* hide the box. Just display an empty box. Maybe have a setting for it, since others would probably like it to disappear. Just mentioning it.

addons/TreasurePool/TreasurePool.lua
```diff
@@ -54,115 +54,113 @@ defaults.display.text.alpha = 255
...
-    end
-end)
+    box.current_string = current_string
+end
+
+Update:loop(0.5)
+
+windower.register_event('logout','zone change', function()
+    items = T{}
+end)
```
z16: Why no new line? :(

z16 (approved):
Looks ok, the things I mentioned are just suggestions. Let me know if you want me to merge this as is, or if you want to modify some of it.

addons/TreasurePool/TreasurePool.lua
KenshiDRK: I did mainly for efficiency and cause I don't really need it to update on every frame, but if efficiency is not a problem I don't mind changing it back.

addons/TreasurePool/TreasurePool.lua
KenshiDRK: When I started the refactoring I didn't hide the box, but if you don't hide the box a little dot remains even if the box is empty, for me its np cause I don't use background (just the strings with stroke), but for some ppl that uses background it was annoying. Thats also the reason I was redefining current_string as empty string in every update, but if Im hiding I can erase that line.

addons/TreasurePool/TreasurePool.lua
KenshiDRK: Dunno. Its new line at end a good coding practice? or for another reason?

addons/TreasurePool/TreasurePool.lua
z16: Well, up to you really, but this is pretty cheap compared to what most people do per frame.

addons/TreasurePool/TreasurePool.lua
z16: Depends on whom you ask, to me it is. That's why we added it to our coding guidelines for v5 :) But it was just a random comment, as I said before, the code is fine as is.

alphaONE2: Not having a newline at the end of a text file can cause problems with some tools. These tools and systems treat the newline character as a line terminator, not a line separator. Many editors/viewers (including GitHub) will give warnings about it for this reason.

PR #1753, opened 2019-05-22
Fix to allow on screen dmg

KenshiDRK:
Fix to allow on screen dmg


cairthenn:
This will make the message timing different for non-numbered and numbered actions. I can't imagine that feels too normal?

Edit: Nevermind, it keeps the original, adds the modified, then removes the original from chat?

KenshiDRK:
yea, it blocks the original from the incoming text event.

KenshiDRK:
This fix doesn't work if simplify is set to false, so I'll update it in a few.

PR #1754, opened 2019-05-30
Added Delete function to AzureSets

DYDR:


addons/azureSets/azuresets.lua
```diff
@@ -210,6 +210,16 @@ function save_set(setname)
     notice('Set '..setname..' saved.')
 end
 
+function delete_set(setname)
+	if settings.spellsets[setname] == nil then
```
Chiaia: Need to replace the tab with spaces.

addons/azureSets/azuresets.lua
```diff
@@ -248,7 +258,10 @@ windower.register_event('addon command', function(...)
             if args[1] ~= nil then
                 save_set(args[1])
             end
-
+		elseif comm == 'delete' then
```
Chiaia: Need to replace the tabs with spaces.

addons/azureSets/azuresets.lua
DYDR: Updated

addons/azureSets/azuresets.lua
DYDR: Updated

Chiaia:
Also needs a version bump then it's good to go.

PR #1760, opened 2019-06-26
Update fields for incoming price response packet (0x03D) from using 'Bag' to 'Type'

ibm2431:
Just going to quote what I said on Discord~

> I suspect the 'Bag' value might really be a flag indicating to the client if the incoming "price response" packet is either an appraisal or a sale.
> 
> I had an addon that would listen for the packet and spit the price to chatlog / filelog: https://github.com/ibm2431/addons/blob/56905074b1f0b13ada3f9fa6c1d01ef7387f8c76/pricelog/pricelog.lua#L194-L210 
> 
> And it was triggering both on the initial price request (with the correct item name from the player's current Inventory), and then later on the sale, but with an item name from the player's Mog Safe

Basically, I think SE uses the packet for double-duty:
0: To send the appraisal price to the client
1: To send the "you sell a(n) <item> to the shop" message

### DIFF
--- a/addons/libs/packets/fields.lua
+++ b/addons/libs/packets/fields.lua
@@ -2132,12 +2132,13 @@ fields.incoming[0x03C] = L{
     {ref=types.shop_item,       label='Item',               count='*'},         -- 08 -   *
 }
 
--- Price response
--- Sent after an outgoing price request for an NPC vendor (0x085)
+-- Price/sale response
+-- Sent in response to an outgoing price request for an NPC vendor (0x085), and in response to player finalizing a sale.
+-- Inventory bag is assumed to be regular Inventory (0)
 fields.incoming[0x03D] = L{
     {ctype='unsigned int',      label='Price',              fn=gil},            -- 04
     {ctype='unsigned char',     label='Inventory Index',    fn=invp+{0x09}},    -- 08
-    {ctype='unsigned char',     label='Bag',                fn=bag},            -- 09
+    {ctype='unsigned char',     label='Type'},                                  -- 09 0 = on apprasial, 1 = when sale is finalized
     {ctype='unsigned short',    label='_junk1'},                                -- 0A
     {ctype='unsigned int',      label='_unknown1',          const=1},           -- 0C
 }
--- a/addons/libs/packets/fields.lua
+++ b/addons/libs/packets/fields.lua
@@ -2139,7 +2139,7 @@ fields.incoming[0x03D] = L{
     {ctype='unsigned char',     label='Inventory Index',    fn=inv+{0}},        -- 08
     {ctype='unsigned char',     label='Type'},                                  -- 09 0 = on price check, 1 = when sale is finalized
     {ctype='unsigned short',    label='_junk1'},                                -- 0A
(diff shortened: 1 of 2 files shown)


addons/libs/packets/fields.lua
```diff
@@ -2132,12 +2132,13 @@ fields.incoming[0x03C] = L{
     {ref=types.shop_item,       label='Item',               count='*'},         -- 08 -   *
 }
 
--- Price response
--- Sent after an outgoing price request for an NPC vendor (0x085)
+-- Price/sale response
+-- Sent in response to an outgoing price request for an NPC vendor (0x085), and in response to player finalizing a sale.
+-- Inventory bag is assumed to be regular Inventory (0)
```
Nifim: Comment on line 2137 does not seem to be needed.

addons/libs/packets/fields.lua
```diff
@@ -2132,12 +2132,13 @@ fields.incoming[0x03C] = L{
...
 fields.incoming[0x03D] = L{
     {ctype='unsigned int',      label='Price',              fn=gil},            -- 04
     {ctype='unsigned char',     label='Inventory Index',    fn=invp+{0x09}},    -- 08
-    {ctype='unsigned char',     label='Bag',                fn=bag},            -- 09
+    {ctype='unsigned char',     label='Type'},                                  -- 09 0 = on apprasial, 1 = when sale is finalized
     {ctype='unsigned short',    label='_junk1'},                                -- 0A
     {ctype='unsigned int',      label='_unknown1',          const=1},           -- 0C
```
Nifim: `_unknown1` appears to be `Quantity Sold`, can you confirm that and add it to your PR.

addons/libs/packets/fields.lua
```diff
@@ -2132,12 +2132,13 @@ fields.incoming[0x03C] = L{
...
--- Price response
--- Sent after an outgoing price request for an NPC vendor (0x085)
+-- Price/sale response
+-- Sent in response to an outgoing price request for an NPC vendor (0x085), and in response to player finalizing a sale.
+-- Inventory bag is assumed to be regular Inventory (0)
 fields.incoming[0x03D] = L{
     {ctype='unsigned int',      label='Price',              fn=gil},            -- 04
     {ctype='unsigned char',     label='Inventory Index',    fn=invp+{0x09}},    -- 08
-    {ctype='unsigned char',     label='Bag',                fn=bag},            -- 09
+    {ctype='unsigned char',     label='Type'},                                  -- 09 0 = on apprasial, 1 = when sale is finalized
```
Nifim: Can you change `apprasial` to `price check` or something along those lines. 
There is an appraisal system in-game so using that word here is may lead to confusion.

Nifim (changes requested):

addons/libs/packets/fields.lua
Chiaia: It'll be a 1 for a Price Check though.

ibm2431: Confirming that I did check (with crystals of varying quantities) and `_unknown1` is quantity sold or 1 depending on type of response

Nifim (approved):

PR #1761, opened 2019-06-28
Scoreboard update and extra info for help and errors.

mac10688:
Adding chatmodes for ls2. Showing valid chatmodes for //scoreboard help. Making error message a little more helpful for tells.


mac10688:
Using single quotes, removing parentheses and changed the method to print out chatmodes.

PR #1763, opened 2019-07-05
Send addon send mobid to receiving players

mac10688:
Near the bottom of the file the gsub string replacements weren't doing anything. They seem kind of pointless because as far as I can see, that's the receiving end and it seems pointless.


addons/send/send.lua
```diff
@@ -4,10 +4,24 @@ _addon.command = 'send'
 _addon.author = 'Byrth'
 
 windower.register_event('addon command',function (...)
-    local term = table.concat({...}, ' ')
+	local term = table.concat({...}, ' ')
```
z16: Please fix the indentation in the entire file to be 4 spaces. This is not your fault, the file was inconsistent before, I'd just like to ask you to do it since you're modifying it.

addons/send/send.lua
```diff
@@ -4,10 +4,24 @@ _addon.command = 'send'
...
+	
+	local target = windower.ffxi.get_mob_by_target('t')
```
z16: This function (as well as the one below with `lastst`) are expensive. You're calling them on every send command. Instead I'd use `gsub` with a replacement function:
```lua
term = term:gsub('<(%a+)id>', function(target_string)
    local entity = windower.ffxi.get_mob_by_target(target_string)
    return entity and entity.id or '<' .. target_string .. 'id>'
end)
```

This way it'll automatically work with everything that `get_mob_by_target` works with, like `<meid>`, `<htid>`, `<stid>`, and so on. It would, however, mean that  `<lstid>` would not work, it would be `<laststid>` instead, which is something I'd suggest to do anyway to keep consistency.

Then you could also remove the block below.

addons/send/send.lua
```diff
@@ -4,10 +4,24 @@ _addon.command = 'send'
...
+	if last_sub_target then
+		term = term:gsub('<lstid>', tostring(last_sub_target.id))
+	end
+
 	local broken_init = split(term, ' ')
 	local qual = table.remove(broken_init,1)
 	local player = windower.ffxi.get_player()
+
+	
+
```
z16: Why? One empty line I could understand, though I wouldn't like it, but three?

addons/send/readme.txt
```diff
@@ -5,3 +5,17 @@ If you wish to change the color of the text added from aecho. Or any plugin that
 windower.add_to_chat(55,msg:sub(5))
 
 You may change the 55 to any number from 1 to 255 to get a (not always) different color. 
+
+Can use <tid> (stands for target id) from the user originating the send and it will replace the <tid> with the senders currently targeted mob id.
+
+/con send @others /ma "Blizzard IV" <tid>
+
+and another program, e.g. gearswap will resolve the target id to the monster and cast blizzard IV on the mob targeted by the sender.
```
z16: I would mention that GS is strictly required for any ID replacement to work. I'd call it an `addon` instead of a `program` though, just to avoid confusion among non-technical users.

z16 (changes requested):
I left some notes on the commit.

PR #1764, opened 2019-07-07
Improve Summoner CheckParams

Acbarakat:
There was an issue where Blood Pact Damage would from Pet: and Avatar: did not add up together. My solution is to make all Blood Pact Damage to be Pet: Blood Pact Damage.

### DIFF
--- a/addons/checkparam/checkparam.lua
+++ b/addons/checkparam/checkparam.lua
@@ -116,7 +116,7 @@ function get_text(id,data)
         split_text(id,v)
     end
     if stats[2] then
-        stats[2] = stats[2]:gsub('\n', ''):trim()
+        stats[2] = stats[2]:trim()
         split_text(id,stats[2],'pet: ')
     end
     local ext = extdata.decode({id=id,extdata=data})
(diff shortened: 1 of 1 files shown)


addons/checkparam/checkparam.lua
```diff
@@ -122,7 +122,13 @@ function get_text(id,data)
     local ext = extdata.decode({id=id,extdata=data})
     if ext.augments then
         for i,v in ipairs(ext.augments) do
-            split_text(id,v)
+			local stats = windower.regex.split(v,'(Pet|Avatar|Automaton|Wyvern|Luopan): ')
```
z16: Fix the indentation, from hard tabs to four spaces.

addons/checkparam/checkparam.lua
```diff
@@ -122,7 +122,13 @@ function get_text(id,data)
...
+			if stats[2] then
+				local pet_text = windower.regex.replace(stats[2],'\n',' ')
```
z16: You should probably just use `stats[2]:gsub('\n', '')`, while the regex functions are very powerful they're usually somewhat slow if you don't need them. In this case you should also be able to use `stats[2]:trim()`.

z16 (changes requested):

addons/checkparam/checkparam.lua
Acbarakat: Changed and tested with gsub in the new commit

z16:
I may have worded it poorly, but you need either `:gsub` or `:trim` but not both, sorry if I was unclear. `:trim()` removes any leading or trailing spaces (including newlines).

PR #1788, opened 2019-10-08
Adding Mug messages to battlemod

KenshiDRK:
Adding mug messages for simplify


addons/battlemod/parse_action_packet.lua
```diff
@@ -192,7 +193,7 @@ function parse_action_packet(act)
                    act.action = {}
                 else m.simp_name = act.action.name or ''
                 end
-
+
```
z16: Not cool :(

KenshiDRK: D: it always happens :/

PR #1797, opened 2019-10-19
Gearswap: Filter enchanted items with duplicate names.

Ivaar:
This addresses https://github.com/Byrth/Lua-Byrth/issues/455


z16:
I'm not well enough versed in it to understand exactly what happens here, but seeing how you reference an `'english'` sub-table I have to ask, will this work for the Japanese version as well?

Ivaar:
Its just checking if an item name has made a first pass, the function called inside that block adds all the language strings to respective tables.

PR #1822, opened 2020-01-13
New addon - EmpyPopTracker

xurion:
Submitting first addon :o
Source repo: https://github.com/xurion/ffxi-empy-pop-tracker
FFXIAH thread: https://www.ffxiah.com/forum/topic/54376/empyrean-pop-tracker-addon-10-years-late/

Thanks!


addons/EmpyPopTracker/EmpyPopTracker.lua
```diff
@@ -0,0 +1,339 @@
...
+            items.safe2[i].id == id or
+            items.locker[i].id == id or
+            items.sack[i].id == id or
+            items.satchel[i].id == id or
+            items.inventory[i].id == id or
+            items.storage[i].id == id then
+                owned = true
+                break
+        end
+    end
```
z16, line 77: While there's no issue here it would perform better to iterate each bag individually and would be easier to read, plus you can make it immune to breakage due to new bags being added, as it will automatically work when we update LuaCore:
```lua
local owns_item = function(id, items)
    for _, bag in ipairs(bags) do
        for _, item in ipairs(bag) do
            if item.id == id then
                return true
            end
        end
    end

    return false
end
```

addons/EmpyPopTracker/EmpyPopTracker.lua
```diff
@@ -0,0 +1,339 @@
...
+DISCLAIMED. IN NO EVENT SHALL Dean James (Xurion of Bismarck) BE LIABLE FOR ANY
+DIRECT, INDIRECT, INCIDENTAL, SPECIAL, EXEMPLARY, OR CONSEQUENTIAL DAMAGES
+(INCLUDING, BUT NOT LIMITED TO, PROCUREMENT OF SUBSTITUTE GOODS OR SERVICES;
+LOSS OF USE, DATA, OR PROFITS; OR BUSINESS INTERRUPTION) HOWEVER CAUSED AND
+ON ANY THEORY OF LIABILITY, WHETHER IN CONTRACT, STRICT LIABILITY, OR TORT
+(INCLUDING NEGLIGENCE OR OTHERWISE) ARISING IN ANY WAY OUT OF THE USE OF THIS
+SOFTWARE, EVEN IF ADVISED OF THE POSSIBILITY OF SUCH DAMAGE.
+]]
+
+_addon.name = "Empy Pop Tracker"
```
z16: You're mixing single and double quotes. Stick to single quotes for this repo.

addons/EmpyPopTracker/EmpyPopTracker.lua
```diff
@@ -0,0 +1,339 @@
...
+            item_identifier = 'Ж '
+        else
+            resource = res.items[pop.id]
+            owns_pop = owns_item(pop.id, items)
+            in_pool_count = item_treasure_pool_count(pop.id, items.treasure)
+        end
+
+        local pop_name = 'Unknown pop'
+        if resource then
+            pop_name = ucwords(resource.en)
```
z16: Use `.name`.

addons/EmpyPopTracker/EmpyPopTracker.lua
```diff
@@ -0,0 +1,339 @@
...
+        if owns_pop then
+            item_colour = colors.success
+        else
+            item_colour = colors.danger
+        end
+
+        local pool_notification = ''
+        if in_pool_count > 0 then
+            pool_notification = colors.warning .. ' [' .. in_pool_count .. ']' .. colors.close
+        end
```
z16, line 158: Would it make sense to display that even if the count is zero? I'm not sure, up to you.

addons/EmpyPopTracker/EmpyPopTracker.lua
```diff
@@ -0,0 +1,339 @@
...
+        text = text .. "\n" .. get_indent(depth) .. pop.dropped_from.name .. "\n" .. get_indent(depth) .. ' >> ' .. item_colour .. item_identifier .. pop_name .. colors.close .. pool_notification
+        if pop.dropped_from.pops then
+            text = text .. generate_text(pop.dropped_from, key_items, items, depth + 1)
+        end
+    end
+
+    return text
+end
+
+EmpyPopTracker.add_to_chat = function(message)
```
z16: Do you need to define your own function for that? And do you need to check the type? :)

And if you just want to make sure the color is easily changeable you can either make it a local function or define a variable for it. The error checking seems pointless though, you're the only one calling it, any errors you should notice during testing. Besides, even if you don't `error` manually the internal LuaCore call would error just as well.

addons/EmpyPopTracker/EmpyPopTracker.lua
```diff
@@ -0,0 +1,339 @@
...
+        error('add_to_chat requires the message arg to be a string')
+    end
+
+    windower.add_to_chat(8, message)
+end
+
+EmpyPopTracker.generate_info = function(nm, key_items, items)
+    local nm_type = type(nm)
+    if nm_type ~= 'table' then
+        error('generate_info requires the nm arg to be a table, but got ' .. nm_type .. ' instead')
```
z16: Same as before, the error check seems unnecessary.

addons/EmpyPopTracker/EmpyPopTracker.lua
```diff
@@ -0,0 +1,339 @@
...
+    info.text = generate_text(nm, key_items, items, 1)
+
+    return info
+end
+
+function find_nms(query)
+    local matching_nms = {}
+    local lower_query = query:lower()
+    for _, nm in pairs(nm_data) do
+        local result = string.match(nm.name:lower(), '(.*' .. lower_query .. '.*)')
```
z16: I'm not sure if you should do substring matching for user input. It could result in situations where users can't possibly get the right NM to display without also matching others they don't care about. You could give users the option to provide wildcards on their own, since many of our mechanisms use this feature most of our users are already accustomed to it.

addons/EmpyPopTracker/EmpyPopTracker.lua
```diff
@@ -0,0 +1,339 @@
...
+        if result then
+            table.insert(matching_nms, result)
+        end
+    end
+    return matching_nms
+end
+
+windower.register_event("addon command", function(command, ...)
+    if commands[command] then
```
z16, line 206: This will crash the addon if a user just types `//ept` without arguments, since `command` will be `nil`. A common convention is to define the command handler like this:
```lua

    -- command checking here
end)
```

This way you protect against invalid usage and default to `help` implicitly. So in your case you could just do `(commands[command] or commands.help)(...)` at that point.

addons/EmpyPopTracker/EmpyPopTracker.lua
```diff
@@ -0,0 +1,339 @@
...
+        commands.show()
+    end
+end
+commands.t = commands.track
+
+commands.hide = function()
+    active = false
+    EmpyPopTracker.text:visible(false)
+    EmpyPopTracker.settings.visible = false
+    EmpyPopTracker.settings:save()
```
z16, line 241: You should not need to modify the settings at all when doing this. It should adjust them and save accordingly. Does that not happen for you? Same in the function below.

z16 (changes requested):
Left most of my notes in the addon, but two more ideas for the various NM files. First, do you need to specify the BG link? Can it not always be inferred from the name?

And second, instead of deep nesting I think something like a relational database would be better. Just define the NMs and pops in a single flat file:
```lua
-- nms.lua
local make = function(type, id)
    return {
        type = type,
        id = id,
    }
end

local item = make:apply('item')
local key_item = make:apply('key_item')

return {
    -- Big NMs, don't drop anything and usually just one KI pop
    -- Though I guess you could also note down regular drops here
    -- That would allow people to search by final drop or Atma they want
    { name = 'Alfard', zone = 254, pos = 'J-8', pops = { key_item(1530) }, drops = { } },
    { name = 'Apademak', zone = 253, pos = 'F-6', pops = { key_item(1525) }, drops = { } },
    { name = 'Azdaja', zone = 254, pos = 'C-8', pops = { key_item() }, drops = { } },
    -- Multiple item pops and KI drop
    { name = 'Ningishzida', zone = 254, pos = 'I-7', pops = { item(3261), item(3262), item(3268) }, drops = { key_item(1530) } },
    -- Multiple KI pops
    { name = 'Sedna', zone = 217, pos = 'K-6', pops = { key_item(1511), key_item(1512) } },
    -- And so on
}
```

After this you can resolve these in your addon to create an easy lookup table, essentially generate the `dropped_from` nested tables. Would make editing much easier and present the data in a format that can easily be consumed by other addons, if they wanted. You can make a similar file for items and key items, though you should be able to easily look those up from resources.

You can even generate this file automatically based on your current data with some transformations. The `Forced` tag I think can be removed, if a NM has pops it is always forced, otherwise always timed, or am I missing something?

Overall it looks good as it is, these are just some ideas that you could improve on, even in future versions, but I'd like you to look over a few of the things I mentioned and report back on some of my open questions.

addons/EmpyPopTracker/EmpyPopTracker.lua
z16: Thinking about it, in your case you could just do:
```lua
    (command and commands[command:lower()] or commands.help)(...)
end)
```

Though not sure how readable that still is :D

addons/EmpyPopTracker/EmpyPopTracker.lua
xurion: I checked the items and key_items resources structure and the `name` attribute does not exist - is this added by Windower somehow? I can't check this myself until later tonight.

addons/EmpyPopTracker/EmpyPopTracker.lua
xurion: I originally had this showing for each item. I decided to only show it when there's one in the pool as it was difficult to identify them at a glance, plus it acts as a kind of alert due to the amber colour.

addons/EmpyPopTracker/EmpyPopTracker.lua
z16: It is, and it evaluates to different languages, depending on your client. That will make it usable by JP players as well. We really didn't document it well, but it was intended to use `.name` in most places and `.english` and `.japanese` where you need to be language-specific. I did not intend for `.en` and `.ja` to be used to at all. Unfortunately it's what people see when opening the files. We did document it [here](https://github.com/Windower/Lua/wiki/Resources-Library), but people don't seem to read that :|

addons/EmpyPopTracker/EmpyPopTracker.lua
xurion: Yeah you're right I'll take this out and variable the colour code instead. This project started out as a proof of concept for using a full test-driven and mocked approach. It was fun for a while but it ended up being too much to deal with. A few patterns still remain due to this :o

xurion:
Thanks for the review! I've fixed a bunch of things in the original repo and I'll migrate them over to the fork soon.

The `Forced` label is just part of the name - I removed this but then realised it caused confusion between regular mobs and NMs, such as Dusk Lizards vs Sirrush, so I backed it out.

I like the idea of migrating over to a relational database instead of the nested structure. The main reason I went nested was for a (mentally) predictable way to get the indentation working in the UI. I would like to tackle this as part of a follow-up PR in the near future :D

I have removed the BG URL from each data file - you're right, this is easily inferred by name.

I'll reply to each of the other questions/comments individually.

addons/EmpyPopTracker/EmpyPopTracker.lua
xurion: Done. Future proof ftw

addons/EmpyPopTracker/EmpyPopTracker.lua
xurion: Done.

addons/EmpyPopTracker/EmpyPopTracker.lua
xurion: Done.

addons/EmpyPopTracker/EmpyPopTracker.lua
xurion: Done. I put the chat mode into settings so it is configurable also.

addons/EmpyPopTracker/EmpyPopTracker.lua
xurion: Done.

addons/EmpyPopTracker/EmpyPopTracker.lua
xurion: Isn't creating unreadable code great for job security? XD

I ran `//ept` and it doesn't crash. Running `if commands[nil] then` just evaluates falsey. Happy to change this to any of the above if you prefer that style however :)

addons/EmpyPopTracker/EmpyPopTracker.lua
xurion: I originally expected this to just work like some of my other addons that use texts and settings. I just checked and it doesn't save it automatically :/

I had just assumed the texts visible setting wasn't saved like the others?

addons/EmpyPopTracker/EmpyPopTracker.lua
xurion: My intention was for this to function the same as findall. I agree this does run that risk, but only if the list of NMs is extended to where the NMs are very similarly named. Maybe when they give us Neo-Abyssea heh...heh...he.....uuurrgghhhhhhh~

Do you have an example of where we use wildcards so I can take a peek?

addons/EmpyPopTracker/EmpyPopTracker.lua
z16: Hm, looks like I derped here. It does indeed return `nil`. Setting the `nil` key is an error, but getting it apparently "works" (for certain definitions of "work"). So nevermind that. But, while not critical, your code would not work with cased commands. I personally actually prefer that, but our convention so far to accept it. I'll let you decide how to handle that.

addons/EmpyPopTracker/EmpyPopTracker.lua
z16: Hm you are right, I just checked and visibility is not automatically saved. I wonder why I did it like that... I guess I did not consider visibility as a user based setting and instead wanted the addon to control that behavior. An argument I no longer agree with, but I think that's what influenced my decision back then.

addons/EmpyPopTracker/EmpyPopTracker.lua
z16: Yet another reason to hate `findall` :(

Treasury for example works with exact matches, but allows using wildcards to add/remove items from its various lists.

z16 (changes requested):
Why on earth did you adjust the indent size? <_< Back to 4 spaces, please... otherwise it looks good, as before, see the two or three new comments I made, but once the indent size is fixed it can be merged.

addons/EmpyPopTracker/EmpyPopTracker.lua
xurion: Hehe no worries - I'll have a look using a wildcard approach.

xurion:
...I have no idea - for some reason I thought indents were 2 not 4 :/ I'll revert these and address the other comments later today. Thanks again!

addons/EmpyPopTracker/EmpyPopTracker.lua
xurion: Implemented wildcard approach. Thanks for the help.

addons/EmpyPopTracker/EmpyPopTracker.lua
xurion: I've made it consistent with other addons now. I've simplified the first few lines of the addon command event.

PR #1826, opened 2020-01-28
Azure sets update to allow for subjob use

Gol-exe:
Updated the addon AzureSets to allow for use while subbing blue mage. This involved refactoring, adding a number of checks to ensure the user is on the correct job, has enough blue magic points to spend, has an open spell slot, and is of the correct level to set that spell.

### DIFF
--- a/addons/azureSets/azuresets.lua
+++ b/addons/azureSets/azuresets.lua
@@ -14,7 +14,7 @@ documentation and/or other materials provided with the distribution.
 names of its contributors may be used to endorse or promote products
 derived from this software without specific prior written permission.
 
-THIS SOFTWARE IS PROVIDED BY THE COPYRIGHT HOLDERS AND CONTRIBUTORS "AS IS" AND
+THIS SOFTWARE IS PROVIDED BY THE COPYRIGHT HOLDERS AND CONTRIBUTORS 'AS IS' AND
 ANY EXPRESS OR IMPLIED WARRANTIES, INCLUDING, BUT NOT LIMITED TO, THE IMPLIED
 WARRANTIES OF MERCHANTABILITY AND FITNESS FOR A PARTICULAR PURPOSE ARE
 DISCLAIMED. IN NO EVENT SHALL The Addon's Contributors BE LIABLE FOR ANY
@@ -25,7 +25,6 @@ ON ANY THEORY OF LIABILITY, WHETHER IN CONTRACT, STRICT LIABILITY, OR TORT
 (INCLUDING NEGLIGENCE OR OTHERWISE) ARISING IN ANY WAY OUT OF THE USE OF THIS
 SOFTWARE, EVEN IF ADVISED OF THE POSSIBILITY OF SUCH DAMAGE.
 ]]
-
 
 _addon.name = 'AzureSets'
 _addon.version = '1.25'
@@ -38,7 +37,6 @@ require('logger')
 local config = require('config')
 local res = require('resources')
 local chat = require('chat')
-local packets = require('packets')
 local spells = res.spells:type('BlueMagic')
 
 local defaults = {}
@@ -66,39 +64,25 @@ local bluJobLevel = nil
 local bluPointsMax = nil
 local bluSlots = nil
 local get_blu_job_data = nil
+local language = windower.ffxi.get_info().language:lower()
 
 local spellsLookup = {}
 for spell in spells:it() do
-    spellsLookup[spell.english] = spell
-    spellsLookup[spell.english:lower()] = spell
-    spellsLookup[spell.japanese] = spell
+    spellsLookup[spell[language]:lower()] = spell
 end
 
 function initialize()
-    update_blu_info()
-    update_current_spellset()
+    local player = windower.ffxi.get_player()
+    local is_blu = player and (player.main_job_id == BLU_JOB_ID or player.sub_job_id == BLU_JOB_ID)
+    local logged_in = windower.ffxi.get_info()
+
+    if is_blu and logged_in then
+        update_blu_info()
+        update_current_spellset()
+    end
 end
 
-windower.register_event('load', function()
-    local player = windower.ffxi.get_player()
-    local is_blu = player and player.main_job_id == BLU_JOB_ID
-    local logged_in = windower.ffxi.get_info()
-    if is_blu and logged_in then
-        initialize()
-    end
-end)
-
-windower.register_event('incoming chunk', function(id, original, modified, injected, blocked) 
-    if id ~= 0x00A then
-        return
-    end
-    local packet = packets.parse('incoming', original)
-    if packet['Main Job'] == BLU_JOB_ID then
-       initialize()
-    end 
-end)
-
-windower.register_event('job change', initialize)
+windower.register_event('load', 'login', 'job change', initialize)
 
 function update_blu_info(player)
     player = player or windower.ffxi.get_player()
@@ -212,7 +196,7 @@ function set_spells_from_spellset(spellset, setPhase)
 end
 
 function find_spell_id_by_name(spellname)
-    local spell = spellsLookup[spellname] or spellsLookup[spellname:lower()]
+    local spell = spellsLookup[spellname:lower()]
     if spell and spell.id then
         return spell.id
     end
@@ -274,14 +258,14 @@ function delete_set(setname)
 end
 
 function get_spellset_list()
-    log("Listing sets:")
+    log('Listing sets:')
     for key,_ in pairs(settings.spellsets) do
         if key ~= 'default' then
             local it = 0
             for i = 1, #settings.spellsets[key] do
                 it = it + 1
             end
-            log("\t"..key..' '..settings.spellsets[key]:length()..' spells.')
+            log('\t'..key..' '..settings.spellsets[key]:length()..' spells.')
         end
     end
 end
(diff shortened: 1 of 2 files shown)


addons/azureSets/azuresets.lua
```diff
@@ -57,22 +57,62 @@ slot09='Blank Gaze', slot10='Radiant Breath', slot11='Light of Penance', slot12=
...
-    spells = res.spells:type('BlueMagic')
-    get_current_spellset()
+    update_blu_info()
+    update_current_spellset()
 end
 
 windower.register_event('load', initialize:cond(function() return windower.ffxi.get_info().logged_in end))
 
-windower.register_event('login', initialize)
+windower.register_event('login', function() coroutine.schedule(initialize, 5) end)
```
z16: I'm not a fan of timed delays. I assume this is to wait for data to be populated? If so, I'd make it dependent on a packet or data being available in memory. Same goes for the load event, which currently only checks for being logged in.

addons/azureSets/azuresets.lua
```diff
@@ -122,22 +161,23 @@ function set_spells_from_spellset(spellset, setPhase)
...
         end
     end
 
     if slotToSetTo ~= nil then
         -- We found an empty slot. Find a spell to set.
-        for k,v in pairs(setToSet) do
-            if not currentSet:contains(v:lower()) then
+        for k, v in pairs(setToSet) do
+            if not currentSpellSet:contains(v:lower()) then
                 if v ~= nil then
```
z16: This check can be removed. I know it was there before, but still :) It does nothing.

addons/azureSets/azuresets.lua
```diff
@@ -238,21 +281,66 @@ function get_spellset_content(spellset)
...
+    end
+    if bluJobLevel and spell.levels and spell.levels[BLU_JOB_ID] and spell.levels[BLU_JOB_ID] > bluJobLevel then
+        errorMessage = "job level too low to set spell"
+    end
+    if not have_enough_points_to_add_spell(id) then
+        errorMessage = "cannot set spell, ran out of blue magic points"
+    end
+    if slot > bluSlots then
+        errorMessage = "slot " .. tostring(slot) .. " unavailable"
+    end
```
z16: The four `if`s here all get executed, but the first one should error and return. As it is now if `spell` is `nil` the first `if` will be true and the message will be set to `spell not found`, but then it will go to the next `if` where it tries to do `spell.levels` and that will error. Either change them to `elseif` or return early from them, which I would prefer:
```lua
if not spell then
    error('spell not found')
    return false
end
-- next
```
This saves you the `errorMessage` variable as well as the last whole block at the end.

addons/azureSets/azuresets.lua
```diff
@@ -159,39 +199,42 @@ function find_spell_id_by_name(spellname)
     return nil
```
z16: This function is not your doing, but since you're adjusting the addon I'll leave this here anyway. For one, you should cache the lowercased spell name. And also, this could be `.name` instead of `['english']`, to make it work on the JP client as well. You'd need to eliminate the `:lower()` call though... or you do:
```lua
if spell.english:lower()  == lower_spellname or spell.japanese == spellname then
```

I think the best approach would be to create a lookup table for all spell names at load:
```lua
local lookup = {}
for spell in spells:it() do
    lookup[spell.english] = spell
    lookup[spell.english:lower()] = spell
    lookup[spell.japanese] = spell
end
```

Then you could just replace the entire functionwith:
```lua
function find_spell_id_by_name(spellname)
    return lookup[spellname] or lookup[spellname:lower()]
end
```

addons/azureSets/azuresets.lua
```diff
@@ -159,39 +199,42 @@ function find_spell_id_by_name(spellname)
...
-function get_current_spellset()
-    if windower.ffxi.get_player().main_job_id ~= 16 then return nil end
-    return T(windower.ffxi.get_mjob_data().spells)
+function update_current_spellset(player)
+    if not get_blu_job_data then
+        currentSpellSet = nil
+        return nil
+    end
+
+    currentSpellSet = T(get_blu_job_data().spells)
```
z16, line 231: This function is called in a loop in `set_spells_from_spellset`. I would call it once on the original call-site and pass it as an argument.

addons/azureSets/azuresets.lua
```diff
@@ -159,39 +199,42 @@ function find_spell_id_by_name(spellname)
     return nil
 end
 
-function set_single_spell(setspell,slot)
-    if windower.ffxi.get_player()['main_job_id'] ~= 16 --[[and windower.ffxi.get_player()['sub_job_id'] ~= 16]] then return nil end
+function set_single_spell(setspell, slot)
+    if not bluJobLevel then return nil end
```
z16: A bit inconsistent that you sometimes check for `bluJobLevel` and sometimes for `get_blu_job_data`, when I think you could drop both. You already check for that at the top of the `addon command` event. I think in all subsequent calls you can omit the check.

addons/azureSets/azuresets.lua
```diff
@@ -159,39 +199,42 @@ function find_spell_id_by_name(spellname)
...
         end
     end
-    tmpTable = nil
 end
 
-function get_current_spellset()
-    if windower.ffxi.get_player().main_job_id ~= 16 then return nil end
-    return T(windower.ffxi.get_mjob_data().spells)
+function update_current_spellset(player)
+    if not get_blu_job_data then
```
z16: As I said above, you should be able to just remove this check.

addons/azureSets/azuresets.lua
```diff
@@ -159,39 +199,42 @@ function find_spell_id_by_name(spellname)
...
     end
     if tonumber(slot) < 10 then slot = '0'..slot end
     --insert spell add code here
-    for spell in spells:it() do
-        if spell['english']:lower() == setspell then
-            --This is where single spell setting code goes.
-            --Need to set by spell id rather than name.
-            windower.ffxi.set_blue_magic_spell(spell['id'], tonumber(slot))
+    local spellId = find_spell_id_by_name(setspell)
+    if spellId then
```
z16: You don't need this check, since you call `verify_and_set_spell` in the next step and that already checks for the validity of the spell ID.

addons/azureSets/azuresets.lua
```diff
@@ -122,22 +161,23 @@ function set_spells_from_spellset(spellset, setPhase)
...
 
     if slotToSetTo ~= nil then
         -- We found an empty slot. Find a spell to set.
-        for k,v in pairs(setToSet) do
-            if not currentSet:contains(v:lower()) then
+        for k, v in pairs(setToSet) do
+            if not currentSpellSet:contains(v:lower()) then
                 if v ~= nil then
                     local spellID = find_spell_id_by_name(v)
                     if spellID ~= nil then
```
z16: Same as before, you don't need the `spellID` check here, since it gets verified in `verify_and_set_spell` afterwards. In fact, if you take my previous suggestion about changing `find_spell_id_by_name` I would just integrate it into `verify_and_set_spell`, i.e. change `verify_and_set_spell` to take a string and slot, and search for the ID in there. That would also fit the description of the function and remove redundancy.

z16 (commented):
Left some notes regarding the changes, let me know what you think.

addons/azureSets/azuresets.lua
Gol-exe: I set it up that way to ensure that the most recent spellset was always the one currently loaded. If the command to set the spell fails then it will attempt to do it again.

Gol-exe (commented):
Updated the addon per recommendations- left one comment regarding updating the spell set.

addons/azureSets/azuresets.lua
```diff
@@ -66,14 +67,36 @@ local bluPointsMax = nil
 local bluSlots = nil
 local get_blu_job_data = nil
 
+local spellsLookup = {}
+for spell in spells:it() do
+    spellsLookup[spell.english] = spell
+    spellsLookup[spell.english:lower()] = spell
+    spellsLookup[spell.japanese] = spell
```
z16, line 75: You can just do:
```lua
local language = windower.ffxi.get_info().language
local spellsLookup 
for spell in spells:it() do
    spellsLookup[spell[language]:lower()] = spell
end
```
Then later always access it with `:lower()`. See below.

addons/azureSets/azuresets.lua
```diff
@@ -66,14 +67,36 @@ local bluPointsMax = nil
...
-windower.register_event('login', function() coroutine.schedule(initialize, 5) end)
+windower.register_event('incoming chunk', function(id, original, modified, injected, blocked) 
+    if id ~= 0x00A then
+        return
+    end
+    local packet = packets.parse('incoming', original)
+    if packet['Main Job'] == BLU_JOB_ID then
+       initialize()
+    end 
+end)
```
z16, line 87: I don't see why these were changed, but you could make it much easier by putting the sanity logic into the `initialize` function. Then you can reduce all events to this line:
```lua
windower.register_event('load', 'login', 'job_change', 'zone_change')
```

Though really, I don't see why you do this on zone change at all. If that `0x00A` thing was meant to be login, then use just the `login` event, not `incoming chunk` and filter for `0x00A`.

addons/azureSets/azuresets.lua
```diff
@@ -283,22 +293,20 @@ end
...
+        error("cannot set spell, ran out of blue magic points")
+        return false
     end
     if slot > bluSlots then
-        errorMessage = "slot " .. tostring(slot) .. " unavailable"
-    end
-
-    if errorMessage then
-        error(errorMessage)
+        error("slot " .. tostring(slot) .. " unavailable")
```
z16: I saw this was there before, but since you're changing this code, please replace the double quotes with single quotes.

addons/azureSets/azuresets.lua
```diff
@@ -191,17 +212,13 @@ function set_spells_from_spellset(spellset, setPhase)
 end
 
 function find_spell_id_by_name(spellname)
-    for spell in spells:it() do
-        if spell['english']:lower() == spellname:lower() then
-            return spell['id']
-        end
+    local spell = spellsLookup[spellname] or spellsLookup[spellname:lower()]
```
z16: If you always use `:lower()` as I did above you can drop the first check. The entire function could be changed to this:
```lua
function find_spell_id_by_name(name)
    local spell = spellsLookup[name:lower()]
    return spell and spell.id
end
```

RubenatorX:
I just want to document that the pre-refactor code seems to work.
https://github.com/Windower/Lua/blob/c03eee3e51708ca96f9fdd38ef1d332ea4622493/addons/azureSets/azuresets.lua
After further investigation the refactor stuff submitted is bugged somehow but I don't actually play blue mage so its difficult for me to test/debug it, but it should be close to working

PR #1827, opened 2020-02-11
Expand MyHome

Chiaia:


Chiaia:
I should of used ipc message over send.

PR #1834, opened 2020-02-24
Added WoC and ADL to trackable NMs

xurion:
* Adjusted addon to allow tracking of NMs that use regular items as pops.
* Allowed tracking of NMs with spaces in their name.

### DIFF
--- a/addons/EmpyPopTracker/EmpyPopTracker.lua
+++ b/addons/EmpyPopTracker/EmpyPopTracker.lua
@@ -166,30 +166,12 @@ function generate_text(data, key_items, items, depth)
 end
 
 EmpyPopTracker.generate_info = function(nm, key_items, items)
-    local info = {
-        has_all_pops = true,
-        text = ''
+    return {
+        has_all_pops = not nm.pops or T(nm.pops):all(function(item)
+            return item.type == 'item' and owns_item(item.id, items) or owns_key_item(item.id, key_items)
+        end),
+        text = generate_text(nm, key_items, items, 1)
     }
-
-    if nm.pops then
-        for _, pop_item_data in pairs(nm.pops) do
-            local has_pop = false
-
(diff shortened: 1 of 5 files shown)


addons/EmpyPopTracker/EmpyPopTracker.lua
```diff
@@ -167,16 +167,22 @@ end
...
-            if not has_pop_ki then
-                info.has_all_kis = false
+            if pop_item_data.type == 'key item' then
+                has_pop = owns_key_item(pop_item_data.id, key_items)
+            elseif pop_item_data.type == 'item' then
+                has_pop = owns_item(pop_item_data.id, items)
+            end
+
+            if not has_pop then
+                info.has_all_pops = false
```
z16: The entire function body can be replaced with:
```lua
return {
    has_all_pops = not nm.pops or nm.pops:all(function(item)
        return pop.type == 'item' and owns_item(pop.id, items) or owns_key_item(pop.id, key_items)
    end)
    text = generate_text(nm, key_items, items, 1)
}
```

If it complains about `all` being `nil` then you need to do `T(nm.pops):all`.

Adjust as you see fit for readability.

z16 (approved):
Minor remark, but looks good. Let me know if you wanna adjust it somehow or if you want me to merge it as it is.

addons/EmpyPopTracker/EmpyPopTracker.lua
xurion: Nice, I like it and I have just pushed the change. Thanks for the speedy review!

RubenatorX:
Is a name change in order? :X WoC isn't exactly an EmpyPop xD

z16:
Eh :D

xurion:
I didn't even want to contemplate what that would mean for backwards compatibility XD In hindsight, I was too quick to choose the name

PR #1850, opened 2020-03-13
Dtw boxdestroyer

dtw:
I found it hard to pick out the recommended action in the chat log so I colored it yellow.

### DIFF
--- a/addons/boxdestroyer/boxdestroyer.lua
+++ b/addons/boxdestroyer/boxdestroyer.lua
@@ -67,6 +67,10 @@ observed_default = {
 
 thief_tools = {[1022] = true}
 
+-- color codes for ease of use
+txt = {}
+txt[36] = string.char(31,36)  -- yellow
+
 -- global variables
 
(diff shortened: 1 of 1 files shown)


z16:
I like the idea, but unsure about the implementation. First, why a table, not just use a single variable? Also, I would make it user configurable via the `config` lib and *disable* it by default. Per convention we use a color spectrum from 1 through 510. The coloring I would do by requiring the `chat`. Overall, something like this:
```lua
require('chat')
config = require('config')

defaults = {
    HighlightResult: false,
    HighlightColor: 36,
}

-- later
local result = clue_value > guess_value and remaining ~= 1 and 'examining the chest' or 'guessing ' .. '%d':format(box[id][math.ceil(#box[id] / 2)])
local formatted_result = settings.HighlightResult and result:color(settings.HighlightColor) or result
windower.add_to_chat(207, 'boxdestroyer recommends ' .. formatted_result .. '.')
```

PR #1860, opened 2020-04-10
Salvage2 Update

tehkrizz:
Updated Salvage2 finally. I used TH Tracker as a baseline since it works. Salvage2 now uses Texts instead of the previous method. for the textboxes, and the settings are structured different. Not sure if it made a difference, but I also adjusted the chat message checks to the format I saw being used. Overall code cleanup as well with the implementation of Texts.

I've done the same thing with DynamisHelper, but it requires a little more testing to ensure it is fully functional again.


addons/salvage2/salvage2.lua
```diff
@@ -75,9 +77,9 @@ local params = {...};
 				windower.text.set_location('salvage_box2', posx, posy)
 			end
 		elseif params[1]:lower() == "hide" then
-			windower.text.set_visibility('salvage_box2', false)
+			salvage_box2:hide()
```
z16: Tabs! They were here before, too, but still, please change all tabs to 4 spaces.

addons/salvage2/salvage2.lua
```diff
@@ -109,29 +111,37 @@ windower.register_event('login', function(name)
 	player = name
 end)
 
-salvage_zones = S{73, 74, 75, 76}
+function settings_create()
+	--	get player's name
```
z16: Tabs!

addons/salvage2/salvage2.lua
```diff
@@ -109,29 +111,37 @@ windower.register_event('login', function(name)
...
+	end
+end
 
 windower.register_event('zone change', function(id)
 	if salvage_zones:contains(id) then
 		windower.send_command('timers c Remaining 6000 up')
 		settings_create()
 		initialize()
-		windower.text.set_visibility('salvage_box2', true)
+		salvage_box2:show()
```
z16: Tabs!

addons/salvage2/salvage2.lua
```diff
@@ -109,29 +111,37 @@ windower.register_event('login', function(name)
...
-		windower.text.set_visibility('salvage_box2', false)
+		salvage_box2:hide()
 	end
 end)
 
 windower.register_event('incoming text',function (original, new, color)
-
-	a,b,pathos,name = string.find(original,'..(.*) removed for (%w+)\46')
-
+	original = original:strip_format()
```
z16: Tabs!

addons/salvage2/data/settings.xml
```diff
@@ -1,7 +1,50 @@
...
+            <stroke>
+                <alpha>255</alpha>
+                <blue>0</blue>
+                <green>0</green>
+                <red>0</red>
+                <width>0</width>
+            </stroke>
+        </text>
+    </global>
+</settings>
```
z16: It is very unfortunately that this file was ever checked in. That should not have happened. Please delete it from source control. It won't be updated for users anyway, everything in the `/data` subfolder for addons is exempt from updates.

z16 (changes requested):
Hope you appreciate my constructive and detailed code review!

tehkrizz:
@z16 , your feedback is always welcome. Especially when I haven't touched the addons in nearly 2 years. I removed settings.xml from my source and pushed that. I've made the tabs change, but I won't push that one until after my next Salvage run to make sure I didn't screw up a simple find/replace.

addons/salvage2/salvage2.lua
```diff
@@ -30,142 +30,146 @@ require('tables')
 require('strings')
 require('maths')
 require('logger')
+texts = require 'texts'
```
z16: And if you wanna do me another favor, add some parens around this :)

z16 (approved):

addons/salvage2/salvage2.lua
tehkrizz: I just copy/pasted it from THTracker. I've updated it in both, and I'll do another push for THTracker later.

tehkrizz:
Committed the last of the changes. I included the THTracker parenthesis and tabs as well.

PR #1862, opened 2020-04-18
New addon - NoCampaignMusic

xurion:
Source repo: https://github.com/xurion/ffxi-no-campaign-music

Thanks!

### DIFF
--- a/addons/NoCampaignMusic/NoCampaignMusic.lua
+++ b/addons/NoCampaignMusic/NoCampaignMusic.lua
@@ -35,8 +35,7 @@ packets = require('packets')
 config = require('config')
 
 defaults = {
-    active = true,
-    notifications = false,
+    Notifications = false,
 }
 
 settings = config.load(defaults)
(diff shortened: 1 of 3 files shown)


addons/NoCampaignMusic/NoCampaignMusic.lua
```diff
@@ -0,0 +1,162 @@
...
+        if parsed['Day Music'] == campaign_id and zone_music_map[parsed.Zone] then
+            campaign_active = true
+            if not settings.active then return end
+
+            parsed['Day Music'] = zone_music_map[parsed.Zone][1]
+            parsed['Night Music'] = zone_music_map[parsed.Zone][1]
+            parsed['Solo Combat Music'] = zone_music_map[parsed.Zone][2]
+            parsed['Party Combat Music'] = zone_music_map[parsed.Zone][3]
+
+            windower.add_to_chat(8, 'Prevented campaign music.')
```
z16: I would remove this, or at the very least make a setting to disable it (and preferably default to `false`).

addons/NoCampaignMusic/NoCampaignMusic.lua
```diff
@@ -0,0 +1,162 @@
...
+
+settings = config.load(defaults)
+
+campaign_id = 247
+solo_id = 101
+party_id = 215
+solo_dungeon_id = 115
+party_dungeon_id = 216
+
+campaign_active = false
```
z16, line 49: This is initialized with `false`, and to `true` when campaign starts, but it's never reset to `false` again when campaign ends. So if you do `//ncm off` the check will always be `true` and it'll reset to campaign music, even if it is not running.

addons/NoCampaignMusic/NoCampaignMusic.lua
```diff
@@ -0,0 +1,162 @@
...
+        local parsed = packets.parse('incoming', data)
+        if parsed['Song ID'] == campaign_id then
+            campaign_active = true
+            if not settings.active then return end
+
+            local info = windower.ffxi.get_info()
+            if not zone_music_map[info.zone] then return end
+
+            if parsed['BGM Type'] == 0 then --only log to the chat once
+                windower.add_to_chat(8, 'Prevented campaign music.')
```
z16, line 102: See above, I feel this is something users will not need to know and rather it happened in the background.

z16 (commented):
Left some comments. Overall it looks good and simple enough, but would be curious about your thoughts on it.

addons/NoCampaignMusic/NoCampaignMusic.lua
xurion: Agreed - I removed one of the two `add_to_chat` calls and put the remaining one behind a settings flag defaulted to false.

addons/NoCampaignMusic/NoCampaignMusic.lua
xurion: Good catch. Lemme know what you think of the `elseif` I added in latest commit to solve this.

addons/NoCampaignMusic/NoCampaignMusic.lua
xurion: As above, this is now behind a settings flag.

addons/NoCampaignMusic/NoCampaignMusic.lua
```diff
@@ -0,0 +1,168 @@
...
+_addon.author = 'Dean James (Xurion of Bismarck)'
+_addon.version = '2.0.0'
+_addon.commands = {'nocampaignmusic', 'ncm'}
+
+packets = require('packets')
+config = require('config')
+
+defaults = {
+    active = true,
+    notifications = false,
```
z16: Please use `CamelCase` for user settings.

addons/NoCampaignMusic/NoCampaignMusic.lua
```diff
@@ -0,0 +1,168 @@
...
+zone_music_map[175] = { 0, 0, solo_dungeon_id, party_dungeon_id } --The Eldieme Necropolis [S]
+
+windower.register_event('incoming chunk', function(id, data)
+    if id == 0x00A then --Zone update (zoned in)
+        local parsed = packets.parse('incoming', data)
+        if parsed['Day Music'] == campaign_id and zone_music_map[parsed.Zone] then
+            campaign_active = true
+            if not settings.active then return end
+
+            parsed['Day Music'] = zone_music_map[parsed.Zone][1]
```
z16: I would cache the lookup:
```lua
local zone_music = zone_music_map[parsed.Zone]
```

Would make it a bit easier to read, and even improve performance ever so slightly :) And the same in the lower packet. In fact, could remove a bit of code duplication if you did this:
```lua
if id ~= 0x00A and id ~= 0x05F then
    return
end

local parsed = packets.parse('incoming', data)
local zone_music = zone_music_map[parsed['Zone'] or windower.ffxi.get_info().zone]

if id == 0x00A then
    -- Zone in packet logic
else
    -- Music change packet logic
end
```

But some people would probably find it a bit over-engineered, so I'll leave it up to you.

z16 (commented):
Left a few more comments, all in all it looks good, I'd adjust the user comments to use `CamelCase`, to give a more unified experience with other addons.

Now there is one thing I'm still unsure about, and that's the `on/off` switch. I'm personally a fan of loading/unloading an addon to enable/disable its core functionality. I always feel weird when I load an addon but it still doesn't work because I forgot to turn it on. The load/unload mechanism is intended to be the enable/disable mechanism, and it's one our users are already familiar with and know the syntax for, so it feels like an unnecessary layer added on top of it. I'll leave this decision up to you, just something that always bothers me personally.

addons/NoCampaignMusic/NoCampaignMusic.lua
xurion: Done.

addons/NoCampaignMusic/NoCampaignMusic.lua
xurion: I like it! Thanks for taking the time to write it 👌

xurion:
> I'm personally a fan of loading/unloading an addon to enable/disable its core functionality.

Totally agree and removed the on/off functions.

PR #1871, opened 2020-04-30
JobChange: Added command to change both jobs at once and prefer dw subjobs on reset

AkadenTK:


addons/JobChange/jobchange.lua
```diff
@@ -27,119 +27,181 @@ SOFTWARE, EVEN IF ADVISED OF THE POSSIBILITY OF SUCH DAMAGE.
...
+    local args = L{...}
+    local job = ''
+    if args[1] then 
+        job = args[1]:lower()
+    end
+    local main = nil
+    local sub = nil
+    if command:lower() == 'main' then
+        main = job
+        if main and main:upper() == p.main_job then main = nil end
```
z16, line 114: `main` here must always exist (so does `sub` below), can be removed from the check. Also, inline `if`, ew :<

addons/JobChange/jobchange.lua
```diff
@@ -27,119 +27,181 @@ SOFTWARE, EVEN IF ADVISED OF THE POSSIBILITY OF SUCH DAMAGE.
...
+windower.register_event('addon command', function(command, ...)
+    local p = windower.ffxi.get_player()
+    local args = L{...}
+    local job = ''
+    if args[1] then 
+        job = args[1]:lower()
+    end
+    local main = nil
+    local sub = nil
+    if command:lower() == 'main' then
```
z16, line 112: `command` should be lowercased at the start, so it does not need to be done in every check. I would also uppercase the job strings, so it does not need to be performed everywhere below. Would make the code more readable imo, but up to you.

addons/JobChange/jobchange.lua
```diff
@@ -27,119 +27,181 @@ SOFTWARE, EVEN IF ADVISED OF THE POSSIBILITY OF SUCH DAMAGE.
...
+        main = job
+        if main and main:upper() == p.main_job then main = nil end
+    elseif command:lower() == 'sub' then
+        sub = job
+        if sub and sub:upper() == p.sub_job then main = nil end
+    elseif command:lower() == 'reset' then
+        log('Resetting Job')
+        sub = p.sub_job:lower()
+    elseif command:contains('/') or command:contains('\\') then
```
z16, line 121: You allow backslashes for job delimiters? Why?

addons/JobChange/jobchange.lua
```diff
@@ -27,119 +27,181 @@ SOFTWARE, EVEN IF ADVISED OF THE POSSIBILITY OF SUCH DAMAGE.
...
+                log('Conflict with '..(change.is_main and 'main' or 'sub')..' job. Changing to: '..res.jobs[change.job_id].ens)
+            else
+                log('Changing '..(change.is_main and 'main' or 'sub')..' job to: '..res.jobs[change.job_id].ens)
+            end
+            jobchange(change.job_id, change.is_main)
+
+            coroutine.sleep(0.5)
+        end
+    else
+        log('Not close enough to a Moogle!')
```
z16, line 205: This case could be checked for at the very start to save some otherwise unnecessary calculations.

z16 (approved):
Just left a few comments, looks fine though. Let me know if you want to adjust some of them or to pull it as is.

addons/JobChange/jobchange.lua
AkadenTK: Why not?

AkadenTK:
I agree with all the comments, but don't have an active sub to test properly (even though they're very basic changes). I've been using it pretty consistently though until I unsubbed. I'd be okay with it as is.

addons/JobChange/jobchange.lua
z16: Added complexity, lower performance, more room for bugs, and no one uses it :)

PR #1876, opened 2020-05-09
fixed images.hover

Tylas11:
There was an error in the bounds calculation. The right/bottom sides were compared to pos + extents. But the extents are already a combination of pos + size.
Tested this using the drag & drop functionality. Before the fix, you could drag images by clicking outside their bounds.

### DIFF
--- a/addons/libs/images.lua
+++ b/addons/libs/images.lua
@@ -348,15 +348,13 @@ function images.hover(t, x, y)
         return false
     end
 
-    local pos_x, pos_y = t:pos()
-    local off_x, off_y = t:get_extents()
-    
-    -- print(x, y, pos_x, pos_y, off_x, off_y)
+    local start_pos_x, start_pos_y = t:pos()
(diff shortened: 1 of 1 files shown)


z16:
Good catch! Though if the offsets are not really offsets, I feel we should change the name to match. How about `start_pos_x`, `start_pos_y`, `end_pos_x` and `end_pos_y`?

Also, feel free to delete the print statement.

Tylas11:
Sure, I could rename them.
I also just noticed in the texts.lua there is a similar setup for hover, but it uses windower.text.get_extents which only returns width / height according to the documentation. Images uses a custom get_extents. This is kind of inconsistent and probably the cause of the bug, but I didn't want to change the images.get_extents function, who knows who uses it and what it would break.
Thoughts?

z16:
Good catch and good instincts. Changing it at this point would likely cause too much breakage. Let's leave it as it is and just adjust this part where it is used. This fits well into the rapidly growing "lesson learned for v5" category.

PR #1886, opened 2020-06-13
Add GearSwap support

DrakeFS:
Disables the equipment slot being used until item is used, then re-enables equipment slot.  Prevents GS from swapping the item until it has been used.

### DIFF
--- a/addons/MyHome/MyHome.lua
+++ b/addons/MyHome/MyHome.lua
@@ -32,12 +32,21 @@ _addon.commands = {'myhome','mh','warp'}
 
 require('logger')
 extdata = require('extdata')
+res_slots = require('resources').slots
 
 lang = string.lower(windower.ffxi.get_info().language)
 item_info = {
     [1]={id=28540,japanese='デジョンリング',english='"Warp Ring"',slot=13},
     [2]={id=17040,japanese='デジョンカジェル',english='"Warp Cudgel"',slot=0},
     [3]={id=4181,japanese='呪符デジョン',english='"Instant Warp"'}}
+
+function gs_disable_slot(slot)
+	windower.send_command('gs disable '..res_slots[slot].en:gsub(' ','_'))
+end
+
+function gs_enable_slot(slot)
+	windower.send_command('wait 3;gs enable '..res_slots[slot].en:gsub(' ','_'))
+end
 
 function search_item()
     local item_array = {}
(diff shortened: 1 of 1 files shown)


RubenatorX:
1) this is probably not the proper way to do this, especially since if they have the slot intentionally disabled then this would re-enable it regardless of if it was originally already disabled
2) `windower.to_shift_jis(stats[lang]) == '"Warp Ring"' or '"Warp Cudgel"'`
This always evaluates to true because `'"Warp Cudgel"'` is Truthy
3) you would probably want to disable the slot before equipping the item, not after waiting for the equip cooldown timer

z16:
I am quite uncomfortable with this PR. At the *very* least this needs to be a toggle and default to off. Though cross-addon commands are something I am generally not very fond of, they assume the user has the other addon installed which may not be the case (even GS, there are more people who do not use it than you may think).

I also feel I am missing something here, since I cannot see this being much of an issue. Does it happen often that GS changes gear between item usage and activation time? That is all that is being prevented here, right? How would that case even occur, the player cannot perform any other actions in that time in the first place, or is this meant to protect against accidental macro activation/key presses during that time?

Regardless, the tabs need to be replaced with four spaces. I will sleep on this and think more about it tomorrow, feel free to provide any additional info.

RubenatorX:
> Does it happen often that GS changes gear between item usage and activation time?
 No, unless you're using a selindrile-esque gearswap which uses prerender to sometimes swap gear to what it things you should be in, meaning, if it discovers that you're not in your idle set it will swap your gear, potentially between equipping the ring and the use going off. But in Selendrile's case, it has a special command to use an item that disables the slot and whatnot till the item gets used. It's just not "smart" enough to try multiple warp items if one is on cooldown.

lili-ffxi:
> this is probably not the proper way to do this, especially since if they have the slot intentionally disabled then this would re-enable it regardless of if it was originally already disabled

I second this, unfortunately this kind of interoperability is wonky because it can't restore previous state since it's not able to check for it in the first place.

>  Does it happen often that GS changes gear between item usage and activation time? 

Adding to what Rubenator mentioned, _lots_ or popular GS profiles use the `status_change()` function to return to Idle states after a cast, to avoid having duplicate logic in `aftercast()`, `buff_change()`, and `status_change()`. This has the side effect that if you gain or lose a buff between equipping an item and losing it, the `status_change()` function will kick in and possibly remove the item.

However, the proper way to go at this imo is to make the gearswap profile behave nicely around item usage, rather than editing MyHome.

Note that I do have the same edit to my own copy of MyHome, but a) I'm 100% sure that I have GS loaded and b) my gs profiles are written knowing that this behavior exists.

DrakeFS:
Will Close request as I can see why you would not want cross addon commands.  I also did not notice it was disabling the slot after the count down (whoops).  The learning continues.

PR #1887, opened 2020-06-13
Hardening against system<->server clock discrepancy

lili-ffxi:
Both Dimmer and Myhome stop the loop if the cooldown of Warp items and Dimensional Rings is greater than 10s. Unfortunately, it's common for Windows' clock to be off from FFXI's clock by several seconds, even when Internet time is turned on, thus falsely increasing this value.

While the actual solution would be for the user to fix their system time, increasing the upper limit of `delay` to 30s is a good workaround to the issue, that has no side effects.

This is the behavior after the change, when the system clock is off:
image

### DIFF
--- a/addons/Dimmer/Dimmer.lua
+++ b/addons/Dimmer/Dimmer.lua
@@ -75,7 +75,7 @@ function search_item()
                             log_flag = false
                             log('Item use within 3 seconds..')
                         end
-                    until ext.usable or delay > 10
+                    until ext.usable or delay > 30
                 end
                 windower.chat.input('/item '..windower.to_shift_jis(stats[lang])..' <me>')
                 break;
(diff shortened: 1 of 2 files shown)


smd111:
this should give you the exact time remaining till you can use rings again

`os.time(os.date("!*t", ext.next_use_time))-os.time()`

at lest this is what i use in my gearswap to automate my cp/xp rings
https://github.com/smd111/Gearswap/blob/master/SMDinclude/includes/extra.lua ::line 157

lili-ffxi:
Thanks for the snippet, but that fixes the wrong problem. This issue was not about FFXI timestamps being weird, rather about local system time being off than "real time" by x seconds, which is a very common occurrence for multiple reasons. Like, it's common for Windows to think that it's 10:45:37 AM when it's actually 10:45:32 AM. This was breaking MyHome and Dimmer because they were being too strict about how much cooldown a ring could have.

smd111:
thats why i used
os.time(os.date("!*t", ext.next_use_time))
os,time is the local system time and i use this to grab the local time that the ring is usable again
at worst you wait longer (seconds) then necessary to use it again but you cant miscalculate

PR #1889, opened 2020-06-14
added position_manager addon

lili-ffxi:
Manages screen position for different characters.
Requires the `WinControl` plugin.

### DIFF
--- a/addons/position_manager/README.md
+++ b/addons/position_manager/README.md
@@ -2,13 +2,22 @@
 
 Set and save screen position per-character.  
 
-Command: //pm set <Charactername|all> pos_x pos_y
+Command: `//pm set pos_x pos_y [name]`
+
+`pos_x` and `pos_y` are obligatory and must be numbers.  
+The `name` parameter is optional. If no name is provided, settings will be saved for the current character.  
+`:all` is a special name that can be used to set the default position.  
+
+**Note**: Characters are only moved after they're logged in. The `:all` position will be used for the character login screen as well.
 
 ### Examples:  
-`//pm set all 0 60`  
-Will set the default positioning for all characters to X: 0 and Y: 60 (the height of the Windows 10 taskbar with 150% UI scaling.)  
+`//pm set 0 0`  
+Will set your _current_ character to the position X: 0, Y: 0.
 
-`//pm set Yourname 1920 0`  
+`//pm set 0 60 :all`  
+Will set the default positioning for all characters to X: 0 and Y: 60 (the height of the Windows 10 taskbar with 150% UI scaling.), and delete all other character-specific settings.  
+
+`//pm set 1920 0 Yourname`  
 Will set the default position for the character called "Yourname" to X: 1920 and Y: 0.  
 This will make the character appear on the secondary screen that is to the right of the main screen - useful for multi-screen setups.
 
@@ -16,8 +25,6 @@ This will make the character appear on the secondary screen that is to the right
 `//pm set Youralt 800 40`  
 Will set your main to X: 0, Y: 40, and your alt to the X: 800, Y: 40.  
 If your laptop screen is 1600px wide, and your instances are both set at 800x600, this will put them side by side.
-
-Note: Characters are only moved after they're logged in. The <all> position will be used for the character login screen as well.
 
 **Warning:** the `all` name will delete every other character-specific settings that are already saved! It's best to use it only once after you install the addon, to set default position for non-specified characters.
 
--- a/addons/position_manager/README.md
+++ b/addons/position_manager/README.md
@@ -2,7 +2,7 @@
 
 Set and save screen position per-character.  
 
-Command: `//pm set pos_x pos_y [name]`
+Command: `//pm set <pos_x> <pos_y> [name]`
 
 `pos_x` and `pos_y` are obligatory and must be numbers.  
 `name` is optional. If no name is provided, settings will be saved for the current character.  
--- a/addons/position_manager/README.md
+++ b/addons/position_manager/README.md
@@ -5,7 +5,7 @@ Set and save screen position per-character.
 Command: `//pm set pos_x pos_y [name]`
 
 `pos_x` and `pos_y` are obligatory and must be numbers.  
-The `name` parameter is optional. If no name is provided, settings will be saved for the current character.  
+`name` is optional. If no name is provided, settings will be saved for the current character.  
 `:all` is a special name that can be used to set the default position.  
 
 **Note**: Characters are only moved after they're logged in. The `:all` position will be used for the character login screen as well.
--- a/addons/position_manager/README.md
+++ b/addons/position_manager/README.md
@@ -1,0 +1,24 @@
+# Position Manager
+
+Set and save screen position per-character.  
+
+Command: //pm set <Charactername|all> pos_x pos_y
+
+### Examples:  
+`//pm set all 0 60`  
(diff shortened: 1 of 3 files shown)


addons/position_manager/position_manager.lua
```diff
@@ -0,0 +1,83 @@
+--Copyright (c) 2020, Lili
```
z16: `(c)` should be a `©`.

addons/position_manager/position_manager.lua
```diff
@@ -0,0 +1,83 @@
...
+_addon.version = '1.0.0'
+_addon.command = 'pm'
+
+if not windower.file_exists(windower.windower_path .. '\\plugins\\WinControl.dll') then
+    print('position_manager: error - Please install the WinControl plugin in the launcher.')
+    windower.send_command('lua u position_manager')
+    return
+end
+
+config = require 'config'
```
z16: Parentheses around `'config'`.

addons/position_manager/position_manager.lua
```diff
@@ -0,0 +1,83 @@
...
+
+settings = config.load(default)
+
+function move(settings)
+    windower.send_command('load wincontrol')
+    coroutine.sleep(0.5)
+    windower.send_command('wincontrol move %s %s':format(settings.x,settings.y))
+    coroutine.sleep(0.5)
+    windower.send_command('unload wincontrol')
+end
```
z16, line 52: I am *certain* this will make users complain. If someone wants to have *WinControl* loaded all the time this will unload it again, because it performs no checks (which is not your fault, it is only really possible with an ugly hack).

I think expecting the plugin to be loaded is reasonable in this case. Would be faster, and would not hurt anyone's current workflow.

addons/position_manager/position_manager.lua
```diff
@@ -0,0 +1,83 @@
...
+            settings.x = tonumber(pos_x)
+            settings.y = tonumber(pos_y)
+            config.save(settings,name)
+            
+            if windower.ffxi.get_info().logged_in then
+                player_name = windower.ffxi.get_player().name
+                if name:lower() == player_name:lower() then
+                    move(settings)
+                end
+                -- TODO: add IPC
```
z16: This is something I am very unsure about. It will make moving the *current* window a chore, since you always have to look up and type out the character name. How about foregoing IPC completely and only make it work locally? It could be that this is me being out of touch with the game, is it so uncommon to switch to a mule's window? I know I wanted to control my mule without ever having to switch to their screen, but that was about gameplay, setting up window positions once in a while is a situation where I feel that having to switch to the respective window is a reasonable expectation.

And do not misunderstand me, I do feel this is a useful feature, but it does come at a cost, namely making the *current* window harder to adjust, which I feel will be a more common use case.

addons/position_manager/position_manager.lua
```diff
@@ -0,0 +1,83 @@
...
+function move(settings)
+    windower.send_command('load wincontrol')
+    coroutine.sleep(0.5)
+    windower.send_command('wincontrol move %s %s':format(settings.x,settings.y))
+    coroutine.sleep(0.5)
+    windower.send_command('unload wincontrol')
+end
+
+function handle_commands(cmd,name,pos_x,pos_y)
+    local cmd = cmd or 'help'
```
z16: Can actually omit the `local` here, it will still be local since it only adjusts itself. And while not necessary, it is quite customary to be case-insensitive for v4 addons, this is a common pattern:
```lua
cmd = cmd and cmd:lower() or 'help'
```

addons/position_manager/README.md
```diff
@@ -0,0 +1,24 @@
+# Position Manager
+
+Set and save screen position per-character.  
+
+Command: //pm set <Charactername|all> pos_x pos_y
```
z16: I would probably add backticks around this for code formatting, but just a minor suggestion.

addons/position_manager/position_manager.lua
lili-ffxi: Fixed.

addons/position_manager/position_manager.lua
lili-ffxi: Serves me right for blindly copy-pasting >_> Fixed.

addons/position_manager/position_manager.lua
lili-ffxi: That's a good point, albeit once position_manager is loaded there's no longer any reason to keep wincontrol loaded at all times, and if one wants to do some testing they can always reload the plugin manually. In any case, I'll remove the unload command.

I still would like to avoid situations where the command is sent but nothing happens, so what do you think about sending a `load wincontrol` command once on addon load, rather than each time a window is moved?

z16: I would much prefer it to loading it for every move, though I still feel it is not *necessary*, but in this case I feel my personal preference for clean design is getting in the way of practicality, so I will completely defer to you :D

z16: Oh and I forgot to say, WinControl also has some other features, resizing and setting it to the foreground. I did not really think that anyone used those, but I also did not think anyone used WinControl at all, so shows how much I know :X

addons/position_manager/position_manager.lua
lili-ffxi: This is 100% designed around my own use case, where I was able to set a position for all of my characters while sitting in front of my main, _including characters that weren't logged in_ (i.e. same-account alts). 

What do you think about a `current` keyword instead? The addon would sub in the current character's name. 
Or I could make the name parameter optional entirely, altough I'm aware that you don't like that too much :P

addons/position_manager/position_manager.lua
lili-ffxi: Fixed.

addons/position_manager/position_manager.lua
z16: Good point, I had not considered logged out characters. That could be useful.

In that case I would suggest switching the order, append the character name, or omit it if you want the current character. Does that sound ok? I would not min the current solution either by the way, it was just something that I thought could be streamlined.

addons/position_manager/position_manager.lua
lili-ffxi: I had no idea those functions of WinControl existed myself! The setting to foreground might be useful to reimplement what FFXISwitchMon used to do, but I digress.

I think I prefer having at least one load onload: if the addon has been enabled in the launcher, there's a reasonable expectation for it to _do something_. With the plugin not loaded, it would fail entirely silently, which would lead users to come asking "hey why doesn't it work", and I'd like to avoid that. At worst it'll blurt out one "plugin already loaded" error, but it's only once and pretty early in the login process, so I feel it wouldn't be too bothersome, and it's preferable imo to failing entirely silently.

addons/position_manager/README.md
lili-ffxi: Fixed.

addons/position_manager/position_manager.lua
lili-ffxi: Switched the order around, changed `all` to `:all` because magic word, added some sanity checks. Gave up on `:current` as it'd be redundant with the order switched.

z16:
Looking good! To be added to the launcher you also need to modify the `addons.xml` file. After that, this should be ready.

<sub>(I would really ❤️ you if you could add spaces after commas though...)</sub>

PR #1890, opened 2020-06-16
Change Dressup Autoupdate default to on

RubenatorX:
Having this default to false makes people new to the addon think that its not working. So would love to change it to true


RubenatorX:
Hold off on this actually, need to modify the update to inject incoming instead of outgoing player update request :/

PR #1892, opened 2020-06-21
Increased delay

lili-ffxi:
Under certain circumstances, sending `/macro book` and `/macro set` too fast can cause the loss of all macros contained in that book. Increasing the delay fixes that.


z16:
Are those circumstances something that can be fixed instead? This feels like a band-aid solution. And the waits here are annoyingly long, is that necessary or can they be reduced?

lili-ffxi:
No, it doesn't look like those circumstances can be fixed windower side - it looks like it's purely a behavior of the client, trying to load macros twice in rapid succession will corrupt the file somehow, and cause the client to wipe it. 

This can be reproduced consistently and easily across multiple systems, I just am not able to pinpoint exactly what causes it. Increasing the delay between the two commands to match the minimum delay possible on the client seems to solve the issue entirely.

PR #1900, opened 2020-07-17
Bug/misc

WarrenFerrell:
Some improvements and bug fixes I've implemented and figure could be useful to others. Not sure about your branching strategy, so branch name is pretty generic. I also have a filtered logger that I could pull together a PR for but involves much more significant changes.

### DIFF
--- a/addons/libs/functions.lua
+++ b/addons/libs/functions.lua
@@ -72,7 +72,7 @@ end
 function functions.prepare(fn, ...)
     local args = {...}
     return function()
-        fn(unpack(args))
+        return fn(unpack(args))
     end
 end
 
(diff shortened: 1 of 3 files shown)


z16:
We don't have a particular naming scheme, the only thing we'd like to ask is to make different PRs for different features, even minor ones, like these. I'd also like to mention there's a bug (an extra `end`) in your `maths.lua` adjustment, and we use 4 spaces in our code for each indentation level, not 2. Also you should call the JSON function `json.to_json` to mirror `xml.to_xml`.

WarrenFerrell:
Sorry about the stray end. I cherry picked the round function because I figured most people don't care about having a round function that rounds to non decimal numbers #duodecimal4life

PR #1902, opened 2020-07-17
return the value from the function created by prepare.

WarrenFerrell:
This issue caused me quite a bit of pain to figure out what was going on. I'm actually curious where the name prepare came from, I haven't been able to find it anywhere.

### DIFF
--- a/addons/libs/functions.lua
+++ b/addons/libs/functions.lua
@@ -72,7 +72,7 @@ end
 function functions.prepare(fn, ...)
     local args = {...}
     return function()
-        fn(unpack(args))
+        return fn(unpack(args))
     end
 end
 


z16:
Not sure what you mean by that, it's just a name :D Most functions did not have a lot of thought put into them, this function works similar to `apply`, which is a more common name for what it does, but I chose `prepare` because you prepare a function to pass it into other functions that take functions without an option to supply additional arguments.

WarrenFerrell:
I'm used to apply, I suppose prepare would generally be considered bad by functional programmers because it's all about side effects but I've found it very useful when repeating the same action over and over.

z16:
Just to make sure I explained myself correctly, we already have an `apply` function as well, and its behavior is slightly different (and closer to what functional programmers are used to with regards to currying). The difference is that the function produced by `prepare` does *not* allow any additional arguments to be passed to the base function. Take these examples:
```lua
local print_a = print:prepare(2, 3)
local print_b = print:apply(2, 3)

print_a('a', 'b', 'c') -- prints: 2    3
print_b('a', 'b', 'c') -- prints: 2    3    a    b    c
```

Or put differently, `apply` produces a new function that takes any number of arguments and *appends* them to the arguments you passed to `apply`, whereas `prepare` produces a new function that takes no more arguments, whatever you pass to the result will be ignored and only the arguments you provided to `prepare` are used.

PR #1903, opened 2020-07-17
Feature/to json

WarrenFerrell:
My sole contributions to this code were to rename stringify to to_json and change the 2 space indenting to 4 (sadly I could not use vscode-lua-format because it is the most finicky and invasive formatter I have ever worked with)

### DIFF
--- a/addons/libs/json.lua
+++ b/addons/libs/json.lua
@@ -281,6 +281,61 @@ function json.classify(tokens, root)
     return parsed:remove()
 end
 
+-- from https://gist.github.com/tylerneylon/59f4bcf316be525b30ab
+local function kind_of(obj)
+  if type(obj) ~= 'table' then return type(obj) end
+  local i = 1
+  for _ in pairs(obj) do
+    if obj[i] ~= nil then i = i + 1 else return 'table' end
+  end
+  if i == 1 then return 'table' else return 'array' end
+end
+
+local function escape_str(s)
+  local in_char  = {'\\', '"', '/', '\b', '\f', '\n', '\r', '\t'}
+  local out_char = {'\\', '"', '/',  'b',  'f',  'n',  'r',  't'}
+  for i, c in ipairs(in_char) do
+    s = s:gsub(c, '\\' .. out_char[i])
+  end
+  return s
+end
+
+function json.stringify(obj, as_key)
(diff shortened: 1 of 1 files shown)


z16:
I'm unhappy about this for several reasons. Even just the idea of adding such a function at this point does not make me very happy, as its use seems very limited for our purposes, and with Windower 4 reaching the end of its life I am reluctant to add any features at this point, even if it were more broadly useful. And this library in particular has never worked all that well.

Also this code does not work well with our built-in libraries, a list (as in `L{}`) would for example not be considered a kind of array under these checks. The style of the code is also very different from what we use, so it would require a bit of a rewrite (you still did not fix the indentation fully, and the one-line `if` is also not something we do and the last error message is stupid, and various other minor things). And it also reimplements a few things we already provide.

But what makes me feel the most uneasy is that it is code copied from elsewhere, which causes several other issues. While the author did clarify in the comments that they put it in the public domain, it's just a comment which does not give me a lot of confidence to just use it. And once we modify it enough to fit our style and work with our API it honestly would be so much easier to code it from scratch, especially because it really is such a simple thing. If you were willing to rewrite it, I would reconsider (in which case I'd ask you to make a new PR), but if you really need this particular function I think it's best if you just use it in your own code without submitting it to our core libs. Even more so considering this one is hardly used and we talked about removing it entirely before.

WarrenFerrell:
Totally don't mind, it's not good code. Find it a little odd actually that there aren't any popular json libraries for lua already out there. What do you mean by end of life for Windower 4 ?

z16:
While JSON is useful for communicating with external APIs, native Lua code is very similar in structure to JSON, so that is what most Lua data is written in. We debated a similar thing when we wrote our ResourceExtractor, but then decided to just output Lua instead of JSON, which was a minimal change and made it easier to load, so I don't find it quite that surprising.

What I mean by that is that Windower 4 as it is is feature complete and we have stopped active development on it. If new PRs are submitted by the community we still vet and merge them, but our own efforts are focused entirely on [Windower 5](https://github.com/Windower/packages/wiki) at this point.

If you want to learn more about it and the current state of Windower and want to discuss similar topics with us feel free to join our [Discord server](https://discord.gg/aUrHCvk).

PR #1910, opened 2020-07-27
[Gearswap] fixed 2 bugs

KenshiDRK:
Fixed a bug when getting charmed and the charming mob dying very soon after and adding more sittuations where you can't summon trust and gs was not calling aftercast.

Captura

### DIFF
--- a/addons/GearSwap/packet_parsing.lua
+++ b/addons/GearSwap/packet_parsing.lua
@@ -334,7 +334,8 @@ parse.i[0x117] = function (data)
 end
 
 parse.i[0x053] = function (data)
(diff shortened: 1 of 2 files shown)


addons/GearSwap/packet_parsing.lua
```diff
@@ -334,8 +334,8 @@ parse.i[0x117] = function (data)
 end
 
 parse.i[0x053] = function (data)
-    if data:unpack('H',0xD) == 0x12D and player then
-        -- You're unable to use trust magic if you're not the party leader or solo
+    if (data:unpack('H',0xD) == 0x12D or data:unpack('H',0xD) == 0x12A or data:unpack('H',0xD) == 0x12B or data:unpack('H',0xD) == 0x12C) and player then
```
z16: This value should really be unpacked once and then compared, the `unpack` operation is far from the slowest in our API but it's not fast either. Also it helps make the code more maintainable.

PR #1912, opened 2020-08-05
[battlemod] fix and enhancements

KenshiDRK:
Fixed a few issues and changed a little how it displays shield block messages while also adding guarding and shield blocks to ws's. (parrying seems to not be used but leaving just in case)


Byrth:
Random, but do you mind also adding desynth HQ support in this PR?
~~~
    elseif id == 0x06F then
        if original:byte(5) == 0 or original:byte(5) == 12 then
            local result = original:byte(6)
            if result == 1 then
                windower.add_to_chat(8,' -------------- HQ Tier 1! --------------')
            elseif result == 2 then
                windower.add_to_chat(8,' -------------- HQ Tier 2! --------------')
            elseif result == 3 then
                windower.add_to_chat(8,' -------------- HQ Tier 3! --------------')
            end
        end
~~~

Byrth (commented):
Thanks!

Issue #1916, opened 2020-08-08
Config lib overwrites global settings when using lists or sets


xurion:
When using the config lib to store a list or set (using `S{}` or `L{}`) the global settings are overloaded after reloading an addon and modifying settings.

You can easily reproduce this issue with the Autojoin addon:

1. Remove settings.xml if it already exists for Autojoin.
2. Load Autojoin.
3. Add to the whitelist with `//autojoin whitelist add xurion`. The settings.xml file will look ok at this point.
4. Reload Autojoin.
5. Add a second entry on to the whitelist.

The `<global>` in settings.xml will now be polluted with incorrect settings that will now be applied to all other characters.

RubenatorX:
this is not a problem with the config lib, Its a problem with the addon. It needs to be re-loading settings on login, and its not doing that.

RubenatorX:
Actually I'm wrong, it definitely decides to add some of the settings to global, not sure why will chat with someone about it.

PR #1917, opened 2020-08-09
MountRoulette addon

xurion:
Summons a mount at random based on your mount key items.

A function to blacklist certain mounts exists on another branch and isn't included in this PR because of the impact of https://github.com/Windower/Lua/issues/1916 (blacklists mounts unexpectedly across multiple characters).

### DIFF
--- a/addons/MountRoulette/MountRoulette.lua
+++ b/addons/MountRoulette/MountRoulette.lua
@@ -1,0 +1,79 @@
+_addon.name    = 'Mount Roulette'
+_addon.author  = 'Dean James (Xurion of Bismarck)'
+_addon.version = '3.0.0'
+_addon.commands = {'mountroulette', 'mr'}
+
+require('lists')
+resources = require('resources')
+
+math.randomseed(os.time())
+
+allowed_mounts = L{}
+possible_mounts = L{}
+for _, mount in ipairs(resources.mounts) do
+    possible_mounts:append(mount.name:lower())
+end
+
+function update_allowed_mounts()
+    local kis = windower.ffxi.get_key_items()
+
+    for _, id in ipairs(kis) do
+        local ki = resources.key_items[id]
+        if ki.category == 'Mounts' and ki.name ~= "trainer's whistle" then -- Don't care about the quest KI
+            local mount_index = possible_mounts:find(function(possible_mount)
+                return windower.wc_match(ki.name:lower(), '♪' .. possible_mount .. '*')
+            end)
+            local mount = possible_mounts[mount_index]
+
+            -- Add this to allowed mounts if it is not already there
+            if not allowed_mounts:contains(mount) then
+                allowed_mounts:append(mount)
+            end
+        end
+    end
+end
+
+windower.register_event('load', function()
+    update_allowed_mounts()
+end)
+
+windower.register_event('incoming chunk', function(id)
+    if id == 0x055 then --ki update
+        update_allowed_mounts()
+    end
+end)
+
+windower.register_event('addon command', function(command)
+    command = command and command:lower() or 'mount'
+
+    if commands[command] then
+        commands[command]()
+    else
+        commands.help()
+    end
+end)
+
+commands = {}
+
+commands.mount = function()
+    local player = windower.ffxi.get_player()
+
+    -- If the player is mounted, dismount now
+    for _, buff in pairs(player.buffs) do
+        if buff == 252 then --mounted buff
+            windower.send_command('input /dismount')
+            return
+        end
+    end
+
+    -- Generate random number and use it to choose a mount
+    local mount_index = math.ceil(math.random() * #allowed_mounts)
+    windower.send_command('input /mount ' .. allowed_mounts[mount_index])
+end
+
+commands.help = function()
+    windower.add_to_chat(8, '---Mount Roulette---')
+    windower.add_to_chat(8, 'Available commands:')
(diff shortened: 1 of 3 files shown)


addons/MountRoulette/MountRoulette.lua
```diff
@@ -0,0 +1,79 @@
...
+    update_allowed_mounts()
+end)
+
+windower.register_event('incoming chunk', function(id)
+    if id == 0x055 then --ki update
+        update_allowed_mounts()
+    end
+end)
+
+windower.register_event('addon command', function(command)
```
z16: While this pattern by itself is nice for more complicated command handlers I feel it's unnecessarily complex here. I'd just make `//mr` run the roulette and remove the help text and all the boilerplate surrounding it, as well as the help message for it. I don't see a benefit to having this setup,

addons/MountRoulette/MountRoulette.lua
```diff
@@ -0,0 +1,79 @@
...
+
+            -- Add this to allowed mounts if it is not already there
+            if not allowed_mounts:contains(mount) then
+                allowed_mounts:append(mount)
+            end
+        end
+    end
+end
+
+windower.register_event('load', function()
```
z16: I would probably forego the `load` event here and just call the function directly. Its only purpose is if the code should execute after the full file has been loaded, but that does not seem to be necessary here.

addons/MountRoulette/MountRoulette.lua
```diff
@@ -0,0 +1,79 @@
...
+            end)
+            local mount = possible_mounts[mount_index]
+
+            -- Add this to allowed mounts if it is not already there
+            if not allowed_mounts:contains(mount) then
+                allowed_mounts:append(mount)
+            end
+        end
+    end
+end
```
z16, line 35: I think this function needs to clear the `allowed_mounts` every time it's called. Otherwise if you log into a character that has all mounts, then log out and into a mule that has none, it will still have all mounts unlocked, no? Or am I missing something here?

addons/MountRoulette/MountRoulette.lua
```diff
@@ -0,0 +1,79 @@
...
+    for _, id in ipairs(kis) do
+        local ki = resources.key_items[id]
+        if ki.category == 'Mounts' and ki.name ~= "trainer's whistle" then -- Don't care about the quest KI
+            local mount_index = possible_mounts:find(function(possible_mount)
+                return windower.wc_match(ki.name:lower(), '♪' .. possible_mount .. '*')
+            end)
+            local mount = possible_mounts[mount_index]
+
+            -- Add this to allowed mounts if it is not already there
+            if not allowed_mounts:contains(mount) then
```
z16: If the `allowed_mounts` are cleared at the top of the function this check could be removed, right? That said, if that was made a set (`S{}`) you could remove it regardless, which is also what I would do, even if it's temporary and converted to a list later:
```lua
allowed_mounts = L(allowed_mounts_set)
```

addons/MountRoulette/MountRoulette.lua
```diff
@@ -0,0 +1,79 @@
...
+    end
+end
+
+windower.register_event('load', function()
+    update_allowed_mounts()
+end)
+
+windower.register_event('incoming chunk', function(id)
+    if id == 0x055 then --ki update
+        update_allowed_mounts()
```
z16, line 41: Have you verified that this works once you get a new mount? I don't think it should, since the received KI packet will trigger the function, but at that point it hasn't reached the game yet, so `windower.ffxi.get_key_items()` would not know about it, I think. You may want to run it in a slightly delayed coroutine:
```lua
update_allowed_mounts:schedule(0.2)
```

addons/MountRoulette/MountRoulette.lua
xurion: No worries - this is the boilerplate left over from  removing the blacklist stuff. I'll cut this out as well.

addons/MountRoulette/MountRoulette.lua
xurion: You're absolutely right, I'll clear it.

addons/MountRoulette/MountRoulette.lua
xurion: Ahh I did not know that - I've probably not ran into this issue due to getting new mounts in Upper Jeuno, then needing to go outside to summon a mount. Given this, do still think it's worth the schedule call?

xurion: Thinking about it, EmpyPopTracker has this same pattern and works ok. Is it more of a race condition?

z16: Actually, now that I think about it this should work after all. We store our own copy of the key items, and `windower.ffxi.get_key_items` returns that, not the game's key item store, so it shouldn't matter. At least I think so...

If you have some mounts left to get you could test it by the function print out the mounts it has found, then you'll see if it's really there.

xurion: I think an alt of mine has some mounts to turn in so I'll check that soon.

xurion: Thankfully it seems to work fine:

image

xurion:
One thing I haven't considered is if someone uses this on a character with zero mounts. You'll get an error when it tries to concat the /mount with the mount name - should I cater for this?

z16:
Ah why not, it's good form.

xurion:
Ok cool, I just threw a return statement in there rather than logging something.

PR #1923, opened 2020-08-20
EPT: Added item of interest tracking

xurion:
EPT now tracks the collectable item of interest. For example, Isgebind Hearts.

https://github.com/xurion/ffxi-empy-pop-tracker/issues/13


addons/EmpyPopTracker/EmpyPopTracker.lua
```diff
@@ -100,6 +100,21 @@ function owns_item(id, items)
     return false
 end
 
+function get_item_count(id, items)
+    local count = 0
+    for _, bag in pairs(items) do
+        if type(bag) == 'table' then
+            for _, item in ipairs(bag) do
+                if item.id == id then
+                    count = count + 1
```
z16: Is it intended that it adds only one for each item, even if there are more of it? Like, if you had three bunches of Pamamas in one inventory slot, it would only add one, instead of three.

If you want to add the proper count you can adjust it very simply:
```lua
count = count + item.count
```

xurion: Oh nice catch, yeah I want the total count. I'll update now thanks!

PR #1930, opened 2020-08-30
Adding NyzulHelper addon.

GlarinAsura:
Initial commit of NyzulHelper addon.

### DIFF
--- a/addons/NyzulHelper/NyzulHelper.lua
+++ b/addons/NyzulHelper/NyzulHelper.lua
@@ -1,5 +1,5 @@
 --[[
-Copyright (c) 2020, Glarin of Asura
+Copyright © 2020, Glarin of Asura
 All rights reserved.
 
 Redistribution and use in source and binary forms, with or without
@@ -10,7 +10,7 @@ modification, are permitted provided that the following conditions are met:
     * Redistributions in binary form must reproduce the above copyright
       notice, this list of conditions and the following disclaimer in the
       documentation and/or other materials provided with the distribution.
-    * Neither the name of PartyMaster nor the
+    * Neither the name of NyzulHelper nor the
       names of its contributors may be used to endorse or promote products
       derived from this software without specific prior written permission.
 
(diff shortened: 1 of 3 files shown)


addons/NyzulHelper/NyzulHelper.lua
```diff
@@ -0,0 +1,306 @@
+--[[
+Copyright (c) 2020, Glarin of Asura
```
z16: This should be `©`, not `(c)`. I notice you also use `PartyMaster` in a line below. Make sure you you use [this text](https://github.com/Windower/Lua/wiki/Writing-Addons#licensing) and replace the marked instances.

addons/NyzulHelper/NyzulHelper.lua
```diff
@@ -0,0 +1,306 @@
...
+completed = 0
+floor_penalities = 0
+potential_tokens = 0
+
+
+-- Handle addon args
+windower.register_event('addon command', function(input, ...)
+
+    local cmd = string.lower(input)
+	local args = {...}
```
z16: Use four spaces for indentation, everywhere. should be safe to do a blanket replace.

addons/NyzulHelper/NyzulHelper.lua
```diff
@@ -0,0 +1,306 @@
...
+current_floor = 0
+completed = 0
+floor_penalities = 0
+potential_tokens = 0
+
+
+-- Handle addon args
+windower.register_event('addon command', function(input, ...)
+
+    local cmd = string.lower(input)
```
z16: This will error if the first parameter is not specified. The common idiom here is:
```lua
local cmd = input and input:lower() or 'help'
```
Or omit the last part, if you don't have/want a help text.

z16 (changes requested):
For the most part the addon looks fine, function-wise. Stylistically, since we try to go for a uniform style please use 4 spaces for one level of indentation, use single quotes for strings and add a new line at the end of the file, and prefer multi-line `if` over single-line.

Also, for the addon to be listed in the launcher you need to add an entry into the `addons.xml` file. You can copy the structure for other addons in there.

I wish I could provide some insight into the functional aspect as well, but I know too little about Nyzul to contribute, so I'll trust your judgment here.

addons/NyzulHelper/NyzulHelper.lua
GlarinAsura: Requested changes made.

addons/NyzulHelper/NyzulHelper.lua
GlarinAsura: Requested changes made.

addons/NyzulHelper/NyzulHelper.lua
GlarinAsura: Requested change made.

GlarinAsura:
Thank you so much for taking the time to review my code.  I have made all of the requested changes.

PR #1932, opened 2020-09-06
Added delay setting to position_manager.

lili-ffxi:
On some systems, wincontrol might end up being loaded after position_manager sends the move request. Adding a delay setting to position_manager to overcome the issue if needed.


addons/position_manager/position_manager.lua
```diff
@@ -43,22 +43,28 @@ config = require('config')
...
-    cmd = cmd:lower() or 'help'
+function handle_commands(cmd, ...)
+    cmd = cmd:lower()
 
     if cmd == 'r' then
         windower.send_command('lua r position_manager')
-    elseif cmd == 'set' and pos_x and pos_y then
+        return
+    elseif cmd == 'set' then
+        local name = arg[3]
```
z16, line 73: Are you sure that works? I know they removed that feature from Lua (finally), but not sure in what version, I did think it was 5.1. Though I guess you tested it?

I still recommend explicitly defining a table, makes it appear less magicky, and is more in line with modern Lua.

addons/position_manager/position_manager.lua
```diff
@@ -67,31 +73,48 @@ function handle_commands(cmd, pos_x, pos_y, name)
...
+    elseif cmd ~= 'help' then
         windower.add_to_chat(207, 'position_manager: %s command not found.':format(cmd))
-        windower.send_command('pm help')
-    end
+    end        
+
+    windower.add_to_chat(207, 'position_manager: Commands:')
+    windower.add_to_chat(207, '  //pm set <x> <y> [name]')
+    windower.add_to_chat(207, '  //pm set delay <seconds> [name]')
+    windower.add_to_chat(207, 'position_manager: See the readme for details.')
```
z16: Not sure why you moved that away from the `help` command. This way it will also print after setting a new delay or position. Was that intended? If so, you have to remember to put `return` at the end of every successful command.

If you want it to print when passing in invalid info I would suggest making a `syntax` function or similar and call that in case of an invalid input (and possibly within the `help` command as well).

addons/position_manager/position_manager.lua
lili-ffxi: It does work correctly in Windower, but I agree with your last statement. Fixed.

addons/position_manager/position_manager.lua
lili-ffxi: > Not sure why you moved that away from the `help` command. This way it will also print after setting a new delay or position. Was that intended? If so, you have to remember to put `return` at the end of every successful command.

It's this way because I was writing with IPC in mind - look at lines 106 and 107, there's a `return` that takes care of those two cases, the line right above is supposed to send an IPC message to update the affected character (the function at 93-98 would then go in the ipc message event, and take care of reloading settings and updating the position).

> If you want it to print when passing in invalid info I would suggest making a `syntax` function or similar and call that in case of an invalid input (and possibly within the `help` command as well).

With the above in mind, a show_help() function won't simplify the code because the if tree would be unchanged, but it'd let me get rid of the `windower.send_command('pm help')` calls, so I'll do it.

PR #1942, opened 2020-09-15
Update Pouches.lua

ianandersonlol:
Updated pouches to check if your inventory is full before trying to use the pouch.

Having pouches try and fail to use items when your inventory is full is mildly problematic. No real player is going to sit there for minutes to hours opening pouches they can't use. If SE decided to get strict this is a good check.

### DIFF
--- a/addons/Pouches/Pouches.lua
+++ b/addons/Pouches/Pouches.lua
@@ -28,7 +28,7 @@ SOFTWARE, EVEN IF ADVISED OF THE POSSIBILITY OF SUCH DAMAGE.
 
 
 _addon.name = 'Pouches'
-_addon.version = '1'
+_addon.version = '2'
 _addon.author = 'Omnys@Valefor'
 _addon.command = 'pouches'
 
@@ -41,7 +41,10 @@ res = require('resources')
 
 inverted = {}
 item = {}
-    
+inventory_info = {} 
+inventory_info.current = windower.ffxi.get_bag_info(0).count
+inventory_info.max = windower.ffxi.get_bag_info(0).max
+
 windower.register_event('load', function(...)
     for k,v in pairs(res.items) do
         inverted[string.lower(v.english)] = {id = k, targets = v.targets, cast = v.cast_time}
@@ -51,11 +54,21 @@ end)
 function use_item()
     windower.chat.input('/item "' .. item.name .. '" <me>')
     item.count = item.count - 1
-    if item.count > 0 and windower.ffxi.get_player().status == 0 then
-        use_item:schedule(item.delay)
+    if item.count > 0 and windower.ffxi.get_player().status == 0 then	
+        if inventory_info.current < inventory_info.max then 
+            use_item:schedule(item.delay)
+        else
+            log("Your inventory is full...Aboriting")
+		  end
     end
 end
 
+windower.register_event('incoming chunk', function(id)
+    if id == 0x020 or 0x01E or 0x01D then
+        inventory_info.current = windower.ffxi.get_bag_info(0).count
+        inventory_info.max = windower.ffxi.get_bag_info(0).max -- this is probably not needed, but I don't want a user to have to reload when they complete a gobbie bag quest.  
+    end
+end)
 windower.register_event('addon command', function(...)
     local inv = windower.ffxi.get_items(0) -- get main inventory
     local args    = T{...}:map(string.lower)
@@ -73,9 +86,13 @@ windower.register_event('addon command', function(...)
             end
         end
         if item.count > 0 then
-            log('Found '..item.count..' '..item.name..'. Commencing Use.')
-            log('You may simply type /heal to stop.')
-            use_item()
+            if inventory_info.current < inventory_info.max then 
+                log('Found '..item.count..' '..item.name..'. Commencing Use.')
+                log('You may simply type /heal to stop.')
+                use_item()
+            else
+                log("Your inventory is full...Aboriting")
+            end
         else
             log('Item, '..item.name..' not found in main inventory, or is not the type of item that can be used.')
         end


addons/Pouches/Pouches.lua
```diff
@@ -51,11 +54,21 @@ end)
 function use_item()
     windower.chat.input('/item "' .. item.name .. '" <me>')
     item.count = item.count - 1
-    if item.count > 0 and windower.ffxi.get_player().status == 0 then
-        use_item:schedule(item.delay)
+    if item.count > 0 and windower.ffxi.get_player().status == 0 then	
+        if inventory_info.current < inventory_info.max then
```
z16, line 58: I think it makes most sense to use this place to query for the current values for `current` and `max`. This would eliminate all other instances of this. Also, you should call the function only once:
```lua
local info = windower.ffxi.get_bag_info(0)
if info.count < info.max then
    -- do eet
else
    -- don't... do eet
end
```

addons/Pouches/Pouches.lua
```diff
@@ -41,7 +41,10 @@ res = require('resources')
 
 inverted = {}
 item = {}
-    
+inventory_info = {} 
+inventory_info.current = windower.ffxi.get_bag_info(0).count
+inventory_info.max = windower.ffxi.get_bag_info(0).max
```
z16, line 46: If you initialize it at check time you can also skip the globals altogether.

addons/Pouches/Pouches.lua
```diff
@@ -51,11 +54,21 @@ end)
 function use_item()
     windower.chat.input('/item "' .. item.name .. '" <me>')
     item.count = item.count - 1
-    if item.count > 0 and windower.ffxi.get_player().status == 0 then
-        use_item:schedule(item.delay)
+    if item.count > 0 and windower.ffxi.get_player().status == 0 then	
+        if inventory_info.current < inventory_info.max then 
+            use_item:schedule(item.delay)
+        else
+            log("Your inventory is full...Aboriting")
```
z16, line 61: Single quotes, and `Aboriting` :D

addons/Pouches/Pouches.lua
```diff
@@ -51,11 +54,21 @@ end)
...
+		  end
```
z16, line 62: Replace the tabs with 4 spaces (elsewhere too, if I missed it, can just do a blanket replace).

addons/Pouches/Pouches.lua
```diff
@@ -51,11 +54,21 @@ end)
...
     end
 end
 
+windower.register_event('incoming chunk', function(id)
+    if id == 0x020 or 0x01E or 0x01D then
+        inventory_info.current = windower.ffxi.get_bag_info(0).count
+        inventory_info.max = windower.ffxi.get_bag_info(0).max -- this is probably not needed, but I don't want a user to have to reload when they complete a gobbie bag quest.
```
z16, line 69: I think it's a good idea to keep it, but again, could save keeping track of this if you do the thang up there.

addons/Pouches/Pouches.lua
```diff
@@ -73,9 +86,13 @@ windower.register_event('addon command', function(...)
             end
         end
         if item.count > 0 then
-            log('Found '..item.count..' '..item.name..'. Commencing Use.')
-            log('You may simply type /heal to stop.')
-            use_item()
+            if inventory_info.current < inventory_info.max then
```
z16, line 89: I missed that you use it here as well. You could place the check in a function:
```lua
local check_space = function()
    local info = windower.ffxi.get_bag_info(0)
    return info.count < info.max
end
```
Then you can just use it wherever.

addons/Pouches/Pouches.lua
```diff
@@ -73,9 +86,13 @@ windower.register_event('addon command', function(...)
...
         if item.count > 0 then
-            log('Found '..item.count..' '..item.name..'. Commencing Use.')
-            log('You may simply type /heal to stop.')
-            use_item()
+            if inventory_info.current < inventory_info.max then 
+                log('Found '..item.count..' '..item.name..'. Commencing Use.')
+                log('You may simply type /heal to stop.')
+                use_item()
+            else
+                log("Your inventory is full...Aboriting")
```
z16, line 94: Single quotes and `Aboriting` again :D Also add a space after the ellipsis, or just use a comma (or a full stop).

z16 (changes requested):
Left some notes.

RubenatorX:
I am actually against this change (at least in its current form). Your inventory not being full is not a requirement to use items. You do not need empty inventory slots to use skillup books or silt or bead pouches etc. So preventing me from clearing my inventory of them using the addon is undesired behavior.
Really you should not be going afk while using the addon (or playing the game in general for that matter). I don't think this change should be in the repo.
If you were to do something like this, you would need to be checking for failure messages saying your inventory is full. Not just blanket preventing use of the addon if your inventory is full.

ianandersonlol:
> I am actually against this change (at least in its current form). Your inventory not being full is not a requirement to use items. You do not need empty inventory slots to use skillup books or silt or bead pouches etc. So preventing me from clearing my inventory of them using the addon is undesired behavior.
> Really you should not be going afk while using the addon (or playing the game in general for that matter). I don't think this change should be in the repo.
> If you were to do something like this, you would need to be checking for failure messages saying your inventory is full. Not just blanket preventing use of the addon if your inventory is full.

I agree with that.

z16:
So how should we proceed? It's a valid argument, do you want to just close this PR, or do you want to see if you can check for the inventory message?

ianandersonlol:
I'll look into the message tonight, if it doesn't work reliably I'll close the PM. I _do_ think there should be a check for full inventories, but there are probably far more people who would be pissed about not being able to open bead pouches at full inventory than people who don't feel comfortable about it spamming a "please sort your inventory" message.

PR #1951, opened 2020-10-12
[battlemod] add crafting display toggle

posimagi:
Added the ability to disable the early display of crafting results that battlemod adds, for people who prefer a more "stock" feel while maintaining the benefits of battlemod.

Also cleaned up README formatting.


KenshiDRK (changes requested):
There is a minor error on my last pull request on line 102/105 targetname should be condensetargetname, could you make me a favor and fix it for me? Can add it to the readme if you want too.

posimagi:
Done. Also added the rest of the toggles to the README.

PR #1957, opened 2020-11-01
add alternate output mode to scoreboard

posimagi:
Added an additional display mode (disabled by default, enabled with `//sb set alternateoutput true`) which minimizes pet/skillchain headers to leave more space for player names, and reports only one entity per line of output to aid legibility.

image

image


posimagi:
Replaced with cleaner PR #1958

PR #1958, opened 2020-11-01
add alternate output mode for scoreboard

posimagi:
Added a mode to scoreboard (disabled by default; enabled with `//sb set alternateoutput true`) which minimizes skillchain/pet headers and maximizes player names, as well as reporting one entity per line for better legibility.

image

image

### DIFF
--- a/addons/scoreboard/display.lua
+++ b/addons/scoreboard/display.lua
@@ -230,7 +230,7 @@ end
 
 -- Takes a table of elements to be wrapped across multiple lines and returns
 -- a table of strings, each of which fits within one FFXI line.
-local function wrap_elements(elements, header, sep)
+local function wrap_elements(elements, header, alt, sep)
     local max_line_length = 120 -- game constant
     if not sep then
         sep = ', '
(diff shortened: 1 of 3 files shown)


posimagi:
Any feedback?

z16:
I have a number of comments, but haven't reviewed it in detail yet. I will try to work through this by this weekend. One thing I can already mention is that I feel a bit apprehensive about combining the various things it does into one vague setting called "alternate output". I think it would be nicer if these settings could be more granular, like one for combining pets, one for combining skillchains, one for reporting one entity per line, etc. If this is something you feel like doing, feel free to, otherwise it will be a few more days before I can review it in detail.

posimagi:
I may be able to do that, probably over the weekend. Thanks.

z16:
Ok, if you're willing to do that I will hold off on the review for now. From a cursory review it looked fine anyway. Let me know when you have more for me to look at, as I don't think I'll get notified by new commits.

posimagi:
Testing the changes, will comment here when they're verified.

posimagi:
Verified. Ready for review.

posimagi:
@z16 Ping. Let me know if you need anything else.

addons/scoreboard/display.lua
```diff
@@ -241,31 +241,43 @@ local function wrap_elements(elements, header, sep)
...
+        return baked_lines
+    else
+        header_line = T{}
+        header_line:append(header)
+        lines:append(header_line)
+        while i <= #elements do
             current_line = T{}
             lines:append(current_line)
-            new_line_length = elements[i]:len() + sep:len()
+            current_line:append(elements[i])
```
z16, line 275: The `current_line` variable is unnecessary here. The `while` body can be refactored to a single line and the variable can be made local to the first `if` case.

z16 (approved):
Hey, really sorry for the delay, but here's the review. Everything looks good, one minor suggestion which is really optional. Let me know if you still want to tackle that or if you consider this done.

PR #1959, opened 2020-11-07
[Dynamis Helper] Fix, Enhance & Refactor

loicteixeira:
4c18a13 has all the original work and requested changed from #1519.
a7794d4 splits the data from the code (as a separate commit so it can be removed if you don't like the idea).

The new interface with 4 vertical sections: time remaining, trigger type, currency drops and time extensions

FWIW, I observed one issue, when you target a monster with no trigger (or a player) it will show the trigger type of the previous monster. But tbh, the infobar addon has the same issue and the dynamishelper addon currently on master is non-functionnal so the new version is already a big improvement.


addons/dynamishelper/dynamishelper.lua
```diff
@@ -24,408 +24,427 @@
...
 -- Proc identifier
--- Lot currency
+-- Light Luggage Profile
 
 _addon.name = 'DynamisHelper'
-_addon.author = 'Krizz, maintainer: Skyrant'
+_addon.author = 'Krizz, Skyrant'
 _addon.commands = {'DynamisHelper','dh'}
-_addon.version = '1.0.2.0'
+_addon.version = '2.3'
```
z16: Feels like a bit of an arbitrary jump, not that I mind :)

addons/dynamishelper/dynamishelper.lua
```diff
@@ -24,408 +24,427 @@
 --(INCLUDING NEGLIGENCE OR OTHERWISE) ARISING IN ANY WAY OUT OF THE USE OF THIS
 --SOFTWARE, EVEN IF ADVISED OF THE POSSIBILITY OF SUCH DAMAGE.
 --Features
```
z16, line 27: This is not your fault as it has been there before, but add a spacer line after the copyright disclaimer. That took me a while to mentally parse.

addons/dynamishelper/dynamishelper.lua
```diff
@@ -24,408 +24,427 @@
...
-								"Vanguard Armorer", "Vanguard Dragontamer", "Vanguard Ronin", "Vanguard Smithy",
-								"Buffrix Eargone", "Cloktix Longnail", "Sparkspox Sweatbrow", "Ticktox Beadyeyes", "Tufflix Loglimbs", "Wyrmwix Snakespecs", "Karashix Swollenskull", "Smeltix Thickhide", "Wasabix Callusdigit", "Anvilix Sootwrists", "Scruffix Shaggychest", "Tymexox Ninefingers", "Scourquix Scaleskin",
-								"Draklix Scalecrust", "Moltenox Stubthumbs", "Ruffbix Jumbolobes", "Shisox Widebrow", "Tocktix Thinlids",
-							 	"Nightmare Crawler", "Nightmare Raven", "Nightmare Uragnite", 
-							 	"Nightmare Fly", "Nightmare Flytrap", "Nightmare Funguar", 
-							 	"Nightmare Gaylas", "Nightmare Kraken", "Nightmare Roc", 
-							 	"Nightmare Hornet", "Nightmare Bugard", 
-								"Shamblix Rottenheart", "Elvaansticker Bxafraff", "Qu'Pho Bloodspiller", "Te'Zha Ironclad", "Koo Rahi the Levinblade", 
-								"Barong", "Alklha", "Stihi", "Fairy Ring", "Stcemqestcint", "Stringes", "Suttung" }
+ProcZones = res.zones:english(string.startswith-{'Dynamis'}):keyset()
```
z16: You may be excused for thinking we don't care much for naming seeing how the entire repository is littered with inconsistencies, but we have been trying to enforce it for a while now. For settings names you're free to use *PascalCase*, otherwise please stick to *snake_case*. This was not your doing as it has existed before, but if you're in the process of refactoring it I would like to ask you to adjust this as well, but it's up to you.

addons/dynamishelper/dynamishelper.lua
```diff
@@ -24,408 +24,427 @@
...
+-------------------------------------------------------------------------------
+-- Define default values ------------------------------------------------------
+-------------------------------------------------------------------------------
```
z16: Also not a huge issue, but one line comments are already too much and are only wasting space, three lines is just showing off :) Goes for the other comments, too. Remove the header/footer comments, and preferably also the content comment, though I will leave that one up to you.

addons/dynamishelper/dynamishelper.lua
```diff
@@ -24,408 +24,427 @@
...
+settings = config.load(defaults)
+
+-- upgrade to new settings file if old settings still exist -------------------
+if settings.trposx and settings.trposy then
+    -- copy position from old config to new one
+    settings.window.pos.x = settings.trposx
+    settings.window.pos.y = settings.trposy
+    -- delete old settings here -----------------------------------------------
+    config.save(settings, 'all')
```
z16: This is quite dangerous and a deal breaker for this PR until it is adjusted. It would overwrite all user settings for all characters. We don't have a mechanism to update settings for all characters, so either you have to read the XML manually and rewrite it before load, or you just update one character at a time, leaving the XML in an inconsistent state. While it sounds bad it shouldn't be an issue, because each character section should be overwritten once the addon is loaded for that character.

I'm just not 100% sure if existing settings can be removed from the XML, but I think it should be possible by just setting them to `nil`. Even if not, it's not a massive deal, because they will just be ignored in the addon.

addons/dynamishelper/dynamishelper.lua
```diff
@@ -24,408 +24,427 @@
...
-							 	"Vanguard Ambusher", "Vanguard Hitman", "Vanguard Pathfinder", "Vanguard Pit", "Vanguard Welldigger",
-							 	"Bandrix Rockjaw", "Lurklox Dhalmelneck", "Trailblix Goatmug", "Kikklix Longlegs", "Snypestix Eaglebeak", "Jabkix Pigeonpecs", "Blazox Boneybod", "Bootrix Jaggedelbow", "Mobpix Mucousmouth", "Prowlox Barrelbelly", "Slystix Megapeepers", "Feralox Honeylips", 
-							 	"Bordox Kittyback", "Droprix Granitepalms", "Routsix Rubbertendon", "Slinkix Trufflesniff", "Swypestix Tigershins",
-								"Nightmare Crab", "Nightmare Dhalmel", "Nightmare Scorpion", 
-								"Nightmare Goobbue", "Nightmare Manticore", "Nightmare Treant", 
-								"Nightmare Diremite", "Nightmare Tiger", "Nightmare Raptor", 
-								"Nightmare Leech", "Nightmare Worm", 
-							 	"Woodnix Shrillwhistle", "Hamfist Gukhbuk", "Lyncean Juwgneg", "Va'Rhu Bodysnatcher", "Doo Peku the Fleetfoot",
-							 	"Nant'ina", "Antaeus"}
+Green = "\\cs(0,255,0)"
```
z16: Please use single quotes for all strings where it makes sense. We allow exceptions for log output or item names, where apostrophes may be regularly encountered, otherwise we try to be consistent here. Also, please `snake_case` the variable name.

addons/dynamishelper/dynamishelper.lua
```diff
@@ -24,408 +24,427 @@
...
-proc = "off"
-trposx = 1000
-trposy = 250
-pposx = 800
-pposy = 250
+-------------------------------------------------------------------------------
+-- Initialize the Currencies array. We need this to keep track of the drops -----
+-------------------------------------------------------------------------------
+function init_currency()
+	for currency in Currencies:it() do
```
z16: A tab managed to sneak in.

addons/dynamishelper/statics.lua
```diff
@@ -0,0 +1,1005 @@
+-------------------------------------------------------------------------------
+-- Currencies -----------------------------------------------------------------
+-------------------------------------------------------------------------------
+
+Currencies = T {
```
z16: You're using `T`-tables for everything here, but I don't know if many of them even need it. Those are always associated with a performance cost. It is a pretty minor issue since the performance cost is *mostly* on the initialization and these here are only initialized once. `Currencies` is iterated with `:it()` above, so it actually makes use of it, this was just meant as a comment on the file in general.

addons/dynamishelper/dynamishelper.lua
```diff
@@ -24,408 +24,427 @@
...
+    end
+end
+init_granules()
+
+-------------------------------------------------------------------------------
+-- The on screen window structure ---------------------------------------------
+-------------------------------------------------------------------------------
+function init_window()
+    local showCurrenciesDivider = false
+    window.text(window,Yellow.."Time remaining: ${time|initializing...}")
```
z16: While this is technically fine, it's meant to be called like this:
```lua
window:text(Yellow .. 'Time remaining: ${time|initializing...}')
```

If you call it the way you do now you can call the library function directly:
```lua
texts.text(window, Yellow .. 'Time remaining: ${time|initializing...}')
```

This saves one indirection. Making a function taking the associated object as the first parameter is commonly used in Lua to enable method-like syntax with `object:method(params)`.

addons/dynamishelper/dynamishelper.lua
```diff
@@ -24,408 +24,427 @@
...
+window = texts.new(" ",settings.window,settings)
+w = T{}
```
z16: Bit of a confusing variable name. What is this table meant to contain? Window-specific information? Also, seems like the `T` is unnecessary here.

addons/dynamishelper/dynamishelper.lua
```diff
@@ -24,408 +24,427 @@
...
 
-settings = config.load()
-timer = settings['timer']
-tracker = settings['tracker']
-trposx = settings['trposx']
-trposy = settings['trposy']
-proc = settings['proc']
-pposx = settings['pposx']
-pposy = settings['pposy']
+window:hide()
```
z16: This is fine, just feels a bit out of place. I would move this line to where the window is created, to make it obvious that it's meant to be invisible from the start. Although I thought they are invisible by default, but not sure.

addons/dynamishelper/dynamishelper.lua
```diff
@@ -24,408 +24,427 @@
...
 
-for i=1, #Currency do
-     Currency[Currency[i]] = 0
+-------------------------------------------------------------------------------
+-- Check if we are in Dynamis and show the overlay ----------------------------
+-------------------------------------------------------------------------------
+if ProcZones:contains(windower.ffxi.get_info().zone) then
```
z16: I feel like this `if` block can be removed and the `zone change` and `load` and `login` events should be merged. Would save on a bit of code duplication and also put the initialization logic all in one place, since they are all used to determine initialization.

addons/dynamishelper/dynamishelper.lua
```diff
@@ -24,408 +24,427 @@
...
+-- 134  Dynamis - Beaucedine --------------------------------------------------
+--  40  Dynamis - Buburimu ----------------------------------------------------
+-- 188  Dynamis - Jeuno -------------------------------------------------------
+--  41  Dynamis - Qufim -------------------------------------------------------
+-- 185  Dynamis - San d'Oria --------------------------------------------------
+--  42  Dynamis - Tavnazia ----------------------------------------------------
+--  39  Dynamis - Valkurm -----------------------------------------------------
+-- 187  Dynamis - Windurst ----------------------------------------------------
+-- 135  Dynamis - Xarcabard ---------------------------------------------------
+-------------------------------------------------------------------------------
```
z16: Not sure what purpose this comment servers? Or just for debugging?

addons/dynamishelper/dynamishelper.lua
```diff
@@ -24,408 +24,427 @@
...
+    local time = nil
+    local item = nil
+    local fiend = original:match("%w+'s attack staggers the (%w+)%!")
+    if settings.timer then
+        if fiend == 'fiend' then
+            windower.send_command('timers c "'..current_mob..'" 30 down stun')
+            return new, color
+        end
+    end
+    if string.find(original,"remaining in Dynamis.") then
```
z16: There are just a few inconsistencies in method calls. The original addon uses `string.find(original, 'something')` whereas you seem to use `original:match('something')`, but left the old code in the old style, which results in mixed invocation styles within the file. I would recommend to use the latter style for every call, but I'll leave this up to you.

Style aside, I would suggest to match the strings as closely as possible. This would also match if a player wrote "Not much longer remaining in Dynamis. We should hurry.". I would check the original mode (called `color` here) as well as matching the text as tightly as possible. I don't know the exact message, but iirc it goes something like "30 minutes remaining in Dynamis.", which I would parse with `original:endswith("%d+ minutes remaining in Dynamis.")` or something like that. If you need to alternate between `minutes` and `hours` you can use our regex matching library (not sure if the game reports hours as well, the code afterwards suggests it does not, just mentioned it in case I'm not remembering it correctly).

This goes for the next few `string.find` checks as well.

addons/dynamishelper/dynamishelper.lua
```diff
@@ -24,408 +24,427 @@
...
+        obj_time = (tonumber(original:match("%d+")) * 60)
+        end_time = os.time() + obj_time
+    end
+    if string.find(original,"will be expelled from Dynamis") then
+        obj_time = (tonumber(original:match("%d+")) * 60)
+        end_time = os.time() + obj_time
+    end
+    if string.find(original,"Your stay in Dynamis has been extended by %d+ minutes.") then
+        end_time = end_time + (tonumber(original:match("%d+")) * 60)
+        w.time = os.date('!%H:%M:%S', end_time)
```
z16: I notice this case adjusts `w.time`, but the others do not. This one does not adjust `ob_time` though. Is that correct?

addons/dynamishelper/dynamishelper.lua
```diff
@@ -24,408 +24,427 @@
...
+    end
+    item = original:match("Obtained key item: ..(%w+ %w+ %w+ %w+)..\46")
+    if item ~= nil then
+        item = item:lower()
```
z16: This is also not your code, but there's really no reason to use `:lower()` on both strings. We get them directly from the game, so they should match case as well.

addons/dynamishelper/dynamishelper.lua
```diff
@@ -24,408 +24,427 @@
...
+    --a,b,item = string.find(original,"%w+ obtains an? ..(%w+ %w+ %w+ %w+)..\46")
+    if item == nil then
+        item = original:match("%w+ obtains an? ..(%w+ %w+ %w+)..\46")
+        --a,b,item = string.find(original,"%w+ obtains an? ..(%w+ %w+ %w+)..\46")
+        if item == nil then
+            item = original:match("%w+ obtains an? ..(%w+%-%w+ %w+)..\46")
+            --a,b,item = string.find(original,"%w+ obtains an? ..(%w+%-%w+ %w+)..\46")
+            if item == nil then
+                item = original:match("%w+ obtains an? ..(%w+ %w+)..\46")
+                --a,b,item = string.find(original,"%w+ obtains an? ..(%w+ %w+)..\46")
```
z16: While I like tight matching, this may be overdoing it :) If you want to catch all those individual cases you could as well just check for the individual items directly:
```lua
if original:match('%w+ obtains a 100 Byne Bill') or
   original:match('%w+ obtains an Ordelle Bronzepiece') or
   -- and so on
```

And in that case you might as well use your `Currencies` table:
```lua
if Currencies:any(function(currency) return original:match('%w+ obtains an? ..' .. currency .. '..\46') end) then
```

It does not get tighter than that. But will that not trigger for messages you receive for other characters as well? This is also not just your code, if it is an issue it was there before your refactoring as well. Or is that expected behavior?

This is honestly something that would be better solved using packets, but I guess that's outside of the scope of this refactoring.

Tl;dr I would either just match it more loosely as it was originally, or use exact matching using the table.

addons/dynamishelper/dynamishelper.lua
```diff
@@ -24,408 +24,427 @@
...
+            end
+        end
+    end
+    if item ~= nil then
+        item = item:lower()
```
z16: See above about `:lower()`.

addons/dynamishelper/dynamishelper.lua
```diff
@@ -24,408 +24,427 @@
...
- 		windower.text.set_bg_visibility('dynamis_box',true)
- 		windower.text.set_font('dynamis_box','Arial',12)
- 		windower.text.set_text('dynamis_box',obtained);
- 	end
-end
-
+-------------------------------------------------------------------------------
+-- Register target change event to get the monster name -----------------------
+-------------------------------------------------------------------------------
 windower.register_event('target change', function(targ_id)
```
z16: Also not your thing, but this variable name is wrong. It should be `targ_index` (or preferably `target_index`), it's not an ID.

addons/dynamishelper/dynamishelper.lua
```diff
@@ -24,408 +24,427 @@
...
-	 	windower.text.set_bg_visibility('proc_box',1)
-	 	windower.text.set_font('proc_box','Arial',12)
-	 	windower.text.set_text('proc_box',' Current proc for \n '..mob..'\n is '..current_proc);
-	 	if proc == "off" then
-	 		windower.text.set_visibility('proc_box', false)
-	 	end
+-------------------------------------------------------------------------------
+-- Print the help Information -------------------------------------------------
+-------------------------------------------------------------------------------
+help = T{
```
z16: This also does not need to be a `T`-table, and the contents should be indented.

addons/dynamishelper/dynamishelper.lua
```diff
@@ -24,408 +24,427 @@
...
+stopacity = "Usage: dh stopacity [0-100]%% - Stroke opacity is currently at %u%%",
+stcolor = "Usage: dh stcolor [Red] [Green] [Blue] - Stroke color - Example: 255 0 0 is red",
+posx = "Usage: dh posx [x] - Current window position is x=%u",
+posy = "Usage: dh posy [y] - Current window position is y=%u",
+ll = "Usage: dh ll create: Creates a light luggage profile to lot all dynamis currency.'"
+}
+
+function printHelp(command,val)
+    if not command and not val then
+        windower.add_to_chat(159,'\nDynamisHelper v2.0')
```
z16: `2.0` is not quite right. You can just use `'DynamisHelper v' .. tostring(_addon.version)`.

addons/dynamishelper/dynamishelper.lua
```diff
@@ -24,408 +24,427 @@
...
+-------------------------------------------------------------------------------
+-- Process options and save settings ------------------------------------------
+-------------------------------------------------------------------------------
+windower.register_event('addon command',function (command, ...)
+    command = command and command:lower() or 'help'
+    local options = {...}
+    if command == "help" then
+        printHelp()
+        return
+    elseif command:lower() == "visible" then
```
z16: `command` is already lowercased. Also, a reminder about quote style :)

addons/dynamishelper/dynamishelper.lua
```diff
@@ -24,408 +24,427 @@
...
+        if window:visible() then
+            window:hide()
+        else
+            window:show()
+        end
+    elseif command == "size" then
+        if options[1] and tonumber(options[1]) then
+            window:size(options[1])
```
z16, line 300: Wouldn't this need to be `tonumber(options[1])` as well?

addons/dynamishelper/dynamishelper.lua
```diff
@@ -24,408 +24,427 @@
...
+        end
+    elseif command == "font" then
+        if options[1] then
+            window:font(options[1])
+            config.save(settings, 'all')
+        else
+            printHelp(command,settings.window.text.font)
+        end
+    elseif command == "posx" then
+        if options[1] then
```
z16, line 313: I think this needs `tonumber` as well? Same for some of the next ones. Maybe just go over these once more to make sure.

addons/dynamishelper/dynamishelper.lua
```diff
@@ -24,408 +24,427 @@
...
+    elseif command:lower() == "visible" then
+        if window:visible() then
+            window:hide()
+        else
+            window:show()
+        end
+    elseif command == "size" then
+        if options[1] and tonumber(options[1]) then
+            window:size(options[1])
+            config.save(settings, 'all')
```
z16, line 301: Saving for `'all'` should not be done without a good reason. This disables per-character settings. What was the reason for doing this?

addons/dynamishelper/dynamishelper.lua
```diff
@@ -24,408 +24,427 @@
...
+            printHelp(command,window:stroke_transparency()*100)
+        end
+    elseif command == "stcolor" then
+        if options[3] then
+            window:stroke_color(options[1],options[2],options[3])
+            config.save(settings, 'all')
+        else
+            printHelp(command,"")
+        end
+    elseif command == "ll" then
```
z16: I would love to remove this command entirely. LL has been deprecated for six years and the code in here is messy and not particularly useful. If you remove it, I will accept the PR, but I'll leave this up to you.

If you do keep it, no need to get the player name again, it is already set in the `player` variable on load and login.

z16 (changes requested):
Sorry for the delay. I added a number of comments with a few things that may need to be discussed. Feel free to respond to the points I made here or message me on Discord.

loicteixeira:
Unfortunately I cannot answer the why questions as I was merely trying to push the original PR to the finish line by applying the existing comments.

Thank you for your time, I'll make sure it was not wasted and push an update in the coming weeks.

z16:
Hm, that explains a bit. I was meaning to ask why not make it a completely new addon, since it seems to completely rewrite the original. Not that I mind, though I think it would have had a good case for being its own addon.

loicteixeira:
Still working through the changes.

I actually got burned out by the game and cancelled my subscription, hence why I didn't come to this sooner. I still have a week left, so I'll try to take care of the latest required changes (and test them) before then.

As for creating a separate addon, the existing one is non-functional, so I reckon we might as well overwrite it instead of creating a new one.

addons/dynamishelper/dynamishelper.lua
loicteixeira: The 2 cases above are statements of how much time left, not time extensions like this one, so there should be no need to update `w.time`. But it would also mean that there is no need to update the internal objects, which leads me to believe it was made in case the plugin is loaded after entering Dynamis.
Anyways, `obj_time`, `end_time` and `w.time` are all linked, so I'll update them together. I'll also rename `obj_time` to `time_remaining_in_seconds` or something alike.

addons/dynamishelper/dynamishelper.lua
loicteixeira: Realistically, this plugin will be used for solo farming, so there's no risk of it matching another character in this situation.
However, nothing stops you from using it with a group (or while multi-boxing) and in both cases, tracking currencies just for you or for the group (so you have the total amount to split at the end) are both valid cases.
This could be made configurable, but I'm running out of time and will therefore change it to only get your own drops. Someone else can contribute later if they want it configurable.

loicteixeira:
I addressed all the issues, except for not using T-tables.

FWIW, time tracking does not work when the plugin is loaded after entering dynamis, which makes sense because how would the plugin now how much time is left? Aside from that, it tracks drops and displays stagger trickers properly.

Unfortunately my subscription is expiring tomorrow, so I'm afraid I won't be able to make further changes. Feel free to close the PR then unless someone wants to pick it up like I pick up the previous PR.

Happy gaming everybody.

PR #1964, opened 2020-11-17
Fix player pet check

dtw:
Pretty sure that player.pet_index doesn't exist/always returns `nil`? Tested by adding debugs and loading autocontrol with pet already active.

The validation on line 93 is the same as used later.

### DIFF
--- a/addons/autocontrol/maneuver.lua
+++ b/addons/autocontrol/maneuver.lua
@@ -91,7 +91,7 @@ Burden_tb = texts.new(str, settings)
 windower.register_event("action", function(act)
     if mjob_id == 18 then
         local abil_ID = act['param']
-        local actor_index = act['actor_index']
+        local actor_id = act['actor_id']
         local player = T(windower.ffxi.get_player())
         
         local pet_index = windower.ffxi.get_mob_by_index(windower.ffxi.get_player()['index'])['pet_index']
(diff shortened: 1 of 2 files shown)


z16:
I think you're right, but to get the current player please use one of these:
```lua
windower.ffxi.get_mob_by_index(player.index)
windower.ffxi.get_mob_by_target('me')
```
The problem with `windower.ffxi.get_mob_by_id` is that it has to iterate the array to find the right match, so its performance is much worse than the other two options. If you have the option, `windower.ffxi.get_mob_by_index` is the best choice, and `windower.ffxi.get_mob_by_target` is the next best.

dtw:
OK, I'll fix that case and the other cases in a subsequent commit

addons/autocontrol/maneuver.lua
```diff
@@ -91,10 +91,10 @@ Burden_tb = texts.new(str, settings)
 windower.register_event("action", function(act)
     if mjob_id == 18 then
         local abil_ID = act['param']
-        local actor_id = act['actor_id']
+        local actor_index = act['actor_index']
```
dtw: Have not check `actor_index` exists but I assume it does.

dtw:
I have not tested this in game yet

addons/autocontrol/maneuver.lua
z16: It does not.

dtw: Shucks.

dtw:
This doesn't work at all as expected in testing.

PR #1969, opened 2020-11-19
Update traits.lua

### LINKED ISSUE #1967: bluguide missing traits
I noticed 3 missing traits in bluguide. The missing traits are Evasion bonus, Resist Silence, and Resist Slow. These are listed in the traits.lua but for some reason not showing up in the UI. I looked at all the code to see if I could fix it but it's too advanced for me. I also noticed voracious trunk has the wrong id in traits.lua. I changed mine from 576 to 579 and voracious trunk works as intended now but still missing the 3 traits.

Cellcubed:
ID for voracious trunk was incorrect changed from 576 to 579

### DIFF
--- a/addons/bluguide/res/traits.lua
+++ b/addons/bluguide/res/traits.lua
@@ -36,7 +36,7 @@ return {
             { name = "Self-Destruct", points = 2, cost = 3, id = 533 },
             { name = "Cold Wave", points = 1, cost = 1, id = 535 },
             { name = "Light of Penance", points = 2, cost = 5, id = 634 },
-            { name = "Voracious Trunk", points = 3, cost = 4, id = 576 },
+            { name = "Voracious Trunk", points = 3, cost = 4, id = 579 },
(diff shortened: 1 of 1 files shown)


RubenatorX:
Just a tangential note here: we actually added the blue point cost to Windower resources, although we have not added how many points the spell adds toward a trait.. not sure if we want to do that or not.

Cellcubed:
Yeah I don't know if that would be necessary, the blu specific addons seem to be doing just fine without that info.

Cellcubed:
I'm very new to github and only use it for windower so far. Am I doing it the right way? I found an issue, i think i fixed the issue, i edited the code on github, proposed a change, and then click submit pull request. I guess what I'm asking is, do I need to hit pull request? Is just hitting propose change sufficient to get it looked at and possibly ready for other people to download via windower?

Nifim:
You did it all correct, someone will review it and merge or request changes.

Which I just did, by verifying your change against the resources repo

PR #1971, opened 2020-11-23
Autocontrol v2.0 dtw

dtw:
This fixes the original bug I found


Nifim:
Looks like this change needs to be applied in the `burdometer` and `autoabils` files as well.

PR #1983, opened 2020-12-28
Fix use of quotes in `gs equip` and exit loop properly

RubenatorX:
**_SHOULD PROBABLY HAVE MULTIPLE PEOPLE TEST THIS BEFORE IT GOES TO LIVE_**
This is intended to allow quotes to be omitted when using brackets`[]` in the equip command. This allows the command to function properly when double quotes are used (because the windower command parser omits them).
The following should now all work:
```
gs equip precast.WS["Rudra's Storm"]
gs equip precast.WS['Rudra's Storm']
gs equip precast.WS[Rudra's Storm]
gs equip precast.WS.Rudra's Storm
```
While I'm at it, the function also would only ever exit after 30 iterations. This is now fixed.
Also corrected some inconsistent use of newlines. Because reasons.

### DIFF
--- a/addons/GearSwap/helper_functions.lua
+++ b/addons/GearSwap/helper_functions.lua
@@ -42,6 +42,7 @@ function string.lower(message)
     end
 end
 
+
 -----------------------------------------------------------------------------------
 --Name: string.upper()
 --Args:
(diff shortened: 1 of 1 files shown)


z16:
Idk if anyone tested this, but people will complain loudly enough if something within GS breaks.

PR #1991, opened 2021-01-23
Updated to version 2.0

lili-ffxi:
Added resize and reset commands, implemented IPC, cleaned up code.

### DIFF
--- a/addons/position_manager/position_manager.lua
+++ b/addons/position_manager/position_manager.lua
@@ -26,7 +26,7 @@
 
 _addon.name = 'position_manager'
 _addon.author = 'Lili'
-_addon.version = '2.0'
+_addon.version = '2.0.1'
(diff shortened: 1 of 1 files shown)


z16:
Most of it looks good, the only thing I'm a bit unhappy about is the liberal use of `windower.ffxi.get_player` and `windower.ffxi.get_info`, which are slightly expensive. It only happens on command usage, so I don't mind too much and I'd be willing to merge this PR as is, but it may be nicer to keep track of login and logout and populate the fields that way (you already have a `player_name` variable for that very purpose). Let me know if you feel like changing it, otherwise I'll merge this as is.

lili-ffxi:
You're always so diplomatic.

That stuff was a leftover of the older code, I was happy to get rid of it. See the new changes.

PR #1994, opened 2021-01-24
Create Drop Addon / Stubborn (CFHProtect Plugin Replacement)

ianandersonlol:
Simple, lightweight addon to drop items from your inventory. 

Since I know for a lot of people the go-to method to drop something is //tr drop add <item> and then immediately doing //tr drop remove <item>. I think an addon to drop something on command is very helpful.


ianandersonlol:
I suggested another addon (A replacement to the old CFHprotect plugin) commit 59fb803

addons/Drop/drop.lua
```diff
@@ -0,0 +1,65 @@
...
+
+check_inv = function(item_id)
+        for k, v in pairs(windower.ffxi.get_items().inventory) do
+            if type(v) == 'table' and v.id == item_id then
+                return true
+            end
+        end
+    return false
+end
+get_item_resource = function(item)
```
z16, line 20: This is a very slow function, and unnecessary. *If* you decide to keep it, call `item:lower()` only once at the start and use the resulting variable. The way it is now you create a lowercase string of it up to 44 thousand times. Also, Japanese does not have casing, and calling `v.japanese:lower()` could actually mess up the resulting string.

But like I said, this entire function is unnecessary. You don't need to iterate all the resources, with its 22 thousand items. Instead it's enough to iterate the inventory of max 80 items, get the item's ID and use the resources to look up by ID, which is fast. Then you can compare the `english` and `japanese` of the found item (if it exists) with the user's string. You can even enable wildcard matching like this, so the user can do something like `//drop * crystal` and have it drop all crystals.

addons/Drop/drop.lua
```diff
@@ -0,0 +1,65 @@
...
+    for k, v in pairs(res.items) do
+        if v.english:lower() == item:lower() or v.japanese:lower() == item:lower()  then
+            return v
+        end
+    end
+    return nil
+end
+
+drop_item = function(item_to_drop)
```
z16, line 29: This function behaves almost the same as the `check_inv` function, only the loop body is different. I would skip the `check_inv` function entirely and call `drop_item` directly. This function can still return `true` or `false` based on if it actually dropped something. You can even combine that with `coroutine.sleep`, since that function also returns nothing, which evaluates to `false`. This makes the function design and invocation both easier.

addons/Drop/drop.lua
```diff
@@ -0,0 +1,65 @@
...
+_addon.commands = {'drop'}
+
+
+packets = require('packets')
+require('luau')
+res = require('resources')
+require('logger')
+
+check_inv = function(item_id)
+        for k, v in pairs(windower.ffxi.get_items().inventory) do
```
z16, line 13: `get_items` is a *very* expensive function call (the most or second most expensive one in our API, depending on the situation), and you call it both here and in `drop_item`. You should call it before either and pass it in to both. That said, you can actually condense the logic into one function, then you wouldn't need to do it, see below.

Also, if you only want the inventory pass it in as a parameter:
```lua
windower.ffxi.get_items('inventory') -- the name of the bag
-- or
windower.ffxi.get_items(0) -- or the ID
```

addons/Drop/drop.lua
```diff
@@ -0,0 +1,65 @@
...
+    end
+end
+
+windower.register_event('addon command', function (...)
+    args = {...};
+    if args[1]:lower() == 'help' then
+        log('//drop <\30\02item\30\01>\nDrops a specific \30\02item\30\01. Does not require quotes, capitalization, and accepts auto-translate.') 
+        return
+    end
+    for i, v in pairs(args) do args[i]=windower.convert_auto_trans(args[i]) end --thanks Akaden
```
z16, line 52: Is there an issue with calling it on the entire string? Not sure why you're doing it on the individual tokens.

addons/Stubborn.lua
```diff
@@ -0,0 +1,32 @@
+_addon = {'Stubborn'}
```
z16, line 1: This line does nothing, it can be removed.

addons/Stubborn/readme.md
```diff
@@ -0,0 +1,7 @@
+# Stubborn
+A Windower 4 addon to prevent unnecessary "Call for help!"
+
+Stubborn is a light weight replacement to the old CFHProtect addon. 
+
+
+If you need to call for help. Type either //stubborn or //lua c stubborn.
```
z16, line 7: No need to mention `//lua c stubborn`, that's not an encouraged way to call commands. It's meant to be used as an API.

z16 (changes requested):
I mentioned a number of things in the individual files, but there are a number of overarching issues.

1. Please close this PR and open two new ones, one for each addon.
2. Add a copyright notice to each file, as described [here](https://github.com/Windower/Lua/wiki/Writing-Addons#licensing).
3. The addons are missing entries in the `addons.xml` file at the root of the folder structure. If they are missing they will not show up in the launcher.
4. While our style requirements for v4 are quite lax compared to v5, we do want the indentation to be correct, the type of quotes throughout the addon to be consistent (preferably single quotes), use of spaces after commas, and generally not squish too much (like `if ... then ... end`) into one line, to keep everything readable. Please adjust those points in each of the addons.

PR #1998, opened 2021-02-09
Packets fields.lua incoming `0x017` and `0x01B` for Assist Channels

RubenatorX:
__This won't work properly in packetviewer till this ResourceExtractor pull request is pushed through to Resources__
https://github.com/Windower/ResourceExtractor/pull/71

`incoming 0x017` changed to have different data for different chat modes
Also tweaked `incoming 0x01B` for Mastery Rank and Mentor icon data

Also small correction on some unknown/junk data I did recently


Chiaia:
Resources have already been updated.

PR #1999, opened 2021-02-09
Update FastCS.lua

ianandersonlol:
Add the unity warp NPCs to the exclusion list(currently, only home points books and waypoints are excluded)


Chiaia:
Please submit into the dev branch any future addons or changes to addons.

ianandersonlol:
oops

PR #2001, opened 2021-02-13
Fixing errors on Linkshell/Linkpearls/Pearlsacks

lili-ffxi:
This change fixes two errors that happen when parsing Linkshells with certain combinations of characters that result in a bitpacked LS name that contains zeroed bytes in the middle.
First change fixes the `nil` error, second change detects the length of the name properly.

### DIFF
--- a/addons/libs/extdata.lua
+++ b/addons/libs/extdata.lua
@@ -1481,7 +1481,7 @@ function tools.bit.l_to_r_bit_packed(dat_string,start,stop)
     local c_count = math.ceil(stop/8)
     while c_count >= math.ceil((start+1)/8) do
         -- Grabs the most significant byte first and works down towards the least significant.
-        local cur_val = dat_string:byte(c_count)
+        local cur_val = dat_string:byte(c_count) or 0
         local scal = 1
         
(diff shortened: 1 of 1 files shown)


lili-ffxi:
A better/more elegant/leaner way of doing it is welcome. This was the most basic way I could come up with.

z16:
> A better/more elegant/leaner way of doing it is welcome. This was the most basic way I could come up with.

If you're sure there is at least one non-null byte (which makes sense, I think? Considering it's only parsed in case of an LS name) you can do this:
```lua
local name_end = #str
while str:byte(name_end) == 0 do
    name_end = name_end - 1
end
```

If you want to make sure it's at least 10 you can add `and name_end > 10` to the loop's condition.

lili-ffxi:
I famously forget that `while` and `until` exist, especially at 4am.

lili-ffxi:
Done. I put the > 10 in, should never actually be needed but I thought it's maybe better safe than sorry.

z16:
> I famously forget that `while` and `until` exist, especially at 4am.

They are usually worth forgetting.

PR #2003, opened 2021-02-13
Dtw macro changer

dtw:
The add_to_chat chatmode is currently "tell"

### DIFF
--- a/addons/macrochanger/macrochanger.lua
+++ b/addons/macrochanger/macrochanger.lua
@@ -139,7 +139,9 @@ windower.register_event('job change',function ()
       notice('Auto Macro Switching Disabled for ' .. job ..'.')
     else
       log('Job changed to ' .. job .. ' - switched to Book: ' .. book .. ' and Page: ' .. page)
-      windower.send_command('input /macro book '..book..';wait 0.2;input /macro set '..page..';')
+      windower.chat.input('/macro book ' .. book)
+      coroutine.sleep(0.2)
+      windower.chat.input('/macro set ' .. page)
     end
   elseif globaldisable == 1 then
     notice('Auto Macro Switching Disabled for All Jobs.')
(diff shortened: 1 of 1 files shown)


addons/macrochanger/macrochanger.lua
```diff
@@ -135,14 +136,14 @@ windower.register_event('job change',function ()
 		end
 
 		if ((book == 'disabled') or (page == 'disabled')) then
-			windower.add_to_chat(17, '                             Auto Macro Switching Disabled for ' .. job ..'.')
+			notice('Auto Macro Switching Disabled for ' .. job ..'.')
 		else
-			windower.add_to_chat(17, '                             Changing macros to Book: ' .. book .. ' and Page: ' .. page .. '.  Job Changed to ' .. job)
+			log('Job changed to ' .. job .. ' - switched to Book: ' .. book .. ' and Page: ' .. page)
 			windower.send_command('input /macro book '..book..';wait 0.2;input /macro set '..page..';')
```
z16: I know this is not your doing, but you can change this line to this:
```lua
windower.chat.input('/macro book ' .. book)
coroutine.sleep(0.2)
windower.chat.input('/macro set ' .. page)
```

Up to you, but I find more organized lines like these better to read.

addons/macrochanger/macrochanger.lua
```diff
@@ -153,16 +154,16 @@ windower.register_event('addon command', function(...)
 	if args[1] == 'disableall' then
 		if args[2] == 'on' then
 			globaldisable = 1
-			windower.add_to_chat(17, 'All automated macro switching disabled.')
+			warning('All automated macro switching disabled.')
```
z16: I would also make this a notice, it's not dangerous behavior, nor can it easily be triggered accidentally.

z16 (approved):
It looks good, I made some notes, and I would also ask you to change all the tabs to four spaces. This was not your doing, it's been like that since it was originally pushed, but since you're modifying it already you have the opportunity :)

PR #2010, opened 2021-03-13
[Battlemod] Singular/Plural addition

KenshiDRK:
Added singular/plural handling.
Added common/proper noum handling.
Added quantity of items for message id 377 (Treasury autostack option sometimes interfere with it).
Some minor changes.


z16 (commented):
I'm finding it rather hard to review this, as it's hard to test without going into detail or being very familiar with chat behavior. One thing I  will say is that there are quite a few performance traps in here. `gsub` is, by itself, an expensive function, and calling it 15 times in a row is atrocious for performance.

Since you're often calling it on the `${}` pattern one thing you can do is create a lookup table that maps the values and pass that in as the second parameter. That way you can condense multiple `gsub` calls into one. For example, this:
```lua
:gsub('${actor} hits ', '${actor} hit ')
:gsub('${actor} casts ', '${actor} cast ')
:gsub('${actor} starts ', '${actor} start ')
...
```
could be changed to this:
```lua
local replacements = {
    ['${actor} hits '] = '${actor} hit ',
    ['${actor} casts '] = '${actor} cast ',
    ['${actor} starts '] = '${actor} start ',
    ...
}

-- later
msg:gsub('${actor} (%w+) ', replacements)
```
This will not work for everything, but would cover many replacements and would significantly reduce the `gsub` calls.

Even with that this will be a very expensive operation and I honestly don't think it should be the default. I see that most of it is hidden behind the `not simplify` setting. If I'm reading it right that is set to `true` by default, right? If so, this should be fine. Just making sure.

z16:
Additionally there is some inconsistency with single vs double quotes. Adjust it to single quotes where possible.

KenshiDRK:
yea I was also worried for the massive calls in gsub, and I been thinking into map the messages, so only the needed gsub calls are used, would this be better for performance than your suggestion?

z16:
Yeah, it would probably be better, but also more effort. If you don't mind doing that, go ahead.

And again, just to make sure because I find the code very confusing (the original BM code, not yours), the `simplify` setting is set to `true` by default, right?

KenshiDRK:
yes simplify is set to true by default

PR #2013, opened 2021-03-19
Add equipviewer addon

RubenatorX:
Still working out a few minor things about crediting people so, don't merge yet but, wanted to get the code review process started. Thank you!


addons/equipviewer/equipviewer.lua
```diff
@@ -0,0 +1,734 @@
...
+    [15] = {slot_name = 'back',       slot_id = 15, display_pos = 12, item_id = 0, image = nil},
+}
+local encumbrance_data = {}
+for i=0,15 do
+    encumbrance_data[i] = { slot_name = 'encumbrance', slot_id = i, display_pos = equipment_data[i].display_pos, image = nil }
+end
+local ammo_count_text = nil
+local bg_image = nil
+
+local defaults = T{
```
z16: I don't think any of the `T` here are necessary, are they?

addons/equipviewer/equipviewer.lua
```diff
@@ -0,0 +1,734 @@
...
+    hide_on_cutscene = true,
+    left_justify = false,
+}
+settings = nil
+local last_encumbrance_bitfield = 0
+
+-- gets the currently equipped item data for the slot information provided
+local function get_equipped_item(slotName, slotId, bag, index)
+    if not bag or not index then -- from memory
+        local equipment = windower.ffxi.get_items().equipment
```
z16: This can be sped up significantly by using `get_items('equipment')`.

addons/equipviewer/equipviewer.lua
```diff
@@ -0,0 +1,734 @@
...
+local function update_equipment_slots(source)
+    for slot in pairs(equipment_data) do
+        update_equipment_slot(source, slot)
+    end
+end
+
+-- Sets up the image and text ui objects for our equipment
+local function setup_ui()
+    refresh_ui_settings()
+    destroy()
```
z16, line 185: This only works because `refresh_ui_settings` and `destroy` are global functions. Which isn't bad, but if you wanted to make it `local` (since many of the other helper functions are), you would need to move those ahead of this function. I'm not sure what the criteria are for making some of them local but not the others, if it's because of accessibility, then moving them ahead of these is the solution.

addons/equipviewer/equipviewer.lua
```diff
@@ -0,0 +1,734 @@
...
+    end
+    display_encumbrance()
+
+    ammo_count_text = texts.new(settings.left_justify and ammo_count_text_settings_left_justify or ammo_count_text_settings)
+    display_ammo_count()
+end
+
+-- Called when the addon is first loaded.
+windower.register_event('load', function()
+    if settings then
```
z16: This `if` check is suspicious. It should always be one or the other, as the code path leading up to this is fully deterministic. Did you experience both cases?

That said, I'd probably load the settings outside of this event. Not because of any advantage, just because this is how it's commonly done.

addons/equipviewer/equipviewer.lua
```diff
@@ -0,0 +1,734 @@
...
+
+-- Called whenever character logs out.
+windower.register_event('logout', function()
+    clear_all_equipment_slots()
+    destroy()
+end)
+
+-- Called whenever character logs in.
+windower.register_event('login', function()
+    if settings then
```
z16: This `if` check is also suspicious, for a slightly different reason. This must always be `true`, because the `load` event is always called before `login`, and if `settings` is `nil` there it will be populated. After that it should never be `nil` again.

addons/equipviewer/equipviewer.lua
```diff
@@ -0,0 +1,734 @@
...
+        for _,slot_data in pairs(equipment_data) do
+            if slot_data.bag_id == bag and slot_data.index == index then
+                slot = slot_data.slot_id
+                break
+            end
+        end
+        
+        if slot then
+            if packet['Status'] ~= 5 then --item not equipped
+                update_equipment_slot:schedule(0, '0x%x':format(id), slot, 0, 0, 0)
```
z16, line 258: `'0x%03X'` would be a better formatting string imo as it would result in packet IDs as we specify them in our `fields.lua` file.

addons/equipviewer/equipviewer.lua
```diff
@@ -0,0 +1,734 @@
...
+                return
+            end
+            if slot == 3 then --ammo
+                local count = packet['Count'] or 0
+                display_ammo_count(count)
+            end
+            local item = packet['Item']
+            update_equipment_slot:schedule(0,'0x%x':format(id), slot, bag, index, item, count)
```
z16, line 266: See above.

addons/equipviewer/equipviewer.lua
```diff
@@ -0,0 +1,734 @@
...
+        end
+    elseif id == 0x01B then -- Job Info (Encumbrance Flags)
+        local packet = packets.parse('incoming', original)
+        display_encumbrance(packet['Encumbrance Flags'])
+    elseif id == 0x0A then -- Finish Zone
```
z16, line 271: Just a minor thing for consistency, this should be adjusted to `0x00A` (and below for `0x00B`).

addons/equipviewer/equipviewer.lua
```diff
@@ -0,0 +1,734 @@
...
+windower.register_event('outgoing chunk', function(id, original, modified, injected, blocked)
+    if id == 0x100 then -- Job Change Request
+        clear_all_equipment_slots()
+    end
+end)
+
+
+-- Destroys all created ui objects
+function destroy()
+    if bg_image then
```
z16, line 290: I feel most of these checks could be avoided. Why set the image to `nil` and then re-create them in the `setup` step? Why not just hide them when they're inactive? If it's just to load them with new settings, the image lib *should* support that, at least if it was modeled after the texts lib (which was the idea originally, but I did not write it, so not sure). Did you try that?

Even if that were the case, I think I'd still just hide them and re-create them when they're being set up. Would just avoid a lot of otherwise needless checking, in this function and others like `show` and `hide`.

Would also make the actual destruction and creation obsolete, which would be a minor performance benefit on login/logout.

addons/equipviewer/equipviewer.lua
```diff
@@ -0,0 +1,734 @@
...
+            slot_data.image:show()
+        end
+    end
+    display_encumbrance()
+    display_ammo_count()
+end
+
+-- Hides all ui objects
+function hide()
+    if bg_image then
```
z16, line 328: See above about potentially removing the checks.

addons/equipviewer/equipviewer.lua
```diff
@@ -0,0 +1,734 @@
...
+        log('Size changed to '..settings.size)
+    elseif cmd == 'scale' then
+        if #cmd_args < 1 then
+            error('Not enough arguments.')
+            log('Current scale: '..settings.size/32)
+            return
+        end
+        local size = tonumber(cmd_args[1])*32
+        if size > 100 then
+            error('Size too large')
```
z16, line 459: This complains about the size but then still proceeds. Is it missing a `return`?

addons/equipviewer/equipviewer.lua
```diff
@@ -0,0 +1,734 @@
...
+        config.save(settings)
+
+        setup_ui()
+
+        log('Size changed to '..settings.size)
+    elseif cmd == 'alpha' or cmd == 'opacity' then
+        if #cmd_args < 1 then
+            error('Not enough arguments.')
+            log('Current alpha/opacity: %d/255 = %d%%':format(
+				settings.icon.alpha, math.floor(settings.icon.alpha/255*100)
```
z16: Some tabs sneaked in :)

addons/equipviewer/equipviewer.lua
```diff
@@ -0,0 +1,734 @@
...
+        config.save(settings)
+
+        setup_ui()
+
+        log('Alpha/Opacity changed to '..settings.icon.alpha..'/255')
+    elseif cmd:contains('transpar') then
+        if #cmd_args < 1 then
+            error('Not enough arguments.')
+            log('Current transparency: %d/255 = %d%%':format(
+				(255-settings.icon.alpha), math.floor((255-settings.icon.alpha)/255)*100
```
z16: Tabbity-tab.

addons/equipviewer/equipviewer.lua
```diff
@@ -0,0 +1,734 @@
...
+
+        log('Transparency changed to '..255-settings.icon.alpha..'/255')
+    elseif cmd == 'background' or cmd == 'bg' then
+        if #cmd_args < 1 then
+            error('Not enough arguments.')
+            log('Current BG color: RED:%d/255 GREEN:%d/255 BLUE:%d/255 ALPHA:%d/255 = %d%%':format(
+				settings.bg.red, settings.bg.green, settings.bg.blue, settings.bg.alpha, math.floor(settings.bg.alpha/255*100)
+			))
+            return
+        elseif #cmd_args == 1 then
```
z16, line 512: Using one command should adjust the alpha? That feels a bit unintuitive, imo. I would prefer another option for that singular choice.

addons/equipviewer/equipviewer.lua
```diff
@@ -0,0 +1,734 @@
...
+                    settings.bg.alpha = math.floor(alpha)
+                end
+            end
+        end
+        config.save(settings)
+
+        setup_ui()
+        
+        log('BG color changed to: RED:%d/255 GREEN:%d/255 BLUE:%d/255 ALPHA:%d/255 = %d%%':format(
+			settings.bg.red, settings.bg.green, settings.bg.blue, settings.bg.alpha, math.floor(settings.bg.alpha/255*100)
```
z16: Tabs :D

addons/equipviewer/equipviewer.lua
```diff
@@ -0,0 +1,734 @@
...
+
+        setup_ui()
+
+        log('Ammo text justification changed to '..tostring(settings.left_justify and 'Left' or 'Right'))
+    elseif cmd == 'testenc' then
+        display_encumbrance(0xffff)
+    elseif cmd == 'debug' then
+        if #cmd_args < 1 then
+            local items = windower.ffxi.get_items()
+            local e = windower.ffxi.get_items().equipment
```
z16: `get_items('equipment')`, same as above.

addons/equipviewer/icon_extractor.lua
```diff
@@ -0,0 +1,255 @@
...
+    [2]={min=4096, max=8191, dat_path='118/107', offset=0},
+    [3]={min=8192, max=8703, dat_path='118/110', offset=0},
+    [4]={min=8704, max=10239, dat_path='301/115', offset=0},
+    [5]={min=10240, max=16383, dat_path='118/109', offset=0},
+    [6]={min=16384, max=23039, dat_path='118/108', offset=0},
+    [7]={min=23040, max=28671, dat_path='286/73', offset=0},
+    [8]={min=28672, max=29695, dat_path='217/21', offset=0},
+    [9]={min=29696, max=30719, dat_path='288/80', offset=0},
+    [10]={min=61440, max=61951, dat_path='288/67', offset=0},
+    [11]={min=65535, max=65535, dat_path='174/48', offset=0},
```
z16: I feel it would look cleaner to use hex values for these. These numbers look pretty magical and hard to visualize where they start and end. I'd recommend looking at these two for reference:
https://github.com/Windower/packages/blob/master/libraries/client_data/client_data/types/items.lua#L220
https://github.com/z16/ResourceBuilder/blob/master/ResourceBuilder/main.cpp#L21

addons/equipviewer/icon_extractor.lua
```diff
@@ -0,0 +1,255 @@
...
+}
+
+local item_by_id = function (id, output_path)
+    local dat_stats = find_item_dat_map(id)
+    local icon_file = open_dat(dat_stats)
+    
+    local id_offset = dat_stats.min + dat_stats.offset
+    icon_file:seek('set', (id - id_offset) * 0xC00 + 0x2BD)
+    local data = icon_file:read(2048)
```
z16: This will sound like a super petty complaint, but it kinda irks me you use hex above but decimal here 😅 `0x800` would look better here imo.

addons/equipviewer/icon_extractor.lua
```diff
@@ -0,0 +1,255 @@
...
+
+    if length == 16 then -- uncompressed
+        data = sub(data, 0x2BE, 0x12BD)
+        data = string.gsub(data, '(...)\128', '%1\255') -- All of the alpha bytes are currently 0 or 0x80.
+    elseif length == 08 then -- color table
+        local color_palette = sub(data, 0x2BE, 0x6BD)
+        color_palette = string.gsub(color_palette, '(...)\128', '%1\255')
+    
+        local n = 0
+        for i = 1, 1024, 4 do
```
z16: Another very minor thing, but same as before, `0x400` would fit with the above, since you do steps of 4, and `4 * 0x400 == 0xC00`, so the numbers all line up :)

z16 (changes requested):
It looks good for the most part, some tabs managed to sneak in, and some minor formatting things that I won't bother mentioning in detail. The things I mentioned are all just suggestions for readability or performance.

One thing that *must* change is an entry in the `addons.xml` file, otherwise the launcher will not know this addon exists. That is its sole reference, the existence of directories is irrelevant to it.

The other comments I'll leave up to you.

addons/equipviewer/equipviewer.lua
RubenatorX: I just added them in when attempting to get around the buggy config stuff. I took them out now.

addons/equipviewer/icon_extractor.lua
RubenatorX: Would you like me to just change most of the numbers to hex? Or just these two?

addons/equipviewer/icon_extractor.lua
RubenatorX: Would you like me to just change most of the numbers to hex? Or just these two?

addons/equipviewer/equipviewer.lua
RubenatorX: Oh uh, I think I just didn't change them from local from the base code I originally worked on. I have no problem making all of them global if you're okay with that

z16: If anything I'd prefer you'd make the others local as well 😅 But can also just leave it as it is. I just thought maybe you made the others global because it didn't work otherwise.

addons/equipviewer/icon_extractor.lua
z16: Personally I think hex should be used everywhere where it is relevant. The size you're reading here is based on the hexadecimal system (same as the base item memory size with `0xC00`), so I think it makes sense. Honestly, in this file most numbers would make sense as hex values, but probably not all. The ones I commented on where just two examples where it stood out to me, but it would need to be looked at on a case-by-case basis. Either way, I wouldn't invest too much time into this, was just a suggestion in the first place.

z16 (approved):

PR #2014, opened 2021-03-24
Five more built in item groups for Treasury.lua

kylericks:
Added built in mapping for a number of objects often pooled by linkshells.  Heroism, Detritus, Moldy Items, Dynamis Divergence Cards, and Shards/Voids.


addons/Treasury/Treasury.lua
```diff
@@ -159,6 +159,22 @@ function find_id(name)
...
+
+
```
z16: One too many.

z16 (changes requested):

addons/Treasury/Treasury.lua
kylericks: Just to confirm, you mean one too many carriage returns correct?  I just want to be sure as I only spotted the extra one after your comment z16

z16: Correct :)

PR #2026, opened 2021-04-20
Keep gearswap active when a user profile cannot be found.

lili-ffxi:
See #2025

### DIFF
--- a/addons/GearSwap/refresh.lua
+++ b/addons/GearSwap/refresh.lua
@@ -79,7 +79,7 @@ function load_user_files(job_id,user_file)
     end
     
     current_file = nil
-    gearswap_disabled = true
+    -- gearswap_disabled = true
     sets = nil
     user_env = nil
     unhandled_command_events = {}
@@ -108,7 +108,7 @@ function load_user_files(job_id,user_file)
     
     if not path then
         current_file = nil
-        gearswap_disabled = true
+        -- gearswap_disabled = true
         sets = nil
         return
     end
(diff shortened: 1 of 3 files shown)


addons/GearSwap/refresh.lua
```diff
@@ -79,7 +79,7 @@ function load_user_files(job_id,user_file)
     end
     
     current_file = nil
-    gearswap_disabled = true
+    -- gearswap_disabled = true
```
z16: Remove this (and the others). No use for commented out code, that's what git is for :)

addons/GearSwap/refresh.lua
```diff
@@ -151,16 +151,16 @@ function load_user_files(job_id,user_file)
...
-    if funct == nil then
+    if err then
         print('User file problem: '..err)
         current_file = nil
-        gearswap_disabled = true
+        -- gearswap_disabled = true
         sets = nil
         return
-    else
+    elseif filename then
```
z16: Since the `else` is irrelevant here you can change this into an
```lua
end

if filename then
```

addons/GearSwap/refresh.lua
lili-ffxi: Done, tho for some reason there's a bunch of other commented out lines already.

addons/GearSwap/refresh.lua
lili-ffxi: I took it out entirely, turns out it's not needed at all!

PR #2045, opened 2021-06-11
[Addon][treasury] Added DisableOnZone setting

lili-ffxi:
The new toggleable feature disables auto-drop when changing zones, to prevent unwanted drop of items.


lili-ffxi:
Two years and a talk with Arcon later, I decided to do things differently, so I'll make a different PR.

PR #2046, opened 2021-06-16
Fix JobChange to work properly in Odyssey

ianandersonlol:
This is a change to job change to allow it to work consistently in Odyssey. 

The first time you enter Idyssey the server will accept an `0x100`, but once you initiate an NM fight the sever will no longer accept an `0x100` until you poke the Moogle again without the BCNM flag. I added a poke for Nomad Moogles as well since technically there is no manual way to change jobs without poking a Nomad Moogle, and if SE wanted to get banhappy it's something they could easily look for.

Note -- changing jobs manually the first time you enter the BC, the last two bytes(unknown 1 & 2) are empty, but once you've killed an NM -- they do contain *some* values, but they don't seem to be related to anything related to the event. The server still accepts a traditional `0x100`.

### DIFF
--- a/addons/JobChange/jobchange.lua
+++ b/addons/JobChange/jobchange.lua
@@ -64,14 +64,23 @@ local find_conflict = function(job_name, p)
     end
 end
 
+injected_poke = false
 local poke = function(npc)
    local p = packets.new('outgoing', 0x1a, {
       ["Target"] = npc.id,
       ["Target Index"] = npc.index,
       })
+      injected_poke = true
       packets.inject(p) 
 end
-
+windower.register_event('incoming chunk',function(id,data,modified,injected,blocked)
+   local p = packets.parse('incoming',data)
+	if id == 0x02E and injected_poke then
+	   injected_poke = false
+      return true
+   end
+end)
+   
 local find_temp_job = function(p)
     for _, job_name in ipairs(temp_jobs) do
         if not find_conflict(job_name, p) and p.jobs[job_name:upper()] > 0 then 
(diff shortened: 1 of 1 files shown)


RubenatorX:
I think adding a poke is great but, isn't this going to awkwardly open the menu? or no?

ianandersonlol:
Nope.

ianandersonlol:
Nomad Moogles are handled pretty much exclusively clientside. While you do send a poke when the client initializes the menu. The server see's everything else the same as if you used the moghouse menu in the main menu within your moghouse.

`0x100` blocked until you poke the Pilgrim Moogle is new and unique.

ianandersonlol:
Oh, actually it does sometimes.

KenshiDRK:
I think the unknown values should be investigated so it doesn't send wrong data to the server, could they be job flags related to the jobs already used in odyssey?

ianandersonlol:
Very possible and I certainly agree. I'm not sure what's worse -- people sending a packet that's getting rejected by the server or the incorrect packet but is being accepted.

For what it's worth when I tested it sending the unknown values without a poke didn't work. Also when I soloed a T1 the `0x100` afterwards had no unknowns.

ianandersonlol:
Good call Rube. Blocked mog menu from opening on injected pokes.

Chiaia:
Till the unknowns are investigated I believe it best to disable the addon from working inside the zones. While still adding the poke for nomad moogles.

This is just a new mechanic that we've never seen before with the job lockouts and I rather be safe at handling it then rush in.

ianandersonlol:
would you be open to disabling it only after the first NM?

Chiaia:
The unknowns seem to just be junk data. I can do the same thing at any mog house moogle or nomad too and the values will shift around or reset as I swap jobs a bunch.

PR #2049, opened 2021-06-24
Addon "Better Check" created

Sephodious:
Created an Addon called Better Check (bcheck) that livens up the check a bit.  I know some other mods did this already but i didn't want all the extra that came with it.  Just wanted to change the /check and nothing else.


addons/BCheck/bcheck.lua
```diff
@@ -0,0 +1,83 @@
...
+
+-- sets local variables and requires
+local packets = require ('packets')
+local message_data = require ('resources').action_messages
+require('chat')
+require('sets')
+
+-- sets local tables
+local valid_message_ids = S{170,171,172,173,174,175,176,177,178}
+local mchal = {'Very Weak':color(259),'Inredibly Easy Prey':color(259),'Easy Prey':color(259),'A Decent Challenge':color(2),'Evenly Matched':color(53),'Tough':color(159),'Very Tough':color(124),'Incredibly Tough':color(167)}
```
z16, line 43: I would split this table (and `mstat` as well) into multiple lines, just so the colors are more easily adjustable. This *could* also be made configurable via settings, so every user can define their own colors, but it's fine for the first version.

addons/BCheck/bcheck.lua
```diff
@@ -0,0 +1,83 @@
...
+       
+        -- tests message_id against valid_message_ids to ensure this only fires off if they match
+        if valid_message_ids:contains(message_id) then
+            
+            local target = windower.ffxi.get_mob_by_id(mobinfo['Target']) or {name=('Unknown')} -- gets mob name
+            local lvl = mobinfo['Param 1']
+            
+            -- gets mob level
+            if lvl > 0x7FFFFFFF then
+                lvl = -1
```
z16, line 64: I think using something like `'?'` would be better than `-1`, I feel that would just be confusing for non-technical people who don't know that `-1` is sometimes used as a dummy value.

addons/BCheck/bcheck.lua
```diff
@@ -0,0 +1,83 @@
...
+            local stat = mstat[message_id - 170]
+            
+            windower.add_to_chat(5, "The [%s] is (Lvl.%s), has ~%s~ and seems like its {%s}":format(target.name:color(2), tostring(lvl):color(213), stat:color(1), chal))
+            return true -- blocks standard /check message
+
+        -- sets output for NMs
+        else if (message_id == 249) then
+          local target = windower.ffxi.get_mob_by_id(mobinfo['Target']) or {name=('Unknown')}
+
+          windower.add_to_chat(2, "[%s] Their power is over 9000!! (Impossible to gauge)":format(target.name:color(167)))
```
z16, line 77: I'm not sure if many people would appreciate that line 😅 Probably better to just use a neutral text, like `Impossible to gauge`.

addons/BCheck/bcheck.lua
```diff
@@ -0,0 +1,83 @@
...
+-- sets local tables
+local valid_message_ids = S{170,171,172,173,174,175,176,177,178}
+local mchal = {'Very Weak':color(259),'Inredibly Easy Prey':color(259),'Easy Prey':color(259),'A Decent Challenge':color(2),'Evenly Matched':color(53),'Tough':color(159),'Very Tough':color(124),'Incredibly Tough':color(167)}
+local mstat = {'high evasion and defense','high evasion','high evasion but low defense','high defense','base defense and evasion','low defense','low evasion but high defense','low evasion','low evasion and defense'}
+
+
+-- uses incoming chunk so that we can block standard /check text
+windower.register_event('incoming chunk',function (id,original,modified,injected,blocked)
+
+    -- ignores packets that don't begin with 29
```
z16, line 50: Technically it ignores packets that don't have ID 0x029 :)

z16 (changes requested):
In addition to the comments within the code, the style of the code is a bit over the place and does not match our common style very well. Especially the indentation is off in some places, try to use 4 space indentations for every level of nesting. You have some parentheses that are unnecessary, some unnecessary spaces before function calls, and missing spaces after commas.

Aside from the indentation we don't require any of the other stylistic changes, but we'd still appreciate it to make our code base be uniform and easier to read.

addons/BCheck/bcheck.lua
RubenatorX: -1 is in line with the behavior of battlemod. The value is used to describe the mobs one would fight at level 1.

addons/BCheck/bcheck.lua
zetaepyon, line 43: Typo on IEP - Inredibly should be Incredibly.
Is there a reason for changing "Too Weak to be Worthwhile" to "Very Weak"? I understand that some of the other slight changes are probably to allow for a more natural flowing sentence, but Too Weak... fits fine as it is.

addons/BCheck/bcheck.lua
```diff
@@ -0,0 +1,83 @@
...
+            
+            -- gets mob level
+            if lvl > 0x7FFFFFFF then
+                lvl = -1
+            end
+            
+            local chal = mchal[mobinfo['Param 2'] - 63]
+            local stat = mstat[message_id - 170]
+            
+            windower.add_to_chat(5, "The [%s] is (Lvl.%s), has ~%s~ and seems like its {%s}":format(target.name:color(2), tostring(lvl):color(213), stat:color(1), chal))
```
zetaepyon, line 70: "seems like its" -> "seems like it's"

addons/BCheck/bcheck.lua
z16: Still feels wrong unless -1 is the *actual* value reported. Speaking of which, what is the actual value?

RubenatorX: `2147483649` (`0x80000001`)

RubenatorX: oh sorry nvm its reporting `4294967295` (`0xFFFFFFFF`) which is twos compliment for -1

Sephodious:
Hey, sorry i haven't been around.  I'll get started on all this later today or tomorrow and resubmit.  Thanks for the suggestions and comments!

addons/BCheck/bcheck.lua
RubenatorX: okay so Sephodious, we talked it over and if you `require('bit')` with the rest of your requires... and then instead of doing...
```
            local lvl = mobinfo['Param 1']
            
            -- gets mob level
            if lvl > 0x7FFFFFFF then
                lvl = -1
            end
```
Just do
```
local lvl = bit.tobit(mobinfo['Param 1']) -- convert unsigned int to signed int
```

addons/BCheck/bcheck.lua
Chiaia: I lol'd . IDK how I feel about this one. Arcon makes a point but these add-ons are made by the community and a little personal flavor text never hurt anyone. Its not like we're some real project.

addons/BCheck/bcheck.lua
Chiaia: Needs to match in-game terms imo.

PR #2051, opened 2021-06-28
Organizer: Fix for index error

johnarnfield:
Discussed at https://www.ffxiah.com/forum/topic/46589/organizer/20/#3584226. Happening for me in Mhaura.

Would it be possible to document `get_mob_list` on https://github.com/Windower/Lua/wiki/FFXI-Functions ?

### DIFF
--- a/addons/organizer/items.lua
+++ b/addons/organizer/items.lua
@@ -41,10 +41,12 @@ do
     
     nomad_moogle = function()
         if #moogles == 0 then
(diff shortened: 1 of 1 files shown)


lili-ffxi:
get_mob_list was implemented precisely to avoid using get_mob_array in a context like this one - it's an entirely new function that, as you noticed, has an issue or two. It's being worked on, and will get added to the wiki soonTM.

PR #2052, opened 2021-07-12
[thtracker] Update to use packets

KenshiDRK:
Update th tracker to use packets instead of text.

### DIFF
--- a/addons/thtracker/thtracker.lua
+++ b/addons/thtracker/thtracker.lua
@@ -27,11 +27,12 @@
 
 _addon.name = 'THTracker'
 _addon.author = 'Krizz'
-_addon.version = 1.1
+_addon.version = 1.2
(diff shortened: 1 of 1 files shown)


addons/thtracker/thtracker.lua
```diff
@@ -51,7 +52,9 @@ defaults.bg.blue = 30
 
 settings = config.load(defaults)
 
-th = texts.new('No current mob', settings)
+th = texts.new('${th_string}', settings)
```
z16, line 55: If a text box contains no actual variable splicing (as in, the entire text box is just one variable), it performs better to not use variables and just update the entire text with `th:text(new_text)`. Not a real issue, just mentioning it for future development.

z16 (approved):

PR #2055, opened 2021-07-15
Update MyHome.lua

ianandersonlol:
It attempts and fails to equip instant warps. This change makes it only try to equip it if the item_info has data in the slot column.


ianandersonlol:
This might not be necessary. Lili told me it didn't work because no one bothered fixing it and I never checked if it *didn't* work beforehand, but looking at the original version I have no reason to believe it didn't work before, but it certainly works with this change.

ianandersonlol:
Yes, this was unnececsary. I blame Lili for leading my astray.

PR #2057, opened 2021-07-17
Update roe.lua

ianandersonlol:
See discord message

### DIFF
--- a/addons/roe/roe.lua
+++ b/addons/roe/roe.lua
@@ -72,7 +72,7 @@ local function accept_roe(id)
     if not id or _roe.complete[id] or _roe.active[id] then return end
     if id and id >= 4008 and id <= 4021 then return end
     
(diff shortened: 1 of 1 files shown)


Chiaia:
The blacklist by default has no entries, so this does nothing almost. Ideally we would ignore blacklist all together and not allow it to set\save those ranges since they are limited time anyway imo,with a warning it could not set certain ROEs. Also you forgot to version bump.

ianandersonlol:
That's odd. My version does have it in the settings file. Not sure how it got there, I wouldn't have put it there myself.

ianandersonlol:
This commit is better -- it just prevents the illicit RoE's from being set. Don't need to worry about the blacklist. (I know you don't like single line if statements I'm just keeping with the convention of the addon. ;)

PR #2065, opened 2021-07-31
Update fields.lua

KenshiDRK:
Update for incoming packet 0x075


Byrth:
Going to defer to @z16 for this one because it uses the more complicated part of the packets interface.

PR #2070, opened 2021-08-11
Add 2nd cmd param to specify dim item.

shastaxc:
I normally carry all 3 dimensional rings with me. I understand that the addon was originally written with the idea in mind that you are using dim rings to get to Reisenjima or Al'Taeiu, and it works great for that purpose. However, I have sometimes desired that I be able to use them simply to get to specific crags, such as while farming void NMs for empyrean trials. This change allows us to specify a second parameter to accept the type of dim item such as `//dim mea` while leaving the original `//dim` functionality intact.

I included a small version bump in this change. Not sure if that is standard practice or not; this is my first pull request.


addons/Dimmer/Dimmer.lua
```diff
@@ -53,17 +53,32 @@ function search_item()
...
                 item_array[item.id].bag_enabled = bag.enabled
             end
         end
     end
-    for index,stats in pairs(item_info) do
+
+    -- Narrow list of teleport options
+    local item_options = {}
+    if name then
+      for index,stats in pairs(item_info) do
```
z16: Use four space indentation everywhere.

addons/Dimmer/Dimmer.lua
```diff
@@ -53,17 +53,32 @@ function search_item()
...
+        if stats.short_name == name then          
+          table.append(item_options, stats)
```
z16: This can be simplified quite a bit. This entire code block could be condensed into one line if you `require('tables')` at the top and prepend the `item_info` table with a `T` like this:
```lua
item_info = T{
    ...
}
```

Then you can write this:
```lua
local item_options = name
    and item_info:filter(function(item) return item.short_name == name end)
    or item_info
```

However, this entire thing asks to be refactored imo. I would probably move the loop contents into its own function `process_item` or whatever, then do this:
```lua
if name then
    local _, stats = item_info:find(function(item) return item.short_name == name end)
    process_item(stats)
else
    for _, stats in pairs(item_options) do
        process_item(stats)
    end
end
```

This could be simplified further if you removed the `short_name` property, and instead just keyed the table by the command:
```lua
item_info = {
    holla={id=26176,japanese='Ｄ．ホラリング',english='"Dim. Ring (Holla)"',slot=13},
    dem  ={id=26177,japanese='Ｄ．デムリング',english='"Dim. Ring (Dem)"',slot=13},
    mea  ={id=26178,japanese='Ｄ．メアリング',english='"Dim. Ring (Mea)"',slot=13},
    mask ={id=10385,japanese="キュムラスマスク+1",english="Cumulus Masque +1",slot=4},
}
```

This can be done because the index is not used anywhere. Now you can simplify it further to:
```lua
if name then
    process_item(item_info[name])
else
    for _, stats in pairs(item_options) do
        process_item(stats)
    end
end
```

addons/Dimmer/Dimmer.lua
```diff
@@ -104,7 +119,7 @@ windower.register_event('addon command',function(...)
         windower.chat.input('//dimmer')
```
z16: I know this was here before your edit, but I'm not liking how it inputs the command again instead of just doing `search_item()`. Would be cool if you could change that.

z16 (changes requested):
The intention of the addon is fine. There are some stylistic changes that are necessary (we use four space indentation everywhere, and preferably spaces after commas, although this was here even before your edit), and I'm not sure why, but the `Stubborn` addon got caught in this PR as well, despite only having whitespace-changes. Would be cool if you could remove it from the PR, just so the change history doesn't get messed up.

The refactoring suggestion is just that, a suggestion. Your code is not wrong, and if you prefer it that way let me know, I'll merge it after the other things have been adjusted.

shastaxc:
I was trying to make as few changes as possible to not step on anyone's toes (changing the structure of the table, importing other libraries). But at your suggestion, I'll go ahead and make those updates.

z16:
> I was trying to make as few changes as possible to not step on anyone's toes (changing the structure of the table, importing other libraries).

Don't worry about that, we're always trying to improve older code of ours. Both to ease future maintenance as well as to increase performance for users.

shastaxc:
I made the changes requested. Most of the changes were made as you suggested. However, as for using short_name for the item_info table's indices, I first did that as you suggested, but later decided to use the item ID instead. This allowed me to do efficient lookups to add data later. Previously, two tables were maintained in order to pull data, which was confusing to me and also caused scope to be annoying after moving the loop contents to a function. I combined these into one table.

shastaxc:
bump

PR #2074, opened 2021-08-27
Update to properly handle more valid commands

RubenatorX:
Now will properly swap for valid commands such as
`/ma "Garuda"<me>` and `/ma "Elemental Siphon"<me>`
Also will now properly handle (ignore) more valid forms of waits such as
`<wait>` and `<wait 1.2>`
and waits that are not preceded by a space
`/ma garuda <me><wait 2.2>`

See discord #development discussion for additional context: https://discord.com/channels/338590234235371531/501099842097905675/880870405966024784

### DIFF
--- a/addons/GearSwap/triggers.lua
+++ b/addons/GearSwap/triggers.lua
@@ -44,8 +44,8 @@ windower.register_event('outgoing text',function(original,modified,blocked,ffxi,
     windower.debug('outgoing text')
     if gearswap_disabled then return modified end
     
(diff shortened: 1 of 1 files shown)


Nifim:
I still find the method used to handle the parsing problematic, I think its made clear by the comment on this line https://github.com/RubenatorX/Lua/blob/45ed71746eaab9b0004fc9e9c0d86ca790c8bea1/addons/GearSwap/triggers.lua#L58

we dont remove the `0x7` char before continuing on in the function, some places we remove it some we dont, It is inconstant. I think we could take the opportunity to do more than a band aid. but it is v4 so ¯\\\_(ツ)_/¯

PR #2076, opened 2021-09-02
Add support for items in shortcuts addon.

Imisnew2:
Adds support for items so that `/<item name>` causes that item to be used -- i.e. translates to `/item "<item name>" <me>`.

Test cases:
`/remedy`
`/rolandaifuku`
`/rolandaifuku+1`

### DIFF
--- a/addons/Treasury/ReadMe.md
+++ b/addons/Treasury/ReadMe.md
@@ -35,6 +35,10 @@ This will either clear the specified list (for the current character only) or li
 
 Lots/passes on all items currently in the pool
 
+`//treasury done`
+
+Passes on all items currently in the pool that have not been lotted or passed yet. Equivalent to the in-game 'Done' menu command.
+
 `//treasury clearall`
 
 Clears all character-specific settings (it will keep global settings)
--- a/addons/Treasury/Treasury.lua
+++ b/addons/Treasury/Treasury.lua
@@ -8,6 +8,9 @@ config = require('config')
 packets = require('packets')
 require('logger')
 
+local MIN_TREASURE_SLOT_INDEX = 0
+local MAX_TREASURE_SLOT_INDEX = 9
+
 defaults = {}
 defaults.Pass = S{}
 defaults.Lot = S{}
@@ -188,7 +191,7 @@ function find_id(name)
 end
 
 function pool_ids()
-    return S(T(windower.ffxi.get_items().treasure):map(table.get-{'item_id'}))
+    return S(T(windower.ffxi.get_items('treasure')):map(table.get-{'item_id'}))
 end
 
 stack = function()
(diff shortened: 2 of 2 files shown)


addons/shortcuts/statics.lua
```diff
@@ -198,9 +198,10 @@ end
 -- Iterate through resources and make validabils.
 function validabils_it(resource)
     for id,v in pairs(res[resource]) do
-        if (not v.monster_level and v.prefix) or (v.monster_level and v.monster_level ~= -1 and v.ja:sub(1,1) ~= '#' ) then
+        if (not v.monster_level and v.prefix) or (v.monster_level and v.monster_level ~= -1 and v.ja:sub(1,1) ~= '#' ) or (v.category == 'Usable') then
```
Imisnew2: This limits the items added to ones in the `Usable` category, which seem to be items like medicines, food, and clusters, but doesn't include usable gear, like `Warp Ring`. This seems fine to me, but highlighting this change in case someone else has a different opinion.

addons/shortcuts/statics.lua
```diff
@@ -209,4 +210,7 @@ validabils_it('spells')
 validabils_it('job_abilities')
 validabils_it('weapon_skills')
 validabils_it('monster_abilities')
-validabils_it('mounts')
\ No newline at end of file
+validabils_it('mounts')
+if settings.include_items then
+    validabils_it('items')
+end
```
Imisnew2: Guards against supporting items. Is false by default.

addons/shortcuts/statics.lua
```diff
@@ -198,9 +198,10 @@ end
 -- Iterate through resources and make validabils.
 function validabils_it(resource)
     for id,v in pairs(res[resource]) do
-        if (not v.monster_level and v.prefix) or (v.monster_level and v.monster_level ~= -1 and v.ja:sub(1,1) ~= '#' ) then
+        if (not v.monster_level and v.prefix) or (v.monster_level and v.monster_level ~= -1 and v.ja:sub(1,1) ~= '#' ) or (v.category == 'Usable') then
         -- Monster Abilities contains a large number of player-usable moves (but not monstrosity-usable). This excludes them.
-            make_abil(strip(v.english),resource,id)
+            local name = resource ~= 'items' and strip_non_alphanumeric_convert_digits_to_roman(v.english) or strip_non_alphanumeric_keep_plus(v.english)
```
Byrth: Are there name collisions if you don't keep the plus and avoid converting digits to roman numerals?

addons/shortcuts/statics.lua
Byrth: This seems okay to me. If someone wants to revisit this decision, they can do it in another PR.

addons/shortcuts/shortcuts.lua
```diff
@@ -117,10 +117,16 @@ default_aliases = {
     cw5="Curing Waltz V",
     hw="Healing Waltz"
 }
+default_settings = {
+    include_items = false,
+}
 
-aliases = config.load('data\\aliases.xml',default_aliases)
+aliases = config.load('data/aliases.xml', default_aliases)
+settings = config.load('data/settings.xml', default_settings)
```
Byrth: Adding another settings file just to allow people to disable compatibility with items seems like a lot of work for no obvious benefit. I would just make it an always-on feature. No one is going to accidentally type `//proether+3` or `//vileelixir+1`

Byrth (commented):
Looks good. Just a few questions.

addons/shortcuts/statics.lua
RubenatorX: I don't see a reason why it couldn't shouldn't check equipped items and make those usable via command.
or, just, checking the flags instead of the category to see if its "use"-able

addons/shortcuts/statics.lua
Imisnew2: I didn't try, but I felt `/rolandaifuku+1` felt more natural than `/rolandaifuku1`.

addons/shortcuts/shortcuts.lua
Imisnew2: I did this because of vague concerns that this would cause problems. https://discord.com/channels/338590234235371531/501099842097905675/883160617706483712

addons/shortcuts/statics.lua
Byrth: That's fine, but you can use `/rolandaifuku+1` either way.

Byrth: My main concern here is that this adds complexity (two slug paths) without adding any obvious benefit.

addons/shortcuts/shortcuts.lua
Byrth: This adds complexity without benefit, so I would eliminate it. Make items always-on and don't add a settings file.

addons/shortcuts/statics.lua
Imisnew2: No. If we only used the original, it would be simply /rolandaifuku for both the normal and +1 versions: the plus gets stripped and the 1 gets omitted when converting to Roman numerals.

PR #2081, opened 2021-09-21
A few functions for decoding dialog entries

trv6:
This was originally written for updating the pointwatch offsets, but there are a couple other addons that could potentially use the functions.

### DIFF
--- a/addons/libs/dialog.lua
+++ b/addons/libs/dialog.lua
@@ -1,0 +1,227 @@
+-- This library was written to help find the ID of a known
+-- action message corresponding to an entry in the dialog tables.
+-- While the IDs can be collected in-game, they occasionally
+-- change and would otherwise need to be manually updated.
+-- It can also be used to find and decode an entry given the ID.
+
+-- Common parameters:
+--
+-- dat: Either the entire content of the zone dialog DAT file or
+-- a file descriptor.
+-- i.e. either local dat = io.open('path/to/dialog/DAT', 'rb')
+-- or  dat = dat:read('*a')
+-- The functions are expected to be faster when passed a string,
+-- but will use less memory when receiving a file descriptor.
+--
+-- entry: The string you are looking for. Whether or not the string
+-- is expected to be encoded should be indicated in the parameter's
+-- name. If you do not know the entire string, use dev.find_substring
+-- and serialize the result.
+
+local xor = require('bit').bxor
+local floor = require('math').floor
+local string = require('string')
+local find = string.find
+local sub = string.sub
+local gsub = string.gsub
+local format = string.format
+local char = string.char
+local byte = string.byte
+require('pack')
+local unpack = string.unpack
+local pack = string.pack
+
+local function decode(int)
+    return xor(int, 0x80808080)
+end
+local encode = decode
+
+local function binary_search(pos, dat, n)
+    local l, r, m = 1, n
+    while l < r do
+        m = floor((l + r) / 2)
+        if decode(unpack('<I', dat, 1 + 4 * m)) < pos then
+        -- offset given by mth ID < offset to string
+            l = m + 1
+        else
+            r = m
+        end
+    end
+    return l - 2 -- we want the index to the left of where "pos" would be placed
+end
+
+local function plain_text_gmatch(text, substring, n)
+    n = n or 1
+    return function()
+        local head, tail = find(text, substring, n, true)
+        if head then n = head + 1 end
+        return head, tail
+    end
+end
+
+local dialog = {}
+
+-- Returns the number of entries in the given dialog DAT file
+function dialog.entry_count(dat)
+    if type(dat) == 'userdata' then
+        dat:seek('set', 4)
+        return decode(unpack('<I', dat:read(4))) / 4
+    end
+    return decode(unpack('<I', dat, 5)) / 4
+end
+
+-- Returns an array-like table containing every ID which matched
+-- the given entry. Note that the tables contain an enormous
+-- number of duplicate entries.
+function dialog.get_ids_matching_entry(dat, encoded_entry)
+    local res = {}
+    local n = 0
+    if type(dat) == 'string' then
+        local last_offset = decode(unpack('<I', dat, 5))
+        local start = 5
+        for head, tail in plain_text_gmatch(dat, encoded_entry, last_offset) do
+            local encoded_pos = pack('<I', encode(head - 5))
+            local offset = find(dat, encoded_pos, start, true)
+            if offset then
+                offset = offset - 1
+                local next_pos
+                if offset > last_offset then
+                    break
+                elseif offset == last_offset then
+                    next_pos = #dat + 1
+                else
+                    next_pos = decode(unpack('<I', dat, offset + 5)) + 5
+                end
+
+                if next_pos - head == tail - head + 1 then
+                    n = n + 1
+                    res[n] = (offset - 4) / 4
+                end
+                start = offset + 1
+            end
+        end
+
+    elseif type(dat) == 'userdata' then
+        dat:seek('set', 4)
+        local offset = decode(unpack('<I', dat:read(4)))
+        local entry_count = offset / 4
+        local entry_length = #encoded_entry
+        for i = 1, entry_count - 1 do
+            dat:seek('set', 4 * i + 4)
+            local next_offset = decode(unpack('<I', dat:read(4)))
+            if next_offset - offset == entry_length then
+                dat:seek('set', offset + 4)
+                if dat:read(entry_length) == encoded_entry then
+                    n = n + 1
+                    res[n] = i - 1
+                end
+            end
+
+            offset = next_offset
+        end
+        local m = dat:seek('end')
+        if m - offset - 4 == entry_length then
+            dat:seek('set', offset + 4)
+            if dat:read(entry_length) == encoded_entry then
+                n = n + 1
+                res[n] = entry_count - 1
+            end
+        end
+    end
+
+    return res
+end
+
+-- Returns the encoded entry from a given dialog table. If you
+-- want to decode the entry, use dialog.decode_string.
+function dialog.get_entry(dat, id)
+    local entry_count, offset, next_offset
+    if type(dat) == 'string' then
+        entry_count = decode(unpack('<I', dat, 5)) / 4
+        if id == entry_count - 1 then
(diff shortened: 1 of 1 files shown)


addons/libs/dialog.lua
```diff
@@ -0,0 +1,158 @@
...
+local find = string.find
+
+local dialog = {}
+
+local function decode(int)
+    return xor(int, 0x80808080)
+end
+local encode = decode
+
+local floor = require('math').floor
```
z16: Should probably put it near the top with the other dependencies.

addons/libs/dialog.lua
```diff
@@ -0,0 +1,158 @@
...
+local function binary_search(pos, dat, n)
+    local l, r, m = 1, n
+    while l < r do
+        m = floor((l+r)/2)
```
z16: Let's add spaces around all operators.

addons/libs/dialog.lua
```diff
@@ -0,0 +1,158 @@
...
+local function plain_text_gmatch(text, substring, n)
+    n = n or 1
+    return function()
+        local pos = find(text, substring, n, true)
+        if pos then n = pos + 1 end
+        return pos
+    end
+end
+
+function dialog.dev_get_offset(dat, id) -- sanity check function
```
z16: I think it would be nicer to have all the development functions in a `dialog.dev.*` subtable. That said, isn't the entire library meant as a dev tool? Or is it meant to be used productively?

addons/libs/dialog.lua
```diff
@@ -0,0 +1,158 @@
...
+    return string.format('%x', decode(unpack('<I', dat, 5+4*id)))
```
z16: Maybe a nitpick, but you added locals for the other string functions, may as well do it for `string.format` :D Same below for `string.sub`, `string.gsub`, `string.char`, `string.byte`.

addons/libs/dialog.lua
```diff
@@ -0,0 +1,158 @@
...
+end
+
+-- Returns an array-like table containing every ID which matched
+-- the given entry.
+-- An important note about this function: The tables contain an
+-- enormous number of duplicate entries. Additionally, some short
+-- entries can occur as substrings of longer entries. Results
+-- which fall into the second case should be filtered out.
+-- If you are dealing with the first case, it may help to look at
+-- any other messages which were received at the same time.
```
z16: Hm, not sure what you mean by that, can you give an example of a DAT where some entries are substrings of other entries? Or do you just mean that some string can occur in two different places, where one contains the other? And if so, what do you mean "should be filtered out"?

addons/libs/dialog.lua
```diff
@@ -0,0 +1,158 @@
...
+            end
+
+            if next_pos-i == #entry then
+                n = n + 1
+                res[n] = (offset-4)/4
+            end
+            start = offset+1
+        end
+    end
+    res.n = n
```
z16: Not sure why these result tables need to contain a length variable. Unless you meant to simulate lists, but then I'd just use lists. Same below.

addons/libs/dialog.lua
trv6: That was originally the plan, but there ended up only being two `dev_*` functions, and the table seemed like a waste. I marked those functions to emphasize that while the others might somehow be useful in an addon, those definitely shouldn't appear: `dev_get_offset` is meant to help alongside a hex editor, and `dev_find_substring` is slow and ambiguous.
There is one case that I can think of where an addon could use the functions outside of development: the messages pointwatch and boxdestroyer use are unique, so the result from calling `get_ids_matching_entry` when you enter one of those zones could be used in place of manually updating the message IDs.

addons/libs/dialog.lua
trv6: The second: some entries were only a few bytes long and the same sequence of bytes showed up within others. The function matches by finding the given string within the file, verifying that the position is a valid offset, and that the length of the entry matches the length of the string. By being "filtered out" I meant that, if you were searching for a short entry, those longer results would not be added to the table returned by the function since the length of the entries would not match.

addons/libs/dialog.lua
trv6: I just tossed it in since it was already available. Otherwise, you'd end up calculating it anyway. It might be better to return it separately? `return res, n`

z16: What I mean is, I see no specific use for it. Not sure why it's there at all, it doesn't help with iteration unless you do it manually (which I don't think people would do, at least not if experience is anything to go by). And if it is explicitly needed, `#` can be used, which works in logarithmic time, which is essentially constant, unless you have tens of thousands of entries.

addons/libs/dialog.lua
```diff
@@ -0,0 +1,165 @@
...
+-- entries can occur as substrings of longer entries. Results
+-- which fall into the second case should be filtered out.
+-- If you are dealing with the first case, it may help to look at
+-- any other messages which were received at the same time.
+function dialog.get_ids_matching_entry(dat, entry)
+    local last_offset = decode(unpack('<I', dat, 5))
+    local res = {}
+    local n = 0
+    local start = 5
+    for i in plain_text_gmatch(dat, entry, last_offset) do
```
z16: Without having tested it myself, are the strings in the DAT not encoded? Or why would a plain text search work?

addons/libs/dialog.lua
z16: Ok, I feel this is pretty self-evident. If I searched for "Hello" I wouldn't want it to return "Hello world".

addons/libs/dialog.lua
```diff
@@ -0,0 +1,165 @@
+-- This library was written to help find the ID of a known
+-- action message corresponding to an entry in the dialog tables.
+-- While the IDs can be collected in-game, they occasionally
+-- change and would otherwise need to be manually updated.
+-- It can also be used to find and decode an entry given the ID.
+
+-- Common parameters:
+-- dat: The entire content of the zone dialog DAT file
+-- i.e. local dat = io.open('path/to/dialog/DAT', 'rb'):read('*a')
```
z16: An interesting overload might be a function that does not take the entire DAT as a string, but a file handle (or path, and get its own handle) and only read as much as needed while iterating. That way the entire DAT wouldn't need to be kept in memory. But just an idea for a possible future version of this file.

addons/libs/dialog.lua
```diff
@@ -0,0 +1,165 @@
...
+        local pos = find(text, substring, n, true)
+        if pos then n = pos + 1 end
+        return pos
+    end
+end
+
+-- This function may help if you are viewing the decoded contents
+-- of a dialog DAT within a hex editor.
+function dialog.dev_get_offset(dat, id)
+    return format('%x', decode(unpack('<I', dat, 5 + 4 * id)))
```
z16: I'd prefer a prefixed and fixed-size offset display like `'0x%8X'`, but not sure if your intention was to use it for some specific tool that expects it in a different format.

addons/libs/dialog.lua
z16: I still think a sub table would make sense, even for just those two. May also add more in the future, I could see a library like this being helpful for a bunch of things.

addons/libs/dialog.lua
trv6: It's a quick way to tell if the message is unique. If `n ~= 1` you have more work to do before you can be certain you've found the correct ID.
I guess I didn't really need it for the one use I had, though.

addons/libs/dialog.lua
trv6: They are encoded. The "plain text" was meant to refer to the "plain" parameter for `string.find`. It searches for the given text, ignoring patterns.
I might go back through later next weekend to see if escaping the string and using string.gmatch is faster.

addons/libs/dialog.lua
trv6: That would be cool. Especially for `get_entry`.

addons/libs/dialog.lua
trv6: I was using it to pipe the offset to my clipboard so that I could paste it into a hex editor. It wouldn't be difficult to trim the first two characters, though, so that's cool with me. I added it as `'0x%08X'`

addons/libs/dialog.lua
z16: I might not be expressing myself properly, I just meant that `#` is used to determine length. Using a custom table key for that feels weird to me. The `lists` library does that, but that was never meant to be publicly used, if Lua supported private variables, it would have been private. It feels very unintuitive, and I don't think anyone would use it over `#` (which would still work).

addons/libs/dialog.lua
z16: No, I think this is the better way, as discussed on Discord. I just meant that if I search for `Hello`, how would it ever match anything, if the file is encoded? It would not appear as `Hello` in the file, which is `\x48\x65\x6C\x6C\x6F`, but as `\xC8\xE5\xEC\xEC\xEF`. So the match would not return any hits, or am I missing something?

trv6: I updated the description for the `entry` parameter. It's meant to be the encoded string, so you'd search for `dialog.encode_string('Hello')`.
That's a little inconvenient on the surface, but in practice I don't think anyone will be typing the entries in. They contain so many special bytes that they're usually unreadable even after decoding. My thought was that most work would be done leaving the strings encoded. You would get the encoded entry either with the ID using `get_entry` or via `dev.find_substring` and then use `dialog.serialize` to get a version of the string that could be pasted into an addon.
I assume the most common use will be to convert from entry to ID, so having the encoded entry saves doing the encoding step later. If you actually needed the decoded string, the earlier results can be passed to to `dialog.decode_string`.
I'll probably switch `dev.find_substring` around so that it uses the decoded string, since it's likely to be used in the way you've shown.

addons/libs/dialog.lua
trv6: Maybe I shouldn't admit this now, but `list.n` became so engrained in my programming that I use it everywhere now. Outside of Windower, and even in some other languages.
I think some of the earlier addons used it because it's shorter than typing `list:length()`, and I copied their style.

trv6:
This should be all set.

- [x] No more "n" key
- [x] The "dat" parameter can be either a string of the file contents or a file descriptor
- [x] Specified in the parameter names when the entry is expected to be un/encoded.
- [x] Formatting for get_offset.

PR #2090, opened 2021-11-06
Update Mount Roulette to 3.1.0

xurion:
Adds the blacklist function.

### DIFF
--- a/addons/MountRoulette/MountRoulette.lua
+++ b/addons/MountRoulette/MountRoulette.lua
@@ -26,14 +26,18 @@ ON ANY THEORY OF LIABILITY, WHETHER IN CONTRACT, STRICT LIABILITY, OR TORT
 SOFTWARE, EVEN IF ADVISED OF THE POSSIBILITY OF SUCH DAMAGE.
 ]]
 
-_addon.name    = 'Mount Roulette'
-_addon.author  = 'Dean James (Xurion of Bismarck)'
-_addon.version = '3.0.1'
+_addon.name = 'Mount Roulette'
+_addon.author = 'Dean James (Xurion of Bismarck)'
+_addon.version = '3.1.0'
 _addon.commands = {'mountroulette', 'mr'}
 
 require('lists')
 require('sets')
 resources = require('resources')
+config = require('config')
+settings = config.load({
+    blacklist = ''
+})
 
 math.randomseed(os.time())
 
@@ -41,6 +45,13 @@ allowed_mounts = L{}
 possible_mounts = L{}
 for _, mount in pairs(resources.mounts) do
     possible_mounts:append(mount.name:lower())
+end
+
+blacklist = {}
+if string.len(settings.blacklist) > 0 then
+    for mount in settings.blacklist:gmatch("([^,]+)") do
+        table.insert(blacklist, mount)
+    end
 end
 
 function update_allowed_mounts()
@@ -55,7 +66,10 @@ function update_allowed_mounts()
             end)
             local mount = possible_mounts[mount_index]
 
-            allowed_mounts_set:add(mount)
+            -- Add this to allowed mounts if it is not already there and it is not blacklisted
+            if not allowed_mounts:contains(mount) and not S(blacklist):contains(mount) then
+                allowed_mounts_set:add(mount)
+            end
         end
     end
 
@@ -70,7 +84,9 @@ windower.register_event('incoming chunk', function(id)
     end
 end)
 
-windower.register_event('addon command', function()
+commands = {}
+
+commands.mount = function()
     local player = windower.ffxi.get_player()
 
     -- If the player is mounted, dismount now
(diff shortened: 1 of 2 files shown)


addons/MountRoulette/MountRoulette.lua
```diff
@@ -86,4 +102,73 @@ windower.register_event('addon command', function()
     -- Generate random number and use it to choose a mount
     local mount_index = math.ceil(math.random() * #allowed_mounts)
     windower.send_command('input /mount "' .. allowed_mounts[mount_index] .. '"')
+end
+
+commands.blacklist = function(args)
+    local operation = args:remove(1)
+
+    if not operation then
```
z16, line 103: Am I reading this right if you don't supply any arguments and just do `//mr` or `//mr mount` it would just list the blacklisted mounts?

addons/MountRoulette/MountRoulette.lua
```diff
@@ -43,6 +47,13 @@ for _, mount in pairs(resources.mounts) do
     possible_mounts:append(mount.name:lower())
 end
 
+blacklist = {}
```
z16: This should really be a set `S{}`. And so should `settings.blacklist`. In fact, not sure why you don't make `settings.blacklist` a set and use that directly. Does that not work for you?

addons/MountRoulette/MountRoulette.lua
```diff
@@ -55,7 +66,10 @@ function update_allowed_mounts()
             end)
             local mount = possible_mounts[mount_index]
 
-            allowed_mounts_set:add(mount)
+            -- Add this to allowed mounts if it is not already there and it is not blacklisted
+            if not allowed_mounts:contains(mount) and not S(blacklist):contains(mount) then
```
z16: The `not allowed_mounts:contains(mount)` check is unnecessary.

addons/MountRoulette/MountRoulette.lua
```diff
@@ -86,4 +102,73 @@ windower.register_event('addon command', function()
...
     local mount_index = math.ceil(math.random() * #allowed_mounts)
     windower.send_command('input /mount "' .. allowed_mounts[mount_index] .. '"')
+end
+
+commands.blacklist = function(args)
+    local operation = args:remove(1)
+
+    if not operation then
+        windower.add_to_chat(8, 'Blacklisted mounts:')
+        for k, v in ipairs(blacklist) do
```
z16: If you make `blacklist` a set, you can use this for more convenient iteration:
```lua
for mount in blacklist:it() do
```

Same below, no need to `ipairs(T(allowed_mounts))`.

addons/MountRoulette/MountRoulette.lua
xurion: If you do `//mr`, line 162 defaults the command to `mount`. If you define the command, like `//mr mount` or `//mr blacklist`, the correct command is referenced from the `commands` table.

z16: Yeah, but removing if you don't provide a further option won't `operation` be `nil` here?

xurion: Oh yeah, if you do `//mr blacklist` it just prints all blacklisted mounts, then returns.

z16: Oh nevermind me, I just realized what's happening here. I was misreading how the command handler works... sorry about that.

addons/MountRoulette/MountRoulette.lua
xurion: Ahh yes you're right! I think I had some issues with lists and saving them in settings (I think it was saving them as `<0>...</0><1>...</1>`), but since then I swapped to using sets, so I've changed this to just use the set directly as suggested and it works fine. Thanks!

RubenatorX:
You may want to consider making the list flexible between black and whitelist. Especially so that people that only want a few mounts in the roulette don't have to go and add a new mount every time one comes out. Just a thought.

xurion:
> You may want to consider making the list flexible between black and whitelist. Especially so that people that only want a few mounts in the roulette don't have to go and add a new mount every time one comes out. Just a thought.

Great point. I've seen a whitelist/blacklist pattern in another addon before - do you happen to know which? Maybe I could ~~steal~~ take inspiration from an existing example.

z16:
> > You may want to consider making the list flexible between black and whitelist. Especially so that people that only want a few mounts in the roulette don't have to go and add a new mount every time one comes out. Just a thought.
> 
> Great point. I've seen a whitelist/blacklist pattern in another addon before - do you happen to know which? Maybe I could ~steal~ take inspiration from an existing example.

*AutoJoin* has that pattern, not sure if that's the one you were looking for though. It has a `mode` variable which can be either `whitelist` or `blacklist`, and depending on which value is set it uses the whitelist to choose valid values or the blacklist to exclude invalid values.

xurion:
Yep that's it. I'll look into this for the next thing. I'll default to blacklist as the mode (my assumption is that blacklisting a certain mount or two will be more utilised than a whitelist) so I'm ok for this PR to proceed and the whitelist option to follow later.

xurion:
@z16 If you agree, I think this should be ok to merge. The whitelist feature will come later.

PR #2097, opened 2021-11-24
Update Pointwatch.lua -- Exemplar Points

### LINKED ISSUE #2095: PointWatch
Add Exemplar Point Watch

RubenatorX:
1) Add support for Exemplar Points
2) Partial rework of packets to both utilize the packets library, and to take advantage of last_incoming when it is first loaded
3) Partial rework of Key Item detection for Dynamis
4) Add function to allow for easy coloring text above a certain value threshold.
5) Other minor code style tweaks

**_Do not merge without first merging_** https://github.com/Windower/Lua/pull/2099/files
Resolves #2095

### DIFF
--- a/addons/pointwatch/pointwatch.lua
+++ b/addons/pointwatch/pointwatch.lua
@@ -24,17 +24,19 @@
 --(INCLUDING NEGLIGENCE OR OTHERWISE) ARISING IN ANY WAY OUT OF THE USE OF THIS
 --SOFTWARE, EVEN IF ADVISED OF THE POSSIBILITY OF SUCH DAMAGE.
 
-texts = require 'texts'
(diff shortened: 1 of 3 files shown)


addons/pointwatch/pointwatch.lua
```diff
@@ -323,6 +355,15 @@ function exp_msg(val,msg)
             -- If a merit point was not gained, 
             lp.current = math.min(lp.current,lp.tnm-1)
         end
+    elseif msg == 810 then
```
genoxd: I believe you also need msg == 809. From my testing I saw that 810 is the message for Exemplar points from a chain and 809 is the message for exemplar with no chain

PR #2106, opened 2021-12-12
pointwatch does not get Capacity/Job Points in Japanese environment.

yyoshisaur:
When packets.parse (incoming packet 0x63 Order=5) in Japanese environment, job name is Japanese name in the parsed data.
For example, Warrior Capacity Point is p['戦士 Capacity Points']. Job Point is the same too.
windower.ffxi.get_player().main_job_full returns English name.
Fixed to get the job name that suits your language environment by using res.jobs[job_id].name.

 I would like you to merge, if there is no problem in English environment.

### DIFF
--- a/addons/pointwatch/pointwatch.lua
+++ b/addons/pointwatch/pointwatch.lua
@@ -125,9 +125,12 @@ packet_handlers = {
             lp.number_of_merits = p['Merit Points']
             lp.maximum_merits = p['Max Merit Points']
         elseif p['Order'] == 5 then
-            local job = windower.ffxi.get_player().main_job_full
-            cp.current = p[job..' Capacity Points']
-            cp.number_of_job_points = p[job..' Job Points']
+            local player = windower.ffxi.get_player()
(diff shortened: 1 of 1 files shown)


z16:
This was an oversight on our part. We fixed it on the dev environment, can you see if that makes it work without your workaround?

z16:
You can also adjust this yourself and see if it works for you with that:
https://github.com/Windower/Lua/commit/d2ac675b35ce0ea96f76888e3ec47bf852052628

yyoshisaur:
I tried that out.
When fixed packets.lua, current version pointwatch looks like it's working fine without this workaround.

RubenatorX:
Alright, will mark this as:
Will get resolved by https://github.com/Windower/Lua/commit/d2ac675b35ce0ea96f76888e3ec47bf852052628
once its merged to live, and then we will close the issue. Thanks for bringing it up

Chiaia:
This went Live a day or so ago #2111

PR #2107, opened 2021-12-14
Make it easy to get the dialog files.

trv6:
~~Not tested yet.~~


trv6:
Ran a few tests. Hopefully works.

Byrth:
A quick grep of the repo reveals that no official addons use this lib. You didn't modify existing behavior.

If it works for you, and it looks like it should, then I think we're good.

PR #2108, opened 2021-12-16
pointwatch: Scrape two high bits off the message ID so that it matches the dialog file ID

trv6:
Pointwatch used to scrape off the two highest bits. Why two? Not sure.
There's a comment in fields.lua that the high bit is sometimes set for unknown reasons, which accounts for the highest bit.
As for the next highest bit... There are a few zones which have more than 2^14-1 ids, but none are Abyssean zones. It might have been a mistake, but it shouldn't matter for pointwatch's purposes.


Byrth (approved):
My memory is that the MSB was set in Abyssea - Attohwa. I don't remember why I did the second highest bit.

Otherwise, this would do what the title says and restores it to the old behavior, so I approve.

PR #2110, opened 2021-12-23
Correct improper ID range check

RubenatorX:


addons/GearSwap/targets.lua
```diff
@@ -28,30 +28,34 @@
...
+            local j = windower.ffxi.get_mob_by_id(id)
+            if j then
+                spelltarget = target_complete(j)
+            end
+            spelltarget.raw = targ
+            return targ, spelltarget
+        elseif targ ~= '' then
+            local mob_array = windower.ffxi.get_mob_array()
+            for i,v in pairs(mob_array) do
+                if v.name:lower()==targ:lower() and (not v.is_npc or v.spawn_type == 14) then
```
z16: `targ:lower` should be moved outside of the loop.

addons/GearSwap/targets.lua
```diff
@@ -28,30 +28,34 @@
 -- Target Processing --
 
 function valid_target(targ)
+    local targ = targ or false
```
z16: Why `false`, and why `local`? If it should have a default value just `targ = targ or false` will already be local, since `targ` is a parameter. This way you're introduced a new variable which shadows the parameter.

I think this line can just be removed, no? `false` doesn't get us anything here, unless I'm missing something.

PR #2112, opened 2022-01-04
GearSwap Motes Command-- set mode with spaces

RubenatorX:
Allow to set a specific mode name with a space in it via `gs c set` (previously would ignore anything after a space, even if contained in quotes)


RubenatorX:
I actually am going to comment here that, it would be better if things were changed to accommodate quotes around things (so maybe mode names could contain spaces too??), but this is still an incremental improvment.

PR #2113, opened 2022-01-05
Update Craft Addon

Snapster-amb:
What
====

- Updated slpp.py and create_recipes.py to python 3
- Ran create_recipes.py to build the latest recipe set

Why?
====

There were some missing recipes and the script for generating
new recipes by pulling data from bg-wiki was not working on python3.

### DIFF
--- a/addons/autocontrol/autocontrol.lua
+++ b/addons/autocontrol/autocontrol.lua
@@ -168,21 +168,27 @@ function set_attachments_from_autoset(autoset,slot)
             if petlessZones:contains(windower.ffxi.get_info().zone) then 
                 return
             else
-                local recasts = windower.ffxi.get_ability_recasts()
-                if settings.AutoActivate and recasts[recast_ids.activate] == 0 then
-                    windower.send_command('input /ja "Activate" <me>')
-                elseif settings.AutoDeusExAutomata and recasts[recast_ids.deusex] == 0 then
-                    log('Activate is down, using Deus Ex Automata instead.')
-                    windower.send_command('input /ja "Deus Ex Automata" <me>')
-                elseif settings.AutoActivate and settings.AutoDeusExAutomata then
-                    log('Activate and Deus Ex Automata timers were not ready.')
-                elseif settings.AutoActivate then
-                    log('Activate timer was not ready.')
-                elseif settings.AutoDeusExAutomata then
-                    log('Deus Ex Automata timer was not ready.')
-                end
+                coroutine.schedule(activate_puppet, 0.5)
             end
         end
+    end
+end
+
+function activate_puppet()
+    local recasts = windower.ffxi.get_ability_recasts()
+    if settings.AutoActivate and recasts[recast_ids.activate] == 0 then
+        windower.send_command('input /ja "Activate" <me>')
+    elseif settings.AutoDeusExAutomata and recasts[recast_ids.deusex] == 0 then
+        if settings.AutoActivate then
+            log('Activate is down, using Deus Ex Automata instead.')
+        end
+        windower.send_command('input /ja "Deus Ex Automata" <me>')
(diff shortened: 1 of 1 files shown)


RubenatorX:
Unless py2 can run that, I'd probably prefer if we also kept a py2 version so we can pick

addons/craft/create_recipes.py
```diff
@@ -1,4 +1,7 @@
-import urllib2
+try:
+    import urllib.request as urllib2
+except ImportError:
+    import urllib2
```
z16: As far as I know `urllib2` no longer exists in Python 3, so why would importing it work in the exception case?

addons/craft/create_recipes.py
```diff
@@ -118,10 +121,10 @@ def get_recipes_from_soup(soup, spheres=False):
 
 def get_items_dictionary():
     if platform.system() == 'Windows':
-        path = 'C:\\Program Files (x86)\\Windower4\\res\\items.lua'
+        path = 'C:/Program Files (x86)/Windower/res/items.lua'
```
z16: This really bothers me, but it bothered me before your change as well. Why is this hard coded instead of using a relative path, or alternatively a path passed in via the command line (or both, with the relative path being a fallback)?

Also, backwards slashes are correcterer.

z16 (commented):
I don't mind it being Python 3 only. I don't think any of the 0.7 people using this have any trouble installing Python 3, if they don't have it already. Just curious about the import and the default path. I assume this was chosen because that's the default path chosen by the new installer? Though that in itself also bothers me a bit, I'll have a talk with the team about this.

addons/craft/create_recipes.py
Snapster-amb: It's supposed to be backwards compatible with python2 although I haven't tested this.  I can give it a test and see if it actually works.

addons/craft/create_recipes.py
Snapster-amb: The reason is mostly laziness.  The python scripts are things that a developer or user might periodically run to update the recipes and create a pull request, so much less thought was put into these kings of details.

addons/craft/create_recipes.py
Snapster-amb: I was able to test this and with a few additional changes that are now part of a PR, it works on python 2 and 3.  One issue I ran into on python2 was getting hit with capcha whereas I didn't with python3.  I'm not sure if that has anything to do with the changes though as that part of the code is untouched.  I was able to test python2 by running the code on python3 first so that the html files were downloaded without issue, and then running it again in the python2 environment.

z16: I'm not a fan of messing the code up to support both versions, Python 2 deserves to die already :P But I won't force you to change it back, I'll leave it up to you.

addons/craft/create_recipes.py
z16: Considering most people will run this script from their Windower installation I'd just make it relative. This path would break for me, for example, as well as for many people who followed the common advice we gave of where to install Windower to until recently. That falls well within the limits of laziness imo :)

Ideally we'd have another mechanism for specifying the path, but those are obviously more work and would exceed the purpose of the PR, especially since this is nothing the PR changed.

addons/craft/create_recipes.py
```diff
@@ -204,10 +208,10 @@ def build_recipe_string(name, crystal, ingredients):
 
 def save_recipes(recipes):
     with open('recipes.lua', 'w') as fd:
```
z16: Would this (and other) `open` functions benefit from the `encoding` parameter? You added to to the file you're reading from, but I'm not sure what exactly we're writing here. Can non-ASCII characters appear, like Japanese text?

PR #2170, opened 2022-03-13
[pointwatch] support Master Levels (specifically)

posimagi:
This diff creates the `ep.master_level` variable which can be used in pointwatch format strings to display the current job's current Master Level. Note that due to the fact that the incoming packet `0x061` reports the incorrect Master Level under Level Sync (which is why the game client itself shows the wrong level), this value will also be wrong under Level Sync. While it's certainly possible to hard-code a TNL table to solve this, I've decided to implement the simpler, non-heuristic solution for now. I have also opted not to rock the boat by adding this new value to the default format strings, but I believe that they probably should be added in the future.

Example of how this can be used:
image


RubenatorX:
The addon does not have a variable for job level, so I'm curious what your argument is for it having master level?

posimagi:
Several people I've been in groups with have commented that they wished it showed Master Level, or said things such as "I wonder how many Master Levels we've gotten" in longer parties. I wouldn't be opposed to it providing glanceable character level either.

RubenatorX:
Okay but, levels are not like merit points, to you don't spend them and they are readily available on the equipment or status screen just like masterb levels and you aren't complaining about that. I don't think it should be in the pointeatch defaults for sure. And if it's added as an option, I feel like job level should be added as well. But again, not as a default

posimagi:
I am happy to add job levels as an option as well, and to leave both as not default. I do realize that all of these values are a few buttons away, but I do think it'd be nice to have the option to display them in the addon.

posimagi:
Added main and sub job level options as well, proof that they work: 
image

posimagi:
Need anything else from me before this is ready to be merged?

RubenatorX:
Bump the version number.
Not a requirement but, while you're in there, you might as well add job names and abbreviations 🤷‍♂️

posimagi:
I'll bump the version number for now. I'll do the jobs in a separate pull.

posimagi:
Updated.

posimagi:
Added jobs.

posimagi:
Test: 
image

addons/pointwatch/statics.lua
```diff
@@ -81,7 +81,57 @@ other_table = {Crimson=10,Azure=10,Amber=10,Alabaster=10,Obsidian=20}
 dynamis_map = {[185]=city_table,[186]=city_table,[187]=city_table,[188]=city_table,
     [134]=other_table,[135]=other_table,[39]=other_table,[40]=other_table,[41]=other_table,[42]=other_table}
 
-    
+
+job_map = {
```
zetaepyon: Maybe this and the abbreviation map could be pulled from resources instead of a hard-coded mapping?

z16: It should be, this and the `job_abbr_map` should be removed and handled via resource lookups. Should also use `.name` to get language-specific values.

posimagi:
Thanks for the comments. I have removed support for job names/abbreviations from this PR and I will submit them separately with the relevant data pulled from resources. Should be good to merge with only the previous functionality included.

PR #2173, opened 2022-03-30
add yell and linkshell2 support

arshesney:
Added filtering for yell and linkshell2 messages, the raw yells have some unicode characters wrapping sender name that require a different matching pattern

### DIFF
--- a/addons/blist/blist.lua
+++ b/addons/blist/blist.lua
@@ -256,7 +256,6 @@ end)
 windower.register_event('incoming text',function (original, modified, mode)
 	if settings.useblist == true then
 		name = "blist"
-		--print(tostring(mode))
 		if (mode == 14 or mode == 214) and settings.linkshell == true then -- linkshell (others)
 			a,z,name = string.find(original,'<(%a+)> ')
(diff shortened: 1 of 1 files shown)


addons/blist/blist.lua
```diff
@@ -254,7 +256,8 @@ end)
 windower.register_event('incoming text',function (original, modified, mode)
 	if settings.useblist == true then
 		name = "blist"
-		if mode == 14 and settings.linkshell == true then -- linkshell (others)
+		--print(tostring(mode))
```
z16: Can remove the `print`.

z16 (commented):
Looks good, just remove the one comment. Also, do you think it makes sense to split up LS1 and LS2? I'm ambivalent on the issue.

arshesney:
I though about splitting the two linkshells, but, at least from my point of view, if I blacklist someone I wouldn't want him popping in other spots. Yet you might be right and having both as a separate option is more in line with the original idea.

z16:
This is good for now, I wasn't even advocating for splitting them up, I was just curious. I'll leave that decision up to you/the community.

PR #2179, opened 2022-04-09
Support bug fix

Atsuke:
Occasionally addon will try to resume crafting too quickly after interacting with support npc causing entire queue to be dumped with "you cannot craft at this time" errors.  Added a short delay in the check_queue function after poke_npc() call.


paulframe:
You haven't said what addon we're talking about.

RubenatorX:
~~Which addon(s)?~~
oh nvm, I thought this was an issue submission, my bad

PR #2182, opened 2022-04-26
Create a log file for debugging pointwatch message ids

trv6:
This would create a small file in the data folder after reading from a dialog dat. If someone noticed buggy behavior with the offsets, they could verify that the number in the log file matches the number they get from following the steps in the fixing pointwatch file.


addons/pointwatch/message_ids.lua
```diff
@@ -211,7 +215,21 @@ local function update_offset(zone_id)
...
+                n = n + 1
+                s[n] = t.name .. '\t' .. tostring(t.offset)
+            end
+        end
+        if n > 0 then
+            local f = io.open(windower.addon_path .. '/data/dia.log', 'w')
+            f:write(os.date() .. '\n' .. table.concat(s, '\n'))
+            f:close()
+        end
+    end
```
z16, line 232: Is this `end` not one too many? Or what is it meant to close?

addons/pointwatch/message_ids.lua
```diff
@@ -211,7 +215,21 @@ local function update_offset(zone_id)
         end
 
         m.offset = res[1]
-        m.name = nil
+        m.update = nil
+        local s = {}
```
z16, line 219: I'd prefer some better variable names. I realize that `m` and `f` also weren't perfect before, but no need to repeat those mistakes :) In fact, can adjust them while you're in there.

addons/pointwatch/message_ids.lua
```diff
@@ -211,7 +215,21 @@ local function update_offset(zone_id)
...
+        local n = 0
+        for zone, t in pairs(messages) do
+            if not t.update then
+                n = n + 1
+                s[n] = t.name .. '\t' .. tostring(t.offset)
+            end
+        end
+        if n > 0 then
```
z16, line 227: Wouldn't `n` always be `> 0`? At least `1`, because you set `m.update` to `nil` above, so if you iterate the messages after that, at least the current message (`m`) would result in `not t.update` being `true`?

z16 (commented):
I'm honestly not sure what this does, but if I also don't know much about PW or how it handles DAT files. Is this done automatically, always? Or only with some debug flag? Because I would very much prefer the latter. If someone has an issue they could still be told to enable the flag and reproduce it, right? This is otherwise just a lot of extra work that seems needless for regular operations.

PR #2191, opened 2022-05-20
Add finalAlert (new Addon)

brangerjoe:
finalAlert is an addon that displays FF7-style toast windows for enemy WS/magic. It can display whether they're interrupted, and emphasize skills by name. I just learned it's become useful to many people, so I thought I'd check it into the Windower codebase. Screenshots below. First PR here - hopefully everything checks out. :)

image
image
image
image
image


addons/finalAlert/finalAlert.lua
```diff
@@ -0,0 +1,303 @@
+Copyright © 2022, Godchain
```
z16, line 2: The entire copyright notice is not wrapped in comments. This way, the addon won't load. Wrap it in `--[[` and `]]`.

addons/finalAlert/finalAlert.lua
```diff
@@ -0,0 +1,303 @@
...
+
+Redistribution and use in source and binary forms, with or without
+modification, are permitted provided that the following conditions are met:
+
+    * Redistributions of source code must retain the above copyright
+      notice, this list of conditions and the following disclaimer.
+    * Redistributions in binary form must reproduce the above copyright
+      notice, this list of conditions and the following disclaimer in the
+      documentation and/or other materials provided with the distribution.
+    * Neither the name of <addon name> nor the
```
z16: Add the addon name here :)

addons/finalAlert/README.md
```diff
@@ -0,0 +1,40 @@
+# finalAlert
+Displays Final Fantasy 7 style alerts when enemies use skills or magic. Allows you to emphasize skills by name (useful for interrupts).
+
+## Commands:
+
+### Test window
```
z16, line 6: I'm honestly not sure if this command should be even present in the finished addon, feel even less enthused about it being documented. But up to you. It seems like a debug command and those should not be user facing imo.

addons/finalAlert/finalAlert.lua
```diff
@@ -0,0 +1,303 @@
...
+background_emphasize = "background_emphasize"
+
+-- Timing
+showing = false
+last_trigger = 0
+
+-- IDs
+weapon_skill_category = 7
+magic_category = 8
+interrupt_id = 28787
```
z16, line 52: How was this determined? If this is a zone resource, it might be different for every zone, and subject to changes after updates.

addons/finalAlert/finalAlert.lua
```diff
@@ -0,0 +1,303 @@
...
+
+defaults = {}
+defaults.x_position = windower.get_windower_settings().x_res / 2
+defaults.y_position = 100
+defaults.background_size = "regular"
+defaults.emphasize = S {}
+defaults.trigger_duration = 3
+defaults.sounds = "on"
```
z16, line 60: I would use boolean values instead of `on`/`off`, and even rather than binary enum values, like for the `background_size` above. I'd reword it to `large` with `true`/`false` as possible values or `small`/`compact`, with the same (but reversed) values. Users are notoriously clumsy with handling a single set of allowed values, you'll not only get users who misspell the allowed values, but also users who will argue about case insensitivity. The boolean approach saves you from a bunch of possible issues.

addons/finalAlert/finalAlert.lua
```diff
@@ -0,0 +1,303 @@
...
+
+settings = config.load(defaults)
+
+windower.register_event(
+    "load",
```
z16, line 65: I wouldn't couple this to the `load` FFXI event, but instead use `config.register` to hook into the settings reload event. This will trigger on load as well, and additionally on relogging to a character with different settings.

addons/finalAlert/finalAlert.lua
```diff
@@ -0,0 +1,303 @@
...
+
+windower.register_event(
+    "addon command",
+    function(cmd, ...)
+        local args = L {...}
+
+        if not cmd or cmd == "help" then
+            local bullet = windower.to_shift_jis("» ")
+            print("=== Usage Examples ===")
+            print(bullet .. "//fa test ws")
```
z16, line 98: See above about the `test` command.

addons/finalAlert/finalAlert.lua
```diff
@@ -0,0 +1,303 @@
...
+
+caption = texts.new({})
+
+background_ability = "background_ability"
+background_magic = "background_magic"
+background_interrupt = "background_interrupt"
+background_emphasize = "background_emphasize"
+
+-- Timing
+showing = false
```
z16, line 46: This variable can be removed entirely and be replaced with `caption:visible()`. It's a very cheap function call.

addons/finalAlert/finalAlert.lua
```diff
@@ -0,0 +1,303 @@
...
+
+            if act.param == interrupt_id then
+                skill_name = "Interrupted!"
+                show_caption(skill_name, "int")
+            else
+                show_caption(skill_name, "ws")
+            end
+        elseif act.category == magic_category and act.actor_id == target then
+            local spell_name =
+                res.spells[act.targets[1].actions[1].param] and res.spells[act.targets[1].actions[1].param].name or
```
z16, line 210: It bothers me a bit that this is broken up into three lines in the WS block, but only two lines in the magic block :\ I'm starting to get the feeling you're using a fixed width and wrap lines exceeding it, in which case I'd vote to just leave them be. Fixed-width line wraps are never a good idea, despite what the Python style guide says :|

addons/finalAlert/finalAlert.lua
```diff
@@ -0,0 +1,303 @@
...
+            end
+        end
+    end
+)
+
+function refresh_backgrounds()
+    create_backgrounds(settings.x_position - 250, settings.y_position)
+end
+
+function create_backgrounds(x, y)
```
z16, line 227: This function uses the `windower.prim.*` API heavily. I'd recommend looking into the `images` library, which is essentially to `windower.prim.*` what the `texts` library is to `windower.text.*`. Just abstracts away some of the boilerplate.

addons/finalAlert/finalAlert.lua
```diff
@@ -0,0 +1,303 @@
...
+        windower.addon_path .. "images/" .. settings.background_size .. "/background_emphasize.png"
+    )
+    windower.prim.set_position(background_emphasize, x, y)
+    windower.prim.set_visibility(background_emphasize, false)
+end
+
+function show_caption(text, type)
+    local event_type
+
+    hide_caption()
```
z16, line 268: Feels weird that you hide it only to show it again. If this is about hiding the currently active box, I'd either maintain which box is currently active in its own variable, or extract the box-hiding code into a new function and call that from both here and `hide_caption`.

addons/finalAlert/finalAlert.lua
```diff
@@ -0,0 +1,303 @@
...
+    showing = true
+    caption:text(text)
+    caption:show()
+
+    if (type == "ws") then
```
z16, line 273: No need for the parantheses here, as well as in the following `if`/`elseif` sections.

z16 (changes requested):
Left some comments, most of them are only suggestions. One thing I didn't mention at any one specific line, because it repeats throughout the addon, please use single quotes for all strings, to keep in line with our current style throughout the addon repo.

PR #2197, opened 2022-06-20
Yush: Added a VerboseKeys setting to enable key display in each macro set.

vessellaj:


addons/Yush/Yush.lua
```diff
@@ -75,12 +77,24 @@ output = function()
     if settings.Verbose then
         names[current] = names[current] or 'Unnamed ' .. tostring(current):sub(8)
 
+        output_text_table = {}
```
z16: This variable should be `local`. I also prefer `output_text_names` over `_table`, which is a bit too generic, and then you could make it a list (`= L{}`) without mangling the meaning of the name. And a list is more fitting here, will make the next part of the code a bit easier.

addons/Yush/Yush.lua
```diff
@@ -75,12 +77,24 @@ output = function()
...
+        table.insert(output_text_table, names[current])
+        if settings.VerboseKeys then
+            for key, val in pairs(current) do
+                if type(val) ~= 'string' then
+                    val = names[val] or 'Unnamed'
```
z16, line 85: I don't have a strong feeling about this, but this could lead to several nested `Unnamed`s, which is a bit meaningless. Instead, you could do the same thing I do above and append `tostring(val):sub(8)`, so it's at least consistent. That said, you could also adjust my previous code and *remove* the `tostring(current):sub(8)` on line 78. Could also live with that, since I don't think appending a table identifier is of much use to anyone. In fact, I'd probably prefer this...

addons/Yush/Yush.lua
```diff
@@ -75,12 +77,24 @@ output = function()
...
+                end
+                table.insert(output_text_table, key_combos[key] .. ': ' .. val)
```
z16: If you made the variable list, as described above, you can do `output_text_names:add(key_combos[key] .. ': ' .. val)` here.

addons/Yush/Yush.lua
```diff
@@ -75,12 +77,24 @@ output = function()
...
+            end
+        end
+        local output_text = table.concat(output_text_table, '\n')
```
z16: If you made the variable list, as described above, you can do `output_text_names:concat('\n')` here. That said, I'm a bit confused about the `\n`. Is multiline-output really desirable for this?

addons/Yush/Yush.lua
```diff
@@ -127,7 +141,9 @@ parse_binds = function(fbinds, top)
 
     rawset(names, top, rawget(_innerG._names, fbinds))
     for key, val in pairs(fbinds) do
+        key_combo = key
```
z16: Either make this variable `local` or even better, change line 145 so it doesn't reassign `key`, but instead using a new variable. This was weird before, so not your fault. But since you're here already, you may as well do me a favor and adjust this :)

addons/Yush/Yush.lua
vessellaj: At least with a text box, I think the multi-line thing looks better, since it makes one command per line. I'm not sure if there's a more preferred way to get a line break.

addons/Yush/Yush.lua
vessellaj: When I try to do this, I get a runtime error:

```
yush: Lua runtime error: yush/yush.lua:87: attempt to call method 'add' (a nil value)
```

This goes the same for when I tried it on the first insert of `names[current]`. I'm not very familiar with Lua so this could easily be a lack of understanding. I've renamed the variable and made it local, as mentioned above.

z16: Did you assign it via `= L{}` (instead of `= {}`)? That's the key part.

addons/Yush/Yush.lua
z16: Hmmm, then maybe concat it depending on the output type:
```lua
        if settings.VerboseOutput == 'Text' then
            label:text(output_text_names:concat('\n'))
        elseif settings.VerboseOutput == 'Chat' then
            log('Changing into macro set %s.':format(output_text_names:concat(' > ')))
        elseif settings.VerboseOutput == 'Console' then
            print('Changing into macro set %s.':format(output_text_names:concat(' > ')))
        end
```

Or whatever separator you think works well for the other two.

addons/Yush/Yush.lua
vessellaj: Yeah. here's what I currently have:

```
        local output_text_names = L{}
        output_text_names:add(names[current])
        if settings.VerboseKeys then
            for key, val in pairs(current) do
                if type(val) ~= 'string' then
                    val = names[val] or 'Unnamed'
                end
                output_text_names:add(key_combos[key] .. ': ' .. val)
            end
        end
        local output_text = output_text_names:concat('\n')
```

Also, what is the L{} thing actually called? I've been trying to do some searching for it and the S{} assignment, and the closest I can tell is S{} is a set and L{} is a list, but I haven't had any luck confirming.

vessellaj: Found it. The method is `append`.

vessellaj:
I'm not sure if there's a clean way to make tables stay in their insertion order. For example, my bind tables have binds in numeric order, but the output text is in hashmap order. It's not a big problem, but it would be a nice improvement.

addons/Yush/Yush.lua
z16: Correct, this was my mistake (both now, and when I "designed" that API...). And yes, `L{}` is a list, i.e. for a contiguous, ordered collection of elements, and `S{}` for a set, i.e. unordered, unique collection of elements. The respective files are `addons/libs/lists.lua` and `addons/libs/sets.lua`.

z16:
> I'm not sure if there's a clean way to make tables stay in their insertion order. For example, my bind tables have binds in numeric order, but the output text is in hashmap order. It's not a big problem, but it would be a nice improvement.

This is essentially the difference between using `L{}` and `{}` (or `T{}`). Regular tables and `T{}` are hash maps, whereas `L{}` is specifically for ordered entries, but has no key-to-value mapping. If you're using Discord you can join us [here](https://discord.gg/aUrHCvk) and give yourself a dev role, to access our development channels, it's easier to discuss these things there and answer any further questions you have. Otherwise feel free to ask here.

PR #2199, opened 2022-06-24
[Gearswap] Add aftercast equip trigger to fallback if the action isn't in the command registry

WinterSolstice8:
This fixes not swapping back to aftercast when an addon edits the outgoing packet 0x01A because gearswap isn't aware the incoming aftercast packet has a different param than it has registered in the command list.

[This ](https://github.com/Roland-J/Windower-Addons/tree/main/NukeDemoter) addon in particular is the addon in question that modifies 0x01A's param.

### DIFF
--- a/addons/GearSwap/triggers.lua
+++ b/addons/GearSwap/triggers.lua
@@ -256,6 +256,7 @@ parse.i[0x028] = function (data)
             equip_sets(prefix..'aftercast',ts,spell)
         elseif debugging.command_registry then
             msg.debugging('Hitting Aftercast without detecting an entry in command_registry')
+			equip_sets(prefix..'aftercast',nil,spell)
         end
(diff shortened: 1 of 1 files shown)


RubenatorX:
This is not the proper solution for this problem.
There are more things in play than just switching to aftercast whenever you hit any aftercast packet, as this would probably make it so that spamming an action could result in you swapping to aftercast in the middle of that same cast for example.
But this problem also involves the fact that gearswap is incapable of being the guaranteed last addon to check all incoming/outgoing actions.
So we cannot accept this pull request as is.

RubenatorX:
Your best bet for now @WinterSolstice8 would probably be to provide a gearswap library that could be included in scripts that would provide a function that could do the demotion in the gearswap logic instead.

WinterSolstice8:
Yes, after discussion that seems to be the only proper solution (until v5).

PR #2204, opened 2022-07-06
Trusts: Added listnames setting.

vessellaj:


RubenatorX:
Wait so... if I'm following this code properly,
You have added a setting (with no command to modify it) that makes it so that the `list` command spits out all trusts that are in all the sets?

It would be better if instead of that, the list command had an optional subcommand that would list out the trusts from a named set -- e.g.: `trusts list vw` 

And then if for some reason you really really wanted to list out basically the entire contents of the settings file (for that character), could add a special flag -- maybe `-verbose` or something (maybe @z16 has thoughts).

But I don't think adding a hidden(read: have to go directly edit the settings file) setting is the way to go for this.

vessellaj:
Fair enough. I've added a toggle to the main commands for the name printing and a subcommand option for `trusts list` as asked.

PR #2209, opened 2022-07-09
Implement Fully-Featured Event Self-Registration for Images & Texts

Roland-J:
Adds the plumbing required for images to support image:register_event('left_click', function) and image:register_event('drag', function), potentially making images much more friendly to work with by lowering the technical barrier to utilizing click events for images. _(This functionality could also be extended to images if this PR is found desirable)_

Sample addon for demo testing:
image

```
--------------------------------------------------------------------------------------------------------------
-- 1. Nonddraggable-image event self(:)-registration demo
--------------------------------------------------------------------------------------------------------------

-- Setup an image (nothing new here)
local images = require('images')
local image2 = images.new()
image2:pos(50, 75)
image2:size(175, 23)
image2:fit(false)
image2:path(windower.addon_path .. 'Button002.png')
image2:draggable(false)
image2:show()

-- Self(:)-register a left-click event (NEW!)
image2:register_event('left_click', function(...)
    print('I\m not draggable and I was just left_clicked!')
end)



--------------------------------------------------------------------------------------------------------------
-- 2. Draggable-image event self(:)-registration demo
-- NOTE: We only fire the left_click event on release and only if there was no dragging
--------------------------------------------------------------------------------------------------------------

local image = images.new()
image:pos(50, 50)
image:size(175, 23)
image:fit(false)
image:path(windower.addon_path .. 'Button002.png')
image:draggable(true)
image:show()

-- Self(:)-register a left-click event (NEW!)
image:register_event('left_click', function(...)
    print('I\'m draggable and I\'ve been clicked and unclicked, without being dragged!')
end)

-- Self(:)-register a drag event! (NEW!)
image:register_event('drag', function(...)
    local curr_x, curr_y = image:pos()
    local last_x, last_y = image:last_pos() --new function: provides the pos_x and pos_y of previous drag's final position
    local x_delta, y_delta = curr_x - last_x, curr_y - last_y --how to get current drag's precise x/y deltas, can be used to move clustered elements
    print('I\'ve been dragged and my delta is ' .. x_delta .. '/' .. y_delta .. '!')
end)
```


addons/libs/images.lua
```diff
@@ -244,6 +246,15 @@ function images.pos_y(t, y)
...
 end
 
+function images.last_pos(t, x, y)
+    if x == nil then
+        return last_pos.x, last_pos.y
+    end
+
+    last_pos.x = x
+    last_pos.y = y
+end
```
z16, line 256: As discussed, this should be removed. The local `last_pos` variable should be used instead, similar to the `dragged` variable.

addons/libs/images.lua
```diff
@@ -10,6 +10,8 @@ local meta = {}
 
 saved_images = {}
 local dragged
+local clicked
+local last_pos = {}
```
Roland-J, line 14: Declared this as a table right away, unlike dragged, so that line 415 can use the new function to assign the values without erroring out due to the function trying to declare keys on a var that was not yet declared as a table.

addons/libs/images.lua
```diff
@@ -407,6 +426,9 @@ windower.register_event('mouse', function(type, x, y, delta, blocked)
             if meta[dragged.image].root_settings then
                 config.save(meta[dragged.image].root_settings)
             end
+            if x == clicked.x and y == clicked.y then
```
z16, line 429: I wonder if we need to be this exact with the click. Maybe a small tolerance is fine? Also maybe wrap it into a `settings.clickable` option?

Either way, this is currently under the `if dragged` condition, this should be moved outside in an analogous `if clicked` condition.

addons/libs/images.lua
```diff
@@ -244,6 +246,15 @@ function images.pos_y(t, y)
     t:pos(meta[t].settings.pos.x, y)
 end
 
+function images.last_pos(t, x, y)
```
Roland-J, line 249: Created this function in order to A) **(most importantly)** expose this local variable to addons and to B) imitate the get/set of the existing pos methods: _images.pos(), images.pos_x(), and images.pos_y()_.

addons/libs/images.lua
```diff
@@ -407,6 +426,9 @@ windower.register_event('mouse', function(type, x, y, delta, blocked)
             if meta[dragged.image].root_settings then
                 config.save(meta[dragged.image].root_settings)
             end
+            if x == clicked.x and y == clicked.y then
+                call_events(dragged.image, 'left_click')
```
z16, line 430: I was debating whether we should leave it as `left_click` or just call it `click` and have `right_click` as the odd one out, because it's used more rarely. I'm not sure though, would be up for discussion. Could also alias it so `click` and `left_click` are identical in behavior.

addons/libs/images.lua
```diff
@@ -407,6 +426,9 @@ windower.register_event('mouse', function(type, x, y, delta, blocked)
...
+            end
             dragged = nil
```
z16, line 432: Like `dragged` before, `clicked` should be unset here.

addons/libs/images.lua
```diff
@@ -387,17 +398,25 @@ windower.register_event('mouse', function(type, x, y, delta, blocked)
     if type == 0 then
         if dragged then
             dragged.image:pos(x - dragged.x, y - dragged.y)
+            call_events(dragged.image, 'drag')
+            dragged.image:last_pos(dragged.image:pos())
```
Roland-J, line 402: This needs to be called after the call_events, not before, in order for the event to retain access to the previous last_pos. Update this after the event so that the next drag event has an updated last_pos.

z16 (changes requested):

addons/libs/images.lua
Roland-J, line 429: We are only registering left_click events for draggable images if the image wasn't dragged between click and release.

addons/libs/images.lua
```diff
@@ -387,17 +398,25 @@ windower.register_event('mouse', function(type, x, y, delta, blocked)
...
             return true
         end
 
     -- Mouse left click
     elseif type == 1 then
         for _, t in pairs(saved_images) do
```
Roland-J, line 408: **SEPARATE PR:** If we wanted to, this could be changed to a backwards iteration that, on the first match, does all the existing stuff but also breaks out. This may be useful for addons that have stacked images, to make the click event only fire for the addon's top-most image in the mouse's x/y region.

Roland-J (commented):
Added some clarifying comments

addons/libs/images.lua
Roland-J: I should be able to easily add support for right click, too. Would it be silly to assume that type 3 = right click and type 4 = right release? I would prefer adding that functionality while we're at it, for the community's sake.

1. Please advise the types (integer) for the right click/release events. (I don't see them documented anywhere.)
2. Please advise how I would setup the alias, if you decide that would be desired. Thanks!

addons/libs/images.lua
Roland-J: Would an average delta of 2 be an ideal tolerance? Using my sample addon, I see that most of my gradual drags have an average delta of 0.5. 

I could change it to this:
`if (math.abs(x - clicked.x) + math.abs(y - clicked.y)) / 2 <= 2 then`
or (maybe safer)
`if (math.abs(x - clicked.x) + math.abs(y - clicked.y)) / 2 <= 1 then`

Please advise how you want the tolerance calculated. Thanks!

addons/libs/images.lua
Roland-J: How can the drag delta be calculated, from the addon's scope, without this helper function? My addon-side drag delta calculations fail without this, as my addon sees the  images.lua's last_pos var as a nil value; _addons appear unable to directly access last_pos_.

Please advise...
1) This bridges the gap, exposing it for addons to read. Is there a better way to bridge that gap?
2) How can we implement your requested change without breaking the ability for addons to calculate drag deltas?

addons/libs/images.lua
Roland-J: Also, I placed it in the `if dragged` condition because all draggable objects would go into that if branch. I though that, since undraggable objects can't be moved, it'd be **unnecessary overhead** to make these undraggable objects get/set the clicked var.

Please advise if, following this explanation, you still find that this needs to change. Thanks!

addons/libs/images.lua
Roland-J: Thanks, done!

Roland-J:
@z16 Thank you for the feedback! I am trying to implement the changes but need some clarifications. Please see my follow up questions. Thanks!

addons/libs/images.lua
z16: Addons should not calculate deltas. That is the lib's job. I'm still not convinced about the validity of this solution for your use case, but even in this case, the deltas could be sent as event parameters. I feel like this could be handled better in general, though. I'll have to think about this.

addons/libs/images.lua
z16: > Using my sample addon, I see that most of my gradual drags have an average delta of 0.5.

Huh? 0.5? How are you measuring delta, by pixel value? If so, an average may be the wrong tool. I'd use something like the 90th percentile. And even then I'd add something like a flat amount to that, to cover people who may be less accurate than you, in general.

There might be an overall better solution to this, but not sure. You seem to want to combine being draggable with being clickable, which makes this problem unnecessarily hard. Ideally I would solve those two problems separately and assume that being clickable and being draggable are mutually exclusive. The more I think about the more I think we should go that route. Would make this easy to solve - if `clicked` is not `nil` just check on mouse release if the mouse cursor is still within the text bounds, and if so, trigger the event. If you still want your behavior (which I really don't think you should, because it's user-unfriendly) you'd have to reimplement that part yourself.

> Also, I placed it in the if dragged condition because all draggable objects would go into that if branch.

This feature is unrelated to drag, even in your use-case, and so should be handled separately.

addons/libs/images.lua
z16: > Please advise the types (integer) for the right click/release events. (I don't see them documented anywhere.)

No idea, easy for you to test. Just `windower.register_event('mouse', print)`, then right click. For reference, it's the `wParam` parameter of [`MouseProc`](https://docs.microsoft.com/en-us/previous-versions/windows/desktop/legacy/ms644988(v=vs.85)), `& 0xFF`. Whatever that is.

> Please advise how I would setup the alias, if you decide that would be desired.

Actually, I'm tending towards leaving it as `left_click` here, but also handling `click` by looking up a mapping table when registering it. Can define a table like this:
```lua
local event_map = {
    click = 'left_click',
    double_click = 'double_left_click',
}
```

Then use this when registering the event, at the very top of the function:
```lua
    key = event_map[key] or key
```

addons/libs/images.lua
z16: Addons should not be able to get that information at all outside of the event itself. Even then it's debatable.

addons/libs/images.lua
z16: Check `if clicked then` later on, this is a bug and only works because Lua is weakly typed.

addons/libs/images.lua
Roland-J: AFAIK, when people are watching drag events, they are using it to keep objects in sync. In such a case, they are using a delta in one form or another.

**A. Top-Left Delta:** _(More complicated? Is this more reliable?)_
1. Find the current image's x/y distance from a bounding rectangle's top-left corner. (one that covers all the shapes)
2. Subtract x/y distance from the last-known position of the bounding rectangle to get the movement delta.

**B. Mouse-Movement Delta** _(Simpler? Is this as reliable?)_
1. Get the last known mouse x/y position
2. Subtract it from the current mouse x/y position to get the delta.

If drag events are going to be doing those calculations either way - _meaning the same overhead_ -  and getting the same result - _the same exact x/y delta, just calculated two different ways_ - it may be nice to provide the simpler method. Same overhead, but seems more user friendly for the addon authors, could be a nice service delivery for the dev base.

Passing it in as an event arg sounds convenient for the addon authors. Would that muck with anything else, though?

addons/libs/images.lua
Roland-J: > > Using my sample addon, I see that most of my gradual drags have an average delta of 0.5.
> 
> Huh? 0.5? How are you measuring delta, by pixel value? If so, an average may be the wrong tool. I'd use something like the 90th percentile. And even then I'd add something like a flat amount to that, to cover people who may be less accurate than you, in general.
> 

Yes, I am using pixel values. "_The mouse moved this many pixels on the x axis and this many pixels on the Y axis._" I'm not opposed to measuring it another way, and adding a flat amount sounds fine.

> There might be an overall better solution to this, but not sure. You seem to want to combine being draggable with being clickable, which makes this problem unnecessarily hard. Ideally I would solve those two problems separately and assume that being clickable and being draggable are mutually exclusive. The more I think about the more I think we should go that route. Would make this easy to solve - if `clicked` is not `nil` just check on mouse release if the mouse cursor is still within the text bounds, and if so, trigger the event. If you still want your behavior (which I really don't think you should, because it's user-unfriendly) you'd have to reimplement that part yourself.
> 

I see what you mean. When typing this comment, if I click down in the middle of the GitHub **Cancel** box, and then drag to the edge - _but still within the box_ - and release, it registers as clicking the box. This is a static setup, though.

On the other hand, browser favorites are similar to my implementation, they just have a healthy tolerance before moving from a click/release to a click/drag scenario.

https://user-images.githubusercontent.com/107378114/178127685-2f46e81d-0946-4605-80b3-2bef6a10d412.mp4

I can't say either method is better than the other. I think it's just a matter of taste/context.

Why don't we provide both? We could let draggable objects default to your suggestion - _no left_click events_ - but transition to the browser's left_click-if-under-threshold/drag-if-over-threshold behavior if the user adds a `image:clickable(true)` to an already draggable image. Viola, users can have it both ways.

For undraggable objects, see my next quote section. 

Thoughts?


> > Also, I placed it in the if dragged condition because all draggable objects would go into that if branch.
> 
> This feature is unrelated to drag, even in your use-case, and so should be handled separately.

Yeah, if it would be better that way, let's go for it! After seeing how GitHub does their cancel button - _click event fires on release if mouse is still anywhere in the button_ - I think I tried to oversimplify undraggable object clicks.

For these images, I think I should add code that verifies on unclick that the mouse is still within the bounds of the originally clicked button, and run the events only if it is. That way users can cancel the click event, by dragging out before releasing, incase the original clickdown was an accident/wrong button.

Sounds like a better experience, what do you think?

addons/libs/images.lua
Roland-J: Drag events are probably going to be used to compute, and then act on, said delta. So it may be a great place to make it available.

addons/libs/images.lua
z16: > On the other hand, browser favorites are similar to my implementation, they just have a healthy tolerance before moving from a click/release to a click/drag scenario.

This is different because there are discrete places to move it to. It's not a continuous space you can move it around in, but only fixed places before and after other elements, so even massive tolerances before registering a click as a drag are fine. But in our case we either have to remove any tolerance, which would result in people occasionally moving a UI object accidentally instead of clicking it (which is not only annoying if it happens in an important situation, such as a battle, but it also destroys the UI layout, which, for people with certain levels of OCD, like me, is an even greater annoyance) or we make the tolerance big enough to tolerate cases like that, but make it significantly harder to fine-adjust UI elements (which, again, for people with certain OCD levels, like me, is a real annoyance).

> Why don't we provide both? We could let draggable objects default to your suggestion - no *left_click* events - but transition to the browser's left_click-if-under-threshold/drag-if-over-threshold behavior if the user adds a `image:clickable(true)` to an already draggable image. Viola, users can have it both ways.

I'm willing to let myself be convinced that this is useful - at this point I just don't see it. I know you seem to want it, though I have to say I still don't understand why, but I think we should ask people in the channel what they think. It would add noise to our API and I just don't see the benefit. Especially since this is something you could add on your own outside of the API.

That said, I'd be more willing to do it without the tolerance, and just let users mess up their UI if they're not careful. At least that way we don't have to have any special handling in our API and we can just blame the addon devs.

addons/libs/images.lua
z16: > AFAIK, when people are watching drag events, they are using it to keep objects in sync.

You're literally the only one doing this, because this feature doesn't even exist yet :P

> Passing it in as an event arg sounds convenient for the addon authors. Would that muck with anything else, though?

No, but I still don't like it, this could be handled entirely in addon code. You'd have to keep track of the objects' original positions, then compare against the current, and adjust accordingly. Even with multiple directly draggable objects it wouldn't be too hard to do.

The reason this bothers me so much is because this feels out of scope for this lib, it's what a widget lib would be for. This lib only concerns individual objects, nothing outside of it, so your use-case of synchronizing with other elements feels like a weak justification.

As for your scenarios, only B makes sense. A is not more reliable. We only report discrete pixel values and they're fully accurate (to the pixel level).

Dev-Roland: > > AFAIK, when people are watching drag events, they are using it to keep objects in sync.
> 
> You're literally the only one doing this, because this feature doesn't even exist yet :P

Why do you say that I'm the only one who syncs objects during drag? [As can be seen here,](https://discord.com/channels/338590234235371531/501099842097905675/994989166377050253) Akirane also does it, and there are likely others that do it as well that we're not aware of. It's not factual to say that I'm the only one, unless you're going on the technicality of _"using image:register_event('drag', fn)"_ vs _"windower.register_event('mouse', fn)"_, which would be missing the point if so.

As far as I can tell, if someone's watching drag then they'll be interested in the deltas - _that's what Akirane, probably unknown others, I are using the drag event for_ - as I don't yet see what else is important for an addon author during a drag.

> > Passing it in as an event arg sounds convenient for the addon authors. Would that muck with anything else, though?
> 
> No[...]

Great! Well, if the drag delta is the only important thing during a drag event, and we strip that delta from the event, we may be gutting the drag event from its usefulness.

> The reason this bothers me so much is because this feels out of scope for this lib, it's what a widget lib would be for. This lib only concerns individual objects, nothing outside of it, so your use-case of synchronizing with other elements feels like a weak justification.

Would a widget lib let the author bind (or whatever technical term) a bunch of images/texts into a cluster and then the widget lib would move them in tandem? That seems like a nice lib to have! I would agree that, if something like that is offered, that would seem like a better place for this. But, until then, what? The drag event could meet people's needs, if we build it to provide them the delta.

> We only report discrete pixel values and they're fully accurate (to the pixel level).

Great to hear!

addons/libs/images.lua
Dev-Roland: Well, if we don't add the tolerance-before-dragging to (the new) draggable-and-clickable objects, addon authors won't be able to add that tolerance before dragging to their addons. That is because A) the images.lua is what is forcing all draggable objects to go into drag mode the moment the mouse is clicked down and B) requiring their userbase to modify their local images.luas to use their addons is likely a death wish for any hopes they had for users adopting their addon.

We would need to add that logic where, for the users who did `image:draggable(true)` and then `image:clickable(true)`, the drag event doesn't start on click until the mouse moves out of the tolerance zone. As is, the drag event starts immediately on click.

addons/libs/images.lua
Roland-J: Akirane also does the same thing in one of their addons, as well as likely unknown others. However, this does seem like it would better belong in a _(not-yet-available)_ widgets lib, like you said. I guess for the meantime such authors should continue doing it themselves, right? Rather than shoehorning it in here, that is.

addons/libs/images.lua
Dev-Roland: If an author such as myself wanted to add this feature - _a tolerance before the drag event begins on draggable objects, where only a click event fires on release if the tolerance is not exceeded_ - would we be able to do that by packaging our own images.lua with our addon?

Roland-J: If an author such as myself wanted to add this feature - _a tolerance before the drag event begins on draggable objects, where only a click event fires on release if the tolerance is not exceeded_ - would we be able to do that by packaging our own images.lua with our addon? I would think that would work.

What I'm wanting, like you said, seems to belong more in a new widgets.lua library, right?

Roland-J:
@z16 Sorry for the delay, but I've implemented much of what was mentioned, plus a lot more. It can be found and demo'd in a fully functional form here: _(feel free to re-use this code)_
https://github.com/Roland-J/SmartSkillup/blob/main/libs/sms_images.lua
https://github.com/Roland-J/SmartSkillup/blob/main/libs/sms_texts.lua

It's a lot of changes - _trying to flesh out the event self-registration feature as much as possible to lower the barrier to UI development and encourage more UI development_ - so there may be too many conflicts to event attempt it in the PR, so I figured I'd first mention it before merging it in. I'll take my next action according to the feedback.

Summary:
1. Added hover/unhover event. _(Only fires if said event has been registered for said element)_
2. Added scroll up/down events. _(Only fires if said event(s) have been registered ffor said element)_
3. Added scroll click event. _(Only fires if said event has been registered for said element)_
4. Added right click/release event. _(Only fires if said event has been registered for said element)_
5. Added right drag.  _(Only occurs if its new flag is enabled.)_
6. Added right drag event. _(Only fires if said event has been registered for said element)_
7. Added scroll drag.  _(Only occurs if its new flag is enabled.)_
8. Added scroll drag event. _(Only fires if said event has been registered for said element)_
9. Differentiated left drag from right drag. (Renamed vars & func and provided alias vars and funcs for legacy support)
10. Finished the drag tolerance feature. _(Defaults to 0, drags only start once tolerance is exceeded)_
11. Refined event sinking with event checking - keep going to first event-registered element - to allow events to sink through a irrelevant hitbox or similar item into the topmost relevant item.
12. Added event returning - _return out on first relevant element_ - to avoid allowing for both the intended (topmost) element and unintended (lower) elements to trigger events; only the intended (topmost) one will trigger.

I've added some event variables to them. You can see what I'm doing with those variables in https://github.com/Roland-J/SmartSkillup/blob/main/libs/sms_ui.lua

Nifim:
Can you update this PR? we had a change to our repo that caused discontinuity between the various forks.

Once you do I can review your code and get this back on track to being merged.

PR #2216, opened 2022-07-28
Config: Allow multiple 'text' children in a single XML tag.

Imisnew2:
I wanted to add comments to CDATA sections I manually added to a config file. This was giving me a 'malformed config' error, so I tried to fix the config library to allow this.

### DIFF
--- a/addons/libs/config.lua
+++ b/addons/libs/config.lua
@@ -277,19 +277,31 @@ function settings_table(node, settings, key, meta)
         return t
     end
 
-    if not node.children:all(function(n)
+    local function tag_or_comment(n)
         return n.type == 'tag' or n.type == 'comment'
-    end) and not (#node.children == 1 and node.children[1].type == 'text') then
+    end
+    local function text_or_comment(n)
+        return n.type == 'text' or n.type == 'comment'
+    end
+
+    if not node.children:all(tag_or_comment) and not node.children:all(text_or_comment) then
         error('Malformatted settings file.')
         return t
     end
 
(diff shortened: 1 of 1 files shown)


addons/libs/config.lua
```diff
@@ -277,19 +277,31 @@ function settings_table(node, settings, key, meta)
...
     end
 
     -- TODO: Type checking necessary? merge should take care of that.
-    if #node.children == 1 and node.children[1].type == 'text' then
-        local val = node.children[1].value
-        if node.children[1].cdata then
+    if #node.children >= 1 and node.children:all(text_or_comment) then
+        if node.children:any(function(n) return n.cdata end) then
             meta.cdata:add(key)
-            return val
```
Imisnew2, line 292: I'm not really sure what `meta.cdata` is tracking or what features it supports. I just added the key once if any of the text nodes had cdata. Before, it was asserting only one node could exist, so the condition used to be a simple 'if that one node has cdata' instead of 'if any node has cdata'.

addons/libs/config.lua
```diff
@@ -277,19 +277,31 @@ function settings_table(node, settings, key, meta)
...
+        end
+
+        local val = ''
+        for child in node.children:it() do
+            if child.type == 'comment' then
+                meta.comments[key] = child.value:trim()
```
Imisnew2, line 301: This was copied from what the other code path does -- the path when all children are tags or comments. Though it seems like this just ends up taking the value of the last comment vs storing all the comments. I don't know what `meta.comments` is used for.

PR #2221, opened 2022-08-20
Update EmpyPopTracker to 2.7.0

xurion:
Add NMs:

* Amphitrite
* Bennu
* Ironclad smiter
* Ironclad sunderer
* Ironclad triturator
* Resheph

Related to:
* https://github.com/xurion/ffxi-empy-pop-tracker/issues/19
* https://github.com/xurion/ffxi-empy-pop-tracker/pull/18

### DIFF
--- a/addons/EmpyPopTracker/EmpyPopTracker.lua
+++ b/addons/EmpyPopTracker/EmpyPopTracker.lua
@@ -29,7 +29,7 @@ SOFTWARE, EVEN IF ADVISED OF THE POSSIBILITY OF SUCH DAMAGE.
 _addon.name = 'Empy Pop Tracker'
 _addon.author = 'Dean James (Xurion of Bismarck)'
 _addon.commands = { 'ept', 'empypoptracker' }
-_addon.version = '2.5.0'
+_addon.version = '2.7.0'
 
 config = require('config')
 res = require('resources')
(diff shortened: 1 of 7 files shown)


addons/EmpyPopTracker/EmpyPopTracker.lua
```diff
@@ -224,7 +224,12 @@ end
 EmpyPopTracker.generate_info = function(nm, key_items, items)
     return {
         has_all_pops = not nm.pops or T(nm.pops):all(function(item)
-            return item.type == 'item' and owns_item(item.id, items) or owns_key_item(item.id, key_items)
+            -- Avoids confusing item and key item IDs
```
z16, line 227: Am I missing something here? Or is this change meant to be cosmetic only? Because it should do the same thing as before.

xurion: That's what I thought. There's an issue wherein, if the `item.type` is "item" but the `owns_item` returns `false`, it falls back to running the `owns_key_item`. This can give us a false positive when we have a matching key item ID.

z16: Ah, that's a good point. Thanks for clarifying.

PR #2238, opened 2022-10-14
[shortcuts] Pass valid commands through if target and command is valid

WinterSolstice8:
I found the real root cause of #2225

Apparently that was being caused by low player IDs, which explains why it would not happen on retail.

Using the menu to use Cure 1 yourself results in a `/magic "Cure" <id>` but shortcuts interpreted that as cure 5 <me> because the player ID I was had was 5.
image

This PR adds a check for a "complete" normal command with a valid target and passes it straight through. Wanted behavior, such as `//cure 5` results in `/ma "cure v" <me>` as expected.

### DIFF
--- a/addons/shortcuts/shortcuts.lua
+++ b/addons/shortcuts/shortcuts.lua
@@ -290,7 +290,7 @@ function command_logic(original,modified)
             windower.send_command('@input '..lastsent)
             return '',false
         end
-    elseif command2_list[command] then
+    elseif command2_list[command] or (command_list[command] and valid_target(potential_targ)) then
         -- If the submitted command does not require ability interpretation and is fine already, send it out.
         lastsent = ''
         if logging then


Byrth (commented):
I don't want to merge this because it is only necessary due to a private server bug (defined as anything that makes the private server different from retail.)

I recommend fixing the private server instead of adding weird unexplained logic to shortcuts. Lord knows it already has enough.

paulframe:
@Byrth this is incorrect. We have encountered retail players with exceptionally low IDs. It is not a private server only issue. It simply affects more of the private server population than retail.

TeoTwawki:
> I don't want to merge this because it is only necessary due to a private server bug (defined as anything that makes the private server different from retail.)

False. It's just that getting a low id on retail is rare nowadays due to the games age.

> I recommend fixing the private server instead of adding weird unexplained logic to shortcuts. Lord knows it already has enough.

There isn't anything to fix.

WinterSolstice8:
> I recommend fixing the private server instead of adding weird unexplained logic to shortcuts.

An ID 1 >= and <= 6 is not a bug. All this PR is doing is optimizing shortcuts, it already knows the command and target is valid. There is nothing unexplained or weird, unless you would like to enlighten me because I am genuinely curious about that. If anything, this is a PR to optimize shortcuts. You are essentially claiming your own functions are unexplained and weird, shortcuts already knows that the command and the target are valid, I'm just adding a check to pass it through.

Byrth:
Ok then!

PR #2243, opened 2022-11-22
Update roe.lua

ianandersonlol:
block monthly RoEs like timed.


RubenatorX:
Making it so that multiboxers can't copy across valid monthlies using the addon... feels a little extreme. So if we can come up with a less aggressive solution, would be cool.

PR #2253, opened 2022-12-04
Fixed xiv bar to use Windower UI scale values.

cinderborn:
An issue with xivbar prevented it from scaling correctly according to UI Scale set in Windower and cause it to behave incorrectly if the windower scale isn't set to 1.

### DIFF
--- a/addons/xivbar/ui.lua
+++ b/addons/xivbar/ui.lua
@@ -88,8 +88,8 @@ end
 
 -- position the images and text
 function ui:position(theme_options)
-    local x = windower.get_windower_settings().x_res / 2 - (theme_options.total_width / 2) + theme_options.offset_x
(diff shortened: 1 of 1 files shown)


RubenatorX:
The problem with doing this, is it breaks everyone's fixes. (meaning, next time ppl that positioned it properly login, it will suddenly be positioned improperly thanks to the change)

cinderborn:
It might make more sense for people who already fixed it once to position it properly again than others who couldn't make it work in the first place and thought the addon was simply broken before removing it.

EvanSwinehart:
It used the wrong values to begin with, and causes issues when users are setting up for the first time. It would be better to fix it than leave the undocumented incorrect functionality that just causes confusion every time someone sets this up, even if it causes some to have to adjust their settings one time.

PR #2254, opened 2022-12-09
Addon: Chatmon (lua version)

Xabis:
This is a lua rewrite of Chatmon.

What is different from the plugin version:

- Compatible with color code injections (e.g. Battlemod)
- Channel filtering allows any numerical id, not just the provided list of named channels.
- Named channels have been expanded to include Linkshell2, Readies, Wearoff, and Casting.
- Spam filtering functionality is exposed and configurable, so that users may tweak as required.
- Triggers, filters, and settings may be added/updated with commands inside the game, instead of needing to change the xml directly.
- Added a default loadout of sounds to enjoy
- Added a default configuration, which will be used if one is not provided.

Care has been made to keep the original chatmon.xml file backwards compatible as possible.


Xabis:
Note: The changes made to the "fixed" xml lua is documented in this pull request: https://github.com/Windower/Lua/pull/2247

RubenatorX:
I'm glad you have examples in your settings file, but, they should all be commented out.

Xabis:
@RubenatorX, is this because defaults should not be provided, or because the fear that future updates may overwrite someones config?

If the latter, the current logic is to copy "Chatmon-defaults.xml" to "Chatmon.xml" only when it is missing, otherwise these defaults are not used.

If the former, I can update the addon to remove this defaulting behavior, so that it always writes out a basic (aka empty) config.

Xabis:
@RubenatorX, I have renamed "Chatmon-Default.xml" to "Chatmon-Example.xml" and removed all code that uses it by default.

If a user wants to make use of the provided example file, they will need to do so manually.

Let me know if this addresses your concerns.

Xabis:
I no longer play, and it is clear that no further addons will be accepted. Closing.

PR #2260, opened 2022-12-15
InfoBar SQL file proposal

xurion:
This is a proposal to swap from a versioned binary database file to a SQL file that creates the database file on load, allowing better change tracking and collaboration.

The SQL file is marginally larger than the database file (1.8mb vs. 1.75mb).

The addon uses the `data.sql` file to create `data.db` instead of `database.db` in order to prevent overwriting anyone's local changes up to this point.


xurion:
Closing this. It was decided over in Discord that Lua files to contain this data wouldn't be any less efficient. I'm working on converting the data over to that now.

PR #2262, opened 2022-12-19
[craft] make bundled recipes higher recipe numbers than their standard equivalents

posimagi:
Move bundled recipes (those that use Bundling Twine, Spindle, etc.) to higher recipe numbers than the standard recipe that makes the basic quantity.

Fix issues with several recipes that erroneously show multiple copies of the bundling guild item.

PR #2267, opened 2023-01-19
[craft] repair and upgrade create_recipes.py to python3

posimagi:
Also made it robust enough to not break on recipe(s) with unknown ingredients such as Pungent Powder III, though the output is not useful--not that it could be crafted with unknown ingredients anyway.

Note: in its prior state the script did not work at all. There were multiple changes on the wiki that had not been picked up within the last several months and I suspect this was due to the script panicking on input as well as the difficulty of resolving the dependencies in a python2 environment.

### DIFF
--- a/addons/craft/create_recipes.py
+++ b/addons/craft/create_recipes.py
@@ -1,4 +1,4 @@
-import urllib2
+import urllib.request
 from bs4 import BeautifulSoup
 from slpp import slpp as lua
 import os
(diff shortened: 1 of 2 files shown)


posimagi:
This script has been tested and the output from it looks good. That output will be in a separate PR, #2262.

Byrth:
Looks good to me!

PR #2268, opened 2023-01-19
[battlemod] support inundation and Costume status effects

posimagi:
This change makes these debuffs display correctly in battlemod when `simplify` is false (they already work when it is true).

Without battlemod:
costume-stock
Before this change:
costume-before
After this change:
costume-after

Without battlemod:
inundation-stock
After this change:
inundation-after


Nifim:
These changes should be done in the fixes file for resources not explicitly handled in battle mod

Byrth:
(arguably should be fixed by SE)

PR #2271, opened 2023-01-24
Add new Emotes addon

iLVL-Key:
Expands upon the games built-in emotes.

### DIFF
--- a/addons/Emotes/emotes.lua
+++ b/addons/Emotes/emotes.lua
@@ -1,0 +1,532 @@
+--[[
+Copyright © 2023, Key
+All rights reserved.
+
+Redistribution and use in source and binary forms, with or without
+modification, are permitted provided that the following conditions are met:
+
+    * Redistributions of source code must retain the above copyright
+      notice, this list of conditions and the following disclaimer.
+    * Redistributions in binary form must reproduce the above copyright
+      notice, this list of conditions and the following disclaimer in the
+      documentation and/or other materials provided with the distribution.
+    * Neither the name of Emotes nor the names of its contributors may be
+	  used to endorse or promote products derived from this software without
+	  specific prior written permission.
+
+THIS SOFTWARE IS PROVIDED BY THE COPYRIGHT HOLDERS AND CONTRIBUTORS "AS IS" AND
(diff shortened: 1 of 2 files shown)


RubenatorX:
This all seems far too hardcoded and not customizable enough.

I also feel like this should allow users to have better control over what commands override what other commands.
For example...
`/jump` in vanilla does the jump motion.
`/jump` with `shortcuts` addon loaded attempts to do the `Jump` job ability if you're on DRG.
`/jump` with both this addon and shortcuts loaded, would... attempt to do more than one thing at once. (Not that you have overridden jump here, but that its possible that some of your commands could interact with other commands).

Could possibly make use of Windower's `alias` system. @z16 thoughts?

So my suggestion is, have your emotes in a default mapping, but allow the users to edit said mapping to their liking (both command wise and text wise) by saving those to a settings file.

You can use placeholders in your strings such as `{$posessive}` for pronouns if you like, which would make things much more dynamic, and prevent you from doing huge if/elseif chains in your code.

iLVL-Key:
Thanks for the comments. I'm pretty new to coding and this was more of a "proof of concept" that I could even make something, so I definitely will take any comments or suggestions to improve. Initially I wanted to be able to take the games built-in emotes and allow you to change the pronouns used on those official ones, but the Shortcuts addon conflicted with the ability to replace an emote with the new adjusted one, so I settled for just doing that with the new ones I added.

I thought about using the aliases, but my goal was to make the new emotes use a single slash so they feel more like built-in ones.

I think I understand what you're saying with putting the emote text inside the settings file, I like that idea. I'll have to figure out how to make that so people can edit those via commands (again still learning, I only figured out getting the settings file added in today).

I'll also have to look into the placeholds in strings, that sounds much more convenient.

Again thanks for the comments!

RubenatorX:
@iLVL-Key instead of using the `outgoing text` event, try the `unhandled command` event <https://github.com/Windower/Lua/wiki/Events>

Also note that, even if you did change pronouns on the original commands, you would only want to use it if the pronoun was different from the default, because the original emotes are automatically converted to Japanese for Japanese players, so you wouldn't want to break that for people.

PR #2276, opened 2023-02-19
Fix Lightningday spelling

lili-ffxi:


Nifim:
Going to merge this but we should look at if using the resources to get the names

PR #2281, opened 2023-03-08
[InfoBar] Addon -- Added Almost all of the Adoulin Enemy Detects

coldReactive:
Not all enemies are included. Some Reive enemies are missing, and a lot of the wikis/ffxiah/etc. don't include all enemies or their weaknesses. I've also added some of the weaknesses to Adoulin enemies I could find (not Yahse, Ceizak nor Rala Waterways.) I've also included a few more Orcish Grunt enemies to vanilla areas that were bothering me.


coldReactive:
Close https://github.com/Windower/Issues/issues/1040

when committed or otherwise.

PR #2285, opened 2023-03-25
Update Dimmer.lua

lili-ffxi:
Prevents dimmer from trying to equip rings from inactive bags - especially useful when the ring is inside a wardrobe that has not been paid for.


lili-ffxi:
Derp.

PR #2288, opened 2023-04-05
adds ability to drag equipviewer in game

andrewretherford:
- makes background image draggable
- removes equipment images while dragging
- runs setup_ui on left mouse up and saves new position to settings

note: I thought about building in a toggle to disable draggable, but didn't think it a worthwhile addition.  I thought it worth mentioning though in case  you want to include that feature in the live addon.

### DIFF
--- a/addons/equipviewer/equipviewer.lua
+++ b/addons/equipviewer/equipviewer.lua
@@ -185,7 +185,7 @@ end
 local function setup_ui()
     refresh_ui_settings()
     destroy()
-    
+
     bg_image = images.new(bg_image_settings)
     bg_image:show()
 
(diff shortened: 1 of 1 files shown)


andrewretherford:
Found Tako's github so will put in a pull request there

RubenatorX:
1) You're making the potentially false assumption that the BG has not been configured to be fully (or nearly) transparent. -- should instead force it to show as a visible translucent or color while dragging.
2) You're destroying they UI elements rather than just updating their position -- this can be fixed by:
2a) don't copy paste (partial) code from another function (as that is not good practice)
2b) instead, call `hide()` and then specifically call `:show()` on the bg
3) this sort of setup makes it entirely too easy to accidentally click on the UI and have it instantly move out of position. -- instead, should have a command that toggles a "move" mode where it becomes draggable.

andrewretherford:
1. I'm not making any assumptions, I noticed that case and looked at a way to solve for it by setting the bg to the default 0 0 0 72, but didn't see a quick fix for it given that I was destroying the elements and then re-creating them with the setup_ui function. In the end, after testing, I decided that the functionality is acceptable even with a transparent background. It's not hard to drag the window to where you want it in small steps, or if the user really cares, they can change their background while re-positioning.
2. hide/show is a good idea, I'll take a look at it tomorrow if you're interested in the code.  If not I'm good with this implementation for my own use.
3. In my pull request I noted this and explained why I didn't implement it

RubenatorX:
@andrewretherford 
1) right, you can change it and itll get put back when you refresh the ui settings
2) I'm the addon author so, I'm certainly interested in the code
3) sure, but I won't accept a drag feature without also having a toggle
More than happy to help and review code in the official discord if you want help/feedback -- otherwise, appreciate you taking the time to submit changes, but those are the changes I'd need to accept the PR

RubenatorX:
Oh and as an aside, a non-trivial number of ppl requested the ability to change the bg to transparent when the addon first came out so, I can't in good conscience accept it being both invisible and draggable at the same time.

RubenatorX:
For extra brownie points, make it scalable with scroll somehow ;p (maybe ctrl+hover+scroll or something like that)

PR #2314, opened 2023-08-31
BLUAlert Additional Features

PrinceHeadache:
Hi! This is a version of BLUAlert I adjusted for my own use, but I thought it offered some nice features that make it worth merging.

Firstly, the addon now plays a sound both when an unknown blue magic spell is used, as well as when the player learns a spell. That way, you don't have to keep a close eye on the chat log or keep your blue magic menu open to verify that you're done learning the spell.

Secondly, I replaced the default text to speech voice that notifies you when unknown blue magic has been used with a Creative Commons-licensed sound effect, as well as another CC sound effect for when you learn a spell.

Finally, I adjusted the readme to account for the new feature, and bumped the version number up to 1.1.0.

Thanks a bunch! Please let me know if you have any questions or comments.


RubenatorX:
I'm down to merge this after you modify the `wav`s to be like, half that volume

PrinceHeadache:
> I'm down to merge this after you modify the `wav`s to be like, half that volume

I'd be happy to! I'll make the change ASAP. Thanks! 

Edit: Hit the wrong button, did not mean to close this.

addons/BLUAlert/BLUAlert.lua
```diff
@@ -76,6 +73,7 @@ function get_action_id(targets)
     end
 end
 
+-- Register the action event listener
```
z16, line 76: This comment doesn't really add much info that isn't immediately obvious from the next line - I would remove it.

addons/BLUAlert/BLUAlert.lua
```diff
@@ -87,3 +85,14 @@ windower.register_event('action', function(action)
         end
     end
 end)
+
+--Prince Headache's Modifications (Ver 1.1.0):
+
+-- Plays a sound when Blue Magic is learned
+windower.register_event('incoming text', function(original, modified, original_mode, modified_mode, blocked)
+    -- Check if the modified message contains "learns" followed by a Blue Magic spell name
+    local spell_name = modified:match("learns%s+(.+)!")
```
z16, line 94: I'd also add a `%s+` in front of the `learns` string. Furthermore, this should match the `original` text, not the `modified` one. It's possible some addon modifies what the text looks like and changes the `learns` word (for example, something like: "Spell learned: <name>"), then it would no longer match it. Using `original` should make it work reliably in any case.

Finally, I'd also add a check for the `original_mode`. Otherwise it would trigger even when someone writes something in a party chat, for example.

addons/BLUAlert/sounds/UnknownBlueMagicUsed.wav
z16, line 1: I would not change this file, but instead add a new one, and let users choose which to pick via the settings.

z16 (changes requested):
See my notes in the code. The most crucial part is probably the setting to let people choose which sound effect to use. This may not sound important, but people really hate it when things change and they don't know why.

RubenatorX:
While you're at it, if you could halve the volume of the original sound effect that would also be great ^_^;

PrinceHeadache:
Thanks for all the feedback! I'll get to work on these changes ASAP.

Totally makes sense to make it a setting so it's an optional change. I hate the tts voice so I hadn't considered that someone would be unhappy to see it changed, but you're right that it ought to be an opt-in. Will modify it soon!

PR #2319, opened 2023-10-03
[barfiller] Fix EXP rate calculation

nmarigoni:
Fixed barfiller's EXP rate calculation to recognize gains from EXP chain messages, and to use current time rather than CPU time for calculations.

### DIFF
--- a/addons/barfiller/statics.lua
+++ b/addons/barfiller/statics.lua
@@ -234,8 +234,8 @@ function display_help()
 end
 
 function exp_msg(val,msg)
-    local t = os.clock()
-    if msg == 8 or msg == 105 then
+    local t = os.time()
+    if msg == 8 or msg == 105 or msg ==253 then
         xp.registry[t] = (xp.registry[t] or 0) + val
         xp.current = math.min(xp.current + val,55999)
         if xp.current > xp.tnl then
@@ -246,12 +246,12 @@ function exp_msg(val,msg)
 end
 
 function analyze_points_table(tab)
-    local t = os.clock()
+    local t = os.time()
     local running_total = 0
-    local maximum_timestamp = 29
+    local maximum_timestamp = 0
     for ts,points in pairs(tab) do
         local time_diff = t - ts
-        if t - ts > 600 then
+        if time_diff > 600 then
             tab[ts] = nil
         else
             running_total = running_total + points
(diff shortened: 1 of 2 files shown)


RubenatorX:
edit: actually I'm mistaken (was corrected by another dev) -- they are the same units, but clock has milliseconds so nvm on all that. You can change it to os.time if you like, seems sufficient for this use case
~~1) clock and time have completely different units, you can't just swap them out and expect them to work as intended when for example compared to `600` (e.g. line `254`)
2) on windows, the os clock function actually just returns the time (rather than clock cycles) anyway.
3) os.time is in sections, and therefore not as granular as os.clock, so if you want to change it to anything, you'd probably use `sockets.gettime()` which has millisecond precision (although per 1 & 2, shouldn't need to change it at all necesarilly)~~

addons/barfiller/statics.lua
```diff
@@ -234,8 +234,8 @@ function display_help()
 end
 
 function exp_msg(val,msg)
-    local t = os.clock()
-    if msg == 8 or msg == 105 then
+    local t = os.time()
+    if msg == 8 or msg == 105 or msg ==253 then
```
lili-ffxi: Missing space between `==` and the value

RubenatorX:
I'm mistaken, adjusted comment above to reflect correction
Although point number 2 is correct.
We are unsure if there is a difference in the `os.clock` implementation in Wine (for the Linux users out there), but os.time should be "more correct" for Linux at least

nmarigoni:
> I'm mistaken, adjusted comment above to reflect correction Although point number 2 is correct. We are unsure if there is a difference in the `os.clock` implementation in Wine (for the Linux users out there), but os.time should be "more correct" for Linux at least

Thanks for the feedback. The change to os.time was intended to be a housekeeping rather than functional change, as the Lua documentation did suggest is was the more appropriate function. I was not aware of the difference in functionality between platforms and am happy to defer if there are any concerns. However, it sounds like the view at this time is that using os.time is appropriate, so I'll leave as-is absent further feedback on this point.

addons/barfiller/statics.lua
nmarigoni: Thanks! Has been corrected.

PR #2322, opened 2023-11-01
Update blist.lua

greezemunkee:
Fixes blist addon not properly blocking /yell messages


paulframe:
What exactly does this change?
The order of items in square brackets should make no difference.

RubenatorX:
The order of items in square brackets has no effect on the result so, what exactly does this change change?

greezemunkee:
This appeared to fix an issue that I was having with the plugin. I reverted my changes and tried to reproduce my error and it appears to be working as it should. I guess I was just using it wrong.  You can remove this.

PR #2339, opened 2024-03-04
Update send.lua

lili-ffxi:
Adjust string processing to reinstate double quotes removed by the command processor.

PR #2342, opened 2024-04-03
Added @party support

zmsl:
@party allows users to send commands to just characters in their party


z16:
While I appreciate the PR, this conflicts heavily with another Send PR that we were about to accept (#2339). I will accept the other PR first, because it's important (and well, it was first), but feel free to adjust yours accordingly, or open a new one, if that's easier for you.

PR #2346, opened 2024-04-05
[gearswap] Update export.lua

lili-ffxi:
Removed some code duplication and reordered some other code.

Added a new `copy`/`clipboard` flag, which makes GS export the table to clipboard instead of to a file.

INPUT NEEDED:
I also added two new flags, `noaugments` and `onlyaugmented` which respectively do:
- `noaugments`/`noaugs` exports the set without recording the augments on each item. Useful if you mostly have single-path items so you don't get a spam of `augments={'Path: A'}` and such; or if you already have augmented items saved.
- `onlyaugmented`/`onlyaugs` *only* exports items that have augments and ignores entirely non-augmented items. Useful if you need to make variables for your augmented gear, or in combination with the `wearable` flag to export all augmented items that your current job can equip, to make a new gearswap file, etc.

The issue I have is that I would like gearswap to print out a notice when provided one of these flags, but I can't come up with good ones myself 😄 input appreciated

### DIFF
--- a/addons/GearSwap/export.lua
+++ b/addons/GearSwap/export.lua
@@ -24,9 +24,11 @@
 --(INCLUDING NEGLIGENCE OR OTHERWISE) ARISING IN ANY WAY OUT OF THE USE OF THIS
 --SOFTWARE, EVEN IF ADVISED OF THE POSSIBILITY OF SUCH DAMAGE.
 
+require('logger')
 function export_set(options)
     local item_list = T{}
-    local targinv,all_items,minify,all_sets,use_job_in_filename,use_subjob_in_filename,overwrite_existing,named_file
+    local targinv,all_items,wearable,all_sets,noaugments,onlyaugmented,minify,clipboard
+    local use_job_in_filename,use_subjob_in_filename,overwrite_existing,named_file
     if #options > 0 then
         for _,v in ipairs(options) do
             if S{'inventory','inv','i'}:contains(v:lower()) then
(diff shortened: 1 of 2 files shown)


addons/GearSwap/export.lua
```diff
@@ -24,9 +24,11 @@
 --(INCLUDING NEGLIGENCE OR OTHERWISE) ARISING IN ANY WAY OUT OF THE USE OF THIS
 --SOFTWARE, EVEN IF ADVISED OF THE POSSIBILITY OF SUCH DAMAGE.
 
+require('logger')
 function export_set(options)
     local item_list = T{}
-    local targinv,all_items,minify,all_sets,use_job_in_filename,use_subjob_in_filename,overwrite_existing,named_file
+    local targinv,all_items,wearable,all_sets,noaugments,onlyaugmented,minify,clipboard
+    local use_job_in_filename,use_subjob_in_filename,overwrite_existing,named_file
```
z16: I know this is not your doing, but I would put these all on separate lines and initialize them with default values (`false` for most of them).

addons/GearSwap/export.lua
```diff
@@ -24,9 +24,11 @@
...
     if #options > 0 then
         for _,v in ipairs(options) do
             if S{'inventory','inv','i'}:contains(v:lower()) then
```
z16: `v` is lowercased all the time. Define it once at the top and use that:
```lua
local option = v:lower()
```

lili-ffxi:
Still working on this, hopefully I can submit the new code soonTM.

lili-ffxi:
PR closed because of too many conflicts, reopened in #2469

PR #2347, opened 2024-04-13
(Send) Added string interpolation support

zmsl:
In certain cases it is necessary to send dynamic values through the Send addon. In order to support those values, I added String interpolation support that allows a Send user to input Lua into a Javascript template literal token format and have that evaluated at runtime. Javascript template literal token format was chosen because it is unlikely that players have existing alias and macros that use ${}, whereas formats like {}, (), or [] may be more common.

Example use:
`send @party /p Hi, my name is ${windower.ffxi.get_player().name}`
Results in all players in a party saying in party chat "Hi, my name is " and their character name


RubenatorX:
This doesn't seem necessary. 
Aside from the fact that your example could be accomplished with `<me>` instead, you can accomplish this functionality simply with a Lua script or via the `eval` addon.

PR #2349, opened 2024-04-20
support true master levels

posimagi:
A hacky implementation for allowing the correct master level to display while level synced


posimagi:
Intended to open this in my own fork while I came up with a better solution. Don't code while tired, kids.

PR #2350, opened 2024-04-21
[pointwatch] support the display of true ML under level sync

posimagi:
For real this time.


RubenatorX:
Let's not completely remove support for displaying your level synced master level.

posimagi:
Updated.

posimagi:
5 commits above were just to fix commit messages. 6th above actually changed code.

addons/pointwatch/statics.lua
```diff
@@ -31,7 +31,8 @@ default_settings = {
...
         abyssea = "string.format('%d/%dXP %sMerits XP/hr:%.1fk %sJP CP/hr:%.1fk %d/%dEP Amber:%d Azure:%d Ruby:%d Pearl:%d Ebon:%d Silver:%d Gold:%d Time-Remaining:%d',xp.current,xp.tnl,max_color('%5.2f':format(math.floor(lp.current/lp.tnm*100)/100+lp.number_of_merits),lp.current/lp.tnm+lp.number_of_merits,lp.maximum_merits,58,147,191),math.floor(xp.rate/100)/10,max_color('%6.2f':format(math.floor(cp.current/cp.tnjp*100)/100+cp.number_of_job_points),cp.current/cp.tnjp+cp.number_of_job_points,500,58,147,191),math.floor(cp.rate/100)/10,ep.current,ep.tnml,abyssea.amber or 0,abyssea.azure or 0,abyssea.ruby or 0,abyssea.pearlescent or 0,abyssea.ebon or 0,abyssea.silvery or 0,abyssea.golden or 0,abyssea.time_remaining or 0)",
-        },
+        ep = "string.format('ML%d %d/%dEP EP/hr:%.1fk',ep.true_master_level,ep.current,ep.tnml,math.floor(ep.rate/100)/10)",
```
RubenatorX: Why is this here?

addons/pointwatch/pointwatch.lua
```diff
@@ -121,6 +121,7 @@ packet_handlers = {
         ep.current = p['Current Exemplar Points']
         ep.tnml = p['Required Exemplar Points']
         ep.master_level = p['Master Level']
+        ep.true_master_level = tnml[p['Required Exemplar Points']]
```
RubenatorX: Let's swap these. Have the true level be `master_level`, and the other be `synced_master_level`.

RubenatorX (commented):
couple things

addons/pointwatch/statics.lua
posimagi: I didn't really want to edit the existing ones but seeing that they all include EP already I did anyway in the latest patchset.

addons/pointwatch/pointwatch.lua
posimagi: Done

addons/pointwatch/pointwatch.lua
```diff
@@ -120,7 +120,8 @@ packet_handlers = {
         accolades.current = p['Unity Points']
         ep.current = p['Current Exemplar Points']
         ep.tnml = p['Required Exemplar Points']
-        ep.master_level = p['Master Level']
+        ep.master_level = tnml[p['Required Exemplar Points']]
```
RubenatorX, line 123: Have you tested what happens when you've not unlocked Master Levels? Isn't this inner value `0` in that case? and if so, I'm pretty sure this would evaluate to `nil`, which I believe would maybe/probably cause string format error(s).

posimagi: I can try it on my mule, though I have no way to test the case where a character is 99 but doesn't have master levels unlocked, if that matters.

RubenatorX: afaik it should have the same behavior

posimagi: Confirmed that on a character without ML unlocked it shows "ML0" as expected.

RubenatorX: I don't see how that's possible since there is no `0` key in your table... so... how is that possible? 🤔

RubenatorX: Oh, or are you saying that even without master breaker it says that Required Exemplar points is 2500?

RubenatorX: Oh, jk, it does have 2500 in it, so no problem at all (although, in a perfect world, would be nice if it would just, not display ML/EP at all if it's not applicable but, that's too much work ;p

posimagi: I'll add that functionality at some unspecified point in the future if I don't entirely rewrite the addon first lol

PR #2351, opened 2024-04-21
[mountmuzzle] support custom BGM per mount

posimagi:
Allow users to set MountMuzzle BGM per mount per https://www.ffxiah.com/forum/topic/37379/lua-addon-ideas/29/#3700676.

New commands:
- `//mm set custom` (alias: `//mm s u`, for "user", to avoid conflict with "chocobo")
- `//mm c[ustom] MountName TrackNumber`

Features:
- Shows a success message when the user sets a track with the mount's name and track's title.
- Throws an error if the user tries to set a track for a mount that doesn't exist.

image


posimagi:
Bump.

addons/MountMuzzle/mountmuzzle.lua
```diff
@@ -258,4 +286,14 @@ windower.register_event('incoming chunk', function(id, data)
         
         tryInject()
     end
+end)
+
+windower.register_event('outgoing chunk',function(id, data)
```
z16: Space after the comma.

addons/MountMuzzle/constants.lua
```diff
@@ -99,13 +100,47 @@ muzzles = {
...
         description = 'Current Zone Music'
+    },
+    custom = {
+        name = 'custom',
+        song = music.songs.custom,
+        description = 'User-Defined Per-Mount Music'
     }
 }
 
+songs = {
```
z16, line 111: Is this the same list that `setbgm` uses? If so, it would be nice to combine them somewhere. Honestly, this looks like something for resources (probably added in the `fixes.xml`). But shouldn't stop this PR, just something to do in the future, so that we don't have to maintain both lists.

addons/MountMuzzle/mountmuzzle.lua
```diff
@@ -87,11 +91,17 @@ local help = {
...
+            z = muzzles.zone.name,
+            u = muzzles.custom.name
         }
     }
 }
 
+local possible_mounts = L{}
+for _, mount in pairs(_resources.mounts) do
+    possible_mounts:append(mount.name:lower())
+end
```
z16: Can condense these lines to this:
```lua
local possible_mounts = _resources.mounts:map(function(mount) return mount.name:lower() end)
```

addons/MountMuzzle/constants.lua
```diff
@@ -62,6 +62,7 @@ music = {
         mount = 84,
         chocobo = 212,
         zone = 0,
+        custom = 9998
```
z16: I assume these are song IDs and you uses `9998` just to make sure it doesn't exist, and because `9999` was already taken? If so, would it be an issue if they both uses `9999`? In general, I'd prefer clearly separate numbers, like negative numbers, or ideally just leaving them `nil`, if that works with the rest of the code.

z16 (changes requested):
Left some notes, the only thing that needs to change is the missing space, the rest is optional and I'll leave to you.

addons/MountMuzzle/constants.lua
RubenatorX: @z16 
There may be an argument for them to be one addon instead of two since, they both simply change game music. :shrug:

lili-ffxi: I haven't read the PR enough to remember if it actually does it, but there were talks of making this addon also able to select a *random* BGM to go along with the random mount. Setbgm wouldn't be able to do that without some refactoring. Besides, associating a specific different song to each mount also feels out of scope for setbgm, but ymmv.

Anyway, +1 to adding custom background music resources. I can even take care of that!

posimagi: This is indeed the same list as setbgm. I'm not at all opposed to combining them and/or moving the track list into resources, etc.

addons/MountMuzzle/constants.lua
posimagi: I don't think there'd be any issue with making it also 9999. I will test it out.

posimagi:
Updated to address comments.

PR #2363, opened 2024-06-23
Update Omen & new addon, AutoFPS

avaera8775:
This is my first PR, please don't hesitate to let me know if there are any mistakes.

**Made some updates to the Omen addon:**
- Added commands for visibility, save settings (position), help text
- Moved messages into their own separate file
- Added more settings for settings.xml (control font family, bg visible, easier to read structure).

**New Addon**
AutoFPS:
- Adjusts the FPS automatically during cutscene events
- When exiting a cutscene, it will resume back to 60 FPS


lili-ffxi:
These should be two separate PRs.

Furthermore, how does AutoFPS differ from FastCS?

avaera8775:
> Furthermore, how does AutoFPS differ from FastCS?

From my understanding, FastCS always runs at uncapped or 60 fps in cutscenes to speed it up, so when you run into a cutscene that is known to bug/freeze the game you have to manually set it to 30 or restart the game. 

Admittedly, my addon will always run it at 30, so they don't serve the same purpose, from my testing it does not cause any freezing behavior.

RubenatorX:
Edit: oh my bad I missed lili's comment

Not sure why you closed this but, this should be two different PRs

PR #2369, opened 2024-08-20
Update labels on Item flags 02 and 03

lili-ffxi:
As per https://github.com/Windower/POLUtils/issues/25

I am not sure about naming (tho these match the labels used on BG-Wiki, and I think LSB too), opinions appreciated.


RubenatorX:
May as well update Flag00 while you're at it
https://github.com/LandSandBoat/server/blob/8f0baff67c947915535abcb8f34b42bc5f06ce68/scripts/enum/item_flag.lua#L8-L12

lili-ffxi:
> May as well update Flag00 while you're at it https://github.com/LandSandBoat/server/blob/8f0baff67c947915535abcb8f34b42bc5f06ce68/scripts/enum/item_flag.lua#L8-L12

...how the duck do I name this? "Wallhanging"? "Can be hung on wall in Mog House"? "BEHOLD, A PAINTING"?

lili-ffxi:
Decided to go with Wall-mounted because FFXIV uses that name and it felt appropriate. Can change it to "Wall Furnishing" if we prefer?

PR #2383, opened 2024-10-23
`switch_focus` addon 

Nifim:
This lets you swap focus between your various logged in characters.


addons/switch_focus/box_list.lua
```diff
@@ -0,0 +1,51 @@
...
+            return self._byValue[value]
+        end
+    end
+}
+local meta = {
+    __index = function(t,k)
+        if t._orderd[k] then
+            return t._orderd[k]
+        end
+    end,
```
z16, line 44: This can be just `return t._ordered[k]`, no need for the `if`.

addons/switch_focus/box_list.lua
```diff
@@ -0,0 +1,51 @@
...
+        self._byValue = {}
+        for k, v in ipairs(self._orderd) do
+            self._byValue[v] = k
+        end
+    end,
+
+    get_index_of = function(self, value)
+        if self._byValue[value] then
+            return self._byValue[value]
+        end
```
z16: This can just be `return self._byValue[value]`, no need for the `if`.

addons/switch_focus/switch_focus.lua
```diff
@@ -0,0 +1,150 @@
...
+
+require('tables')
+require('strings')
+local box_list = require('box_list')
+
+local back_name
+local switch = {cmd = {}}
+
+function switch.command(cmd, ...)
+    local args = T{...}:map(string.lower) or {}
```
z16: The `or {}` has no effect, as the previous table can never be falsy.

addons/switch_focus/box_list.lua
```diff
@@ -0,0 +1,51 @@
...
+            return
+        end
+
+        table.insert(self._orderd, value)
+        table.sort(self._orderd)
+
+        self._byValue = {}
+        for k, v in ipairs(self._orderd) do
+            self._byValue[v] = k
+        end
```
z16, line 20: Why does the table need to be cleared and repopulated? Couldn't you just assign the new value? Same in `remove`, there you coul just remove the value.

addons/switch_focus/switch_focus.lua
```diff
@@ -0,0 +1,150 @@
...
+    cmd = cmd and cmd:lower() or 'help'
+    if switch.cmd[cmd] then
```
z16: Since you're already assigning a new variable before, may as well do:
```lua
    local command = switch.cmd[cmd and cmd:lower() or 'help']
```

Then you can just check that and save a few table lookups, and you wouldn't be overwriting a local variable. If you're feeling cheeky, you could also add `or functions.empty` after it, then you don't need the `if` anymore, and can just call it directly.

addons/switch_focus/box_list.lua
```diff
@@ -0,0 +1,51 @@
+local box_list = {
+    _orderd = {},
+    _byValue = {},
```
z16: You use `snake_case` for everything else, but `camelCase` here. Would prefer to `snake_case` this as well.

addons/switch_focus/switch_focus.lua
```diff
@@ -0,0 +1,150 @@
...
+
+        next = box_list[(index % box_list:len()) + 1]
+    end
+end
+switch.cmd.n = switch.cmd.next
+
+switch.ipc = {}
+local function ipc_message(raw_msg)
+    raw_msg = raw_msg and raw_msg:lower()
+    print(raw_msg)
```
z16: Did you mean to remove that?

addons/switch_focus/switch_focus.lua
```diff
@@ -0,0 +1,150 @@
...
+    local msg = string.split(''..raw_msg, ',')
```
z16: Why is the `''..` there? Shouldn't `raw_msg` always be a string? And even if it wasn't, I'd much prefer `tostring(raw_msg)` over this.

addons/switch_focus/switch_focus.lua
```diff
@@ -0,0 +1,150 @@
...
+        switch.ipc[msg[1]](msg)
+    end
+end
+windower.register_event('ipc message', ipc_message)
+
+function switch.ipc.to(msg)
+    local from_name = msg[2]
+    local to_name = msg[3]
+    local player =  windower.ffxi.get_player()
+    if windower.has_focus ~= true and ((player and to_name == player.name:lower()) or (to_name == '@lobby' and not windower.ffxi.get_info().logged_in)) then
```
z16: Shouldn't this be `windower.has_focus()`? And I think a `not` in front would be a bit clearer than `~= true`, but up to you.

addons/switch_focus/box_list.lua
Nifim: the order could change, meaning the value for each name to index could change.
for example: i have Nifim logged in then Vanim joins so `Nifim = 1, Vanim = 2` then i log into Jotim when we add him the indexes become `Jotim = 1, Nifim = 2, Vanim = 3` he causes a shift in the whole array so clearing out the table for each addition or removal is necessary.

z16 (approved):

PR #2394, opened 2024-11-07
Libs/Strings: Fix Shift-JIS character code range

LycorisXI:
There is an issue with the Shift-JIS iterator definition in libs/strings.lua where it cannot iterate correctly when characters with a second byte in the range of 0xF0–0xF9 are included in the string.

The ranges for the first and second bytes in Shift-JIS are as follows:

First byte: 0x81–0x9F, 0xE0–0xEF
Second byte: 0x40–0x7E, 0x80–0xFC
Strictly speaking, there are some unmapped characters within these ranges, but for the purposes of determining how many bytes to advance for iteration (rather than for character mapping), these ranges should be sufficient.

I have modified the condition to perform checks within these defined ranges.

You can verify this by running a Lua script like the following:
```
require 'strings'
str = "abc" .. string.char(0x081,0x0F4) .. "def" -- "abc♪def"
windower.send_command('input /echo ' .. str)
pattern = string.char(0x081,0x0F4)
i, j = str:find(pattern, string.encoding.shift_jis, 1)
print(i, j)
i, j = str:find(pattern, string.encoding.shift_jis, 5) -- error
print(i, j)
```

### DIFF
--- a/addons/libs/strings.lua
+++ b/addons/libs/strings.lua
@@ -109,9 +109,10 @@ do
             [string.encoding.shift_jis] = function(str, from, to)
                 return process(str, from, to, function(byte)
                     return
-                        (byte < 0x80 or byte >= 0xA1 and byte <= 0xDF) and 1 or
-                        (byte >= 0x1E and byte <= 0x1F or byte >= 0x80 and byte <= 0x9F or byte >= 0xE0 and byte <= 0xEF or byte >= 0xFA and byte <= 0xFC) and 2 or
-                        byte == 0xFD and 6
+                        byte == 0xFD and 6 or
+                        ((byte >= 0x81 and byte <= 0x9F) or (byte >= 0xE0 and byte <= 0xEF)) and 2 or
+                        (byte < 0x80 or (byte >= 0x40 and byte <= 0x7E) or (byte >= 0x80 and byte <= 0xFC)) and 1 or
+                        nil
                 end)
             end,
             [string.encoding.binary] = function(str, from, to)


LycorisXI:
The following code is somewhat redundant, with overlapping ranges, but I wrote it this way to explicitly represent the Shift-JIS range.
```
(byte < 0x80 or (byte >= 0x40 and byte <= 0x7E) or (byte >= 0x80 and byte <= 0xFC)) and 1 or
```

Nifim:
Did you test this with autotranslate and the ffxi color code?

RubenatorX:
FFXI's uses **modified** Shift-JIS -- do you have a practical example of this character range getting used for something in game?
Or are you using the algorithm on shift-jis from outside of the game? Or trying to make it comply with the unmodified standard?
If you're using it on text from outside the game, then we may need to have an internal discussion about that use case because, that isn't what we designed it for but, we can consider it. (And it's less of a big deal for counting characters, and possibly bigger deal for other related things)

LycorisXI:
I have confirmed about autotranslate, although the verification is on my private server. Before the modification, sending the auto-translated string with send addon resulted in an error, but after this range modification, it can be executed without any problem.

Before:
image
After:
image

I do not know how to test the FFXI color code. Is there any good way to verify it?

RubenatorX:
@lili-ffxi @z16 -- While the iterator adjustment seems to be potentially justified (probably not this exact change since it is treating 2nd byte same as the 1st):
1) They appear to have found a way to get it to return a boolean instead of a number, which seems bad (or possibly just symptomatic of passing in things that it wasn't designed to handle, see 2)
2) I haven't had time to check what `send` is calling or passing in in this situation but, seems like auto-translates are possibly problematic in whatever is going on
Edit: correction, seems like maybe the Japanese characters resulting from the auto-translate being converted before sending it over ipc are not covered by the character counting -- see message below. Related question for thought: Should `send` even be converting auto-translate before sending? 🤔

LycorisXI:
> do you have a practical example of this character range getting used for something in game?

The character code range I mentioned above is the general range for Shift-JIS and is not specific to FFXI, but my understanding is that characters within this range are used in FFXI as well.

Many characters exist where the second byte falls within the range 0xF0–0xF9. For example, in "栄光の凱旋マーチ" (Victory March), the character "光" is represented by 0x8CF5, and "旋" by 0x90F9.

When executing //send selfname /song "栄光の凱旋マーチ" <me>, string.find() is called within the send addon to parse the command, and it seems that an iteration error occurs in this find() call.

Nifim:
The check on this line should be falsely rather then explicitly `nil`, seeing as how the function cant return nil
https://github.com/Windower/Lua/blob/a413648abd74bc3b91775059616c2fde0e4620e6/addons/libs/strings.lua#L87C1-L89C20

lili-ffxi:
> 2\. I haven't had time to check what `send` is calling or passing in in this situation but, seems like auto-translates are possibly problematic in whatever is going on
>     Edit: correction, seems like maybe the Japanese characters resulting from the auto-translate being converted before sending it over ipc are not covered by the character counting -- see message below. Related question for thought: Should `send` even be converting auto-translate before sending? 🤔

Send's internals should never be able to see autotranslate strings, as they're converted to regular text immediately before anything else is done to them:
https://github.com/Windower/Lua/blob/a413648abd74bc3b91775059616c2fde0e4620e6/addons/send/send.lua#L30

The feature of being able to pass unconverted AT terms is planned (and half written on my hard disk), but we had decided to postpone it to a later date at the time of the send rehaul, mostly due to a lack of time on everyone's part. So it's Temporary<sup> TM</sup> and the final goal is to duplicate the exact string as sent by the user, we're just not there yet.

Nifim:
> I do not know how to test the FFXI color code. Is there any good way to verify it?

@LycorisXI  I dont know that you can do it from the chat, but `0x1F` followed by a second byte between 0 and 255, and you pair that with a `0x1E` followed by a byte = n - 254, where n is the value from your `0x1F` second byte

z16:
The second byte should not be relevant to our algorithm - it can be in any range, but if the lead byte indicates a multi-byte sequence, it will be skipped. If not, this is a bug in our mechanism.

Can you give a concrete example of a string that is not properly iterated with our current mechanism? I tried with `栄光の凱旋マーチ` and it worked fine, see this snippet:
```lua
require('strings')

local utf8 = '栄光の凱旋マーチ'
local sjis = windower.to_shift_jis(utf8)

local utf8chars = utf8:it(string.encoding.utf8):pack()
local sjischars = sjis:it(string.encoding.shift_jis):pack()

print(utf8:hex(' '))
print(sjis:hex(' '))

for i = 1, 10 do
	print(utf8chars[i]:hex(' ') .. '        ' .. sjischars[i]:hex(' '))
end
```

This printed the following:
```
E6 A0 84 E5 85 89 E3 81 AE E5 87 B1 E6 97 8B E3 83 9E E3 83 BC E3 83 81
89 68 8C F5 82 CC 8A 4D 90 F9 83 7D 81 5B 83 60
E6 A0 84        89 68
E5 85 89        8C F5
E3 81 AE        82 CC
E5 87 B1        8A 4D
E6 97 8B        90 F9
E3 83 9E        83 7D
E3 83 BC        81 5B
E3 83 81        83 60
```

I also verified that our `windower.to_shift_jis` function works correctly, all byte sequences are correct. I also tried with your `abc♪def` string from the original post, and got the expected result (and again, verififed the byte sequence for ♪ to be correct):
```
61 62 63 E2 99 AA 64 65 66
61 62 63 81 F4 64 65 66
61        61
62        62
63        63
E2 99 AA        81 F4
64        64
65        65
66        66
```

Whatever issue you're experiencing, I would expect it to be unrelated to the decoder. But again, if you can find an example that doesn't iterate correctly, let me know.

That said, even outside of a possible issue, your suggested changes ignore a number of game-internal code sequences, such as color codes and resets. I already mentioned in our internal discussion that `shift_jis` is probably an inaccurate name for this encoding, maybe we should call it `ffxi` or `game` to indicate we use *whatever the game uses*, which just happens to be Shift JIS + a number of extensions. I'm open to suggestions for this, the main reason why I originally went with `shift_jis` is because our conversion functions are also called `windower.from_shift_jis` and `windower.to_shift_jis`, and keeping parity with those is a nice bonus, but if it confuses people, I'd be open to changing it.

LycorisXI:
@Nifim Thanks.
So 0x1F, 0x1E should be treated as a multibyte character along with any following color code. The code I presented did not take that into account.

LycorisXI:
The issue I presented may have deviated from the essence of the problem. Indeed, there is no issue when iterating through multibyte characters sequentially from the beginning. The problem arises when iteration starts from the middle of a multibyte character (the second byte), and that starting byte is in the range of 0xF0-0xF9.

I believe that starting iteration from the middle of a multibyte character is inherently incorrect, but this actually occurs in cases such as with string.find().

The following script will result in an error `Lua runtime error: libs/strings.lua:91: attempt to perform arithmetic on local 'length' (a boolean value)`:
```
require 'strings'
local utf8 = '栄光の凱旋マーチ <me>'
local sjis = windower.to_shift_jis(utf8)
i, j = sjis:find('<me>', string.encoding.shift_jis, 1)
print(i, j)
```

This is just a guess, but it might be that findpattern performs pattern matching by advancing one byte at a time, regardless of whether it is targeting multibyte characters.
https://github.com/Windower/Lua/blob/dev/addons/libs/strings.lua#L547-L552

z16:
You're right, this looks like an issue in the `findpattern` function. I will look into it. I will also adjust the iterator to return `nil` so the check works properly, so we get a better error message in the future. I will close this PR, unless you have objections.

z16:
Pushed some changes to the dev branch. There is another issue that I would like to address, but it will require significant effort, so I will do it another time. Feel free to test it and report back if it works for you.

mazdarxseven:
Just wanted to add my experiences here since the issue looks like it might not be resolved yet.
Doing some simple tests with //send to my alt for some SCH spells,
Works:
`send Alt /ja "女神降臨の章" <me>`
(Accession)

strings.lua line 91 error, as described earlier in the thread:
`send Alt /ma "鼓舞激励の策" <me>`
(Adloquium)

The version published about 19 hours ago fixed adloquium, but when I tried:
`send Alt /ja "机上演習" <me>`
(Sublimation)

I got an error in strings.lua line 88: Invalid code point

I hope this helps.

z16:
> Just wanted to add my experiences here since the issue looks like it might not be resolved yet. Doing some simple tests with //send to my alt for some SCH spells, Works: `send Alt /ja "女神降臨の章" <me>` (Accession)
> 
> strings.lua line 91 error, as described earlier in the thread: `send Alt /ma "鼓舞激励の策" <me>` (Adloquium)
> 
> The version published about 19 hours ago fixed adloquium, but when I tried: `send Alt /ja "机上演習" <me>` (Sublimation)
> 
> I got an error in strings.lua line 88: Invalid code point
> 
> I hope this helps.

Can you open an issue with this? Easier for us to track and I see no reason to reopen this PR.

mazdarxseven:
> > Just wanted to add my experiences here since the issue looks like it might not be resolved yet. Doing some simple tests with //send to my alt for some SCH spells, Works: `send Alt /ja "女神降臨の章" <me>` (Accession)
> > strings.lua line 91 error, as described earlier in the thread: `send Alt /ma "鼓舞激励の策" <me>` (Adloquium)
> > The version published about 19 hours ago fixed adloquium, but when I tried: `send Alt /ja "机上演習" <me>` (Sublimation)
> > I got an error in strings.lua line 88: Invalid code point
> > I hope this helps.
> 
> Can you open an issue with this? Easier for us to track and I see no reason to reopen this PR.

That makes sense, my bad.
I've opened an issue here: https://github.com/Windower/Lua/issues/2396

PR #2401, opened 2024-11-12
Update to findAll.lua adding the AF cards tracking feature

Nalfey:
Added feature that tracks the AF cards you have and need for upgrades. It looks for the AFs NQ, +1, +2, +3 that you have for as specific job and tells you how many cards you have and need to augment them.

Added a new command:
//find cards {JOB}

Examples:
//find cards GEO
//find cards WHM

Capture
Capture2

### DIFF
--- a/addons/findAll/findAll.lua
+++ b/addons/findAll/findAll.lua
@@ -27,8 +27,8 @@ SOFTWARE, EVEN IF ADVISED OF THE POSSIBILITY OF SUCH DAMAGE.
 ]]
 
 _addon.name    = 'findAll'
-_addon.author  = 'Zohno'
-_addon.version = '1.20190514'
+_addon.author  = 'Zohno & Nalfey'
+_addon.version = '1.2020000'
 _addon.commands = {'findall'}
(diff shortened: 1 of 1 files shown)


lili-ffxi:
This looks to me like it's severely out of scope for findall, and would be much better as a standalone addon (even if it would duplicate some code from findall).

Nalfey:
I don't think creating a new addon just for this feature is a good idea, not only would it duplicate code, people would also need to duplicate their character item data files which is less than ideal.

But if you're really set on making a standalone addon I'm happy to branch it off

lili-ffxi:
You don't need to duplicate item data files, you can let findall keep managing that and process the data files from another addon. For example, dupefind does exactly that: https://github.com/lili-ffxi/FFXI-Addons/blob/5cf032500916f60215ec51b7c8386d6ffa54d7db/dupefind/dupefind.lua#L182

Nalfey:
ah thanks that's a good idea, I'll make a new addon called Cards then

PR #2405, opened 2024-11-21
Chatmon v1.1.0

Nifim:
Fix some incorrect implementations of ported plugin features.
Add `senderName` and `notSenderName` fields for chat mode types.
Convert autotranslate text for evaluation.
Strip color codes from text for evaluation.

### DIFF
--- a/addons/chatmon/chatmon.lua
+++ b/addons/chatmon/chatmon.lua
@@ -72,6 +72,7 @@ local function check_triggers(from, text)
         return
     end
 
+    text = windower.convert_auto_trans(text)
     for _, trigger in ipairs(triggers) do
         local match = get_match(trigger.match)
         if trigger.from:contains(from) and not trigger.notFrom:contains(from) and windower.wc_match(text, match) and not windower.wc_match(text, trigger.notMatch) then
--- a/addons/chatmon/chatmon.lua
+++ b/addons/chatmon/chatmon.lua
@@ -1,3 +1,31 @@
+--[[
+Copyright © 2024, Windower
+All rights reserved.
+
+Redistribution and use in source and binary forms, with or without
+modification, are permitted provided that the following conditions are met:
+
+    * Redistributions of source code must retain the above copyright
+      notice, this list of conditions and the following disclaimer.
+    * Redistributions in binary form must reproduce the above copyright
+      notice, this list of conditions and the following disclaimer in the
+      documentation and/or other materials provided with the distribution.
+    * Neither the name of chatmon nor the
+      names of its contributors may be used to endorse or promote products
+      derived from this software without specific prior written permission.
+
+THIS SOFTWARE IS PROVIDED BY THE COPYRIGHT HOLDERS AND CONTRIBUTORS "AS IS" AND
+ANY EXPRESS OR IMPLIED WARRANTIES, INCLUDING, BUT NOT LIMITED TO, THE IMPLIED
+WARRANTIES OF MERCHANTABILITY AND FITNESS FOR A PARTICULAR PURPOSE ARE
+DISCLAIMED. IN NO EVENT SHALL Windower BE LIABLE FOR ANY
+DIRECT, INDIRECT, INCIDENTAL, SPECIAL, EXEMPLARY, OR CONSEQUENTIAL DAMAGES
+(INCLUDING, BUT NOT LIMITED TO, PROCUREMENT OF SUBSTITUTE GOODS OR SERVICES;
+LOSS OF USE, DATA, OR PROFITS; OR BUSINESS INTERRUPTION) HOWEVER CAUSED AND
+ON ANY THEORY OF LIABILITY, WHETHER IN CONTRACT, STRICT LIABILITY, OR TORT
+(INCLUDING NEGLIGENCE OR OTHERWISE) ARISING IN ANY WAY OUT OF THE USE OF THIS
+SOFTWARE, EVEN IF ADVISED OF THE POSSIBILITY OF SUCH DAMAGE.
+]]
+
 _addon = {}
 _addon.name = 'chatmon'
 _addon.version = '1.0.5'
--- a/addons/chatmon/chatmon.lua
+++ b/addons/chatmon/chatmon.lua
@@ -28,7 +28,7 @@ SOFTWARE, EVEN IF ADVISED OF THE POSSIBILITY OF SUCH DAMAGE.
 
 _addon = {}
 _addon.name = 'chatmon'
-_addon.version = '1.0.5'
+_addon.version = '1.1,0'
 _addon.author = 'WindowerDevTeam'
 _addon.commands = {'chatmon'}
 
--- a/addons/chatmon/chatmon.lua
+++ b/addons/chatmon/chatmon.lua
@@ -47,42 +47,22 @@ local function play_sound(sound)
     end
 end
 
-local function get_match(match)
-    if match ~= '<name>' then
-        return match
-    end
-
-    -- this is done to have parity with the old plugin.
-    local player_name = windower.ffxi.get_player().name:lower()
-    return '* ' .. player_name .. '|'
-                .. player_name .. ' *|*\''
-                .. player_name .. '\'*|*('
-                .. player_name .. ')*|'
-                .. player_name .. '|* '
-                .. player_name .. ' *|* '
-                .. player_name .. '? *|* '
-                .. player_name .. '?|'
-                .. player_name .. '? *|'
-                .. player_name .. '?|*<'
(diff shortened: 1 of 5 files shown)


addons/chatmon/chatmon.lua
```diff
@@ -1,6 +1,6 @@
 _addon = {}
 _addon.name = 'chatmon'
-_addon.version = '1.0.4'
+_addon.version = '1.0.5'
```
z16: This deserves, at the *very* least, an increase to `1.1.0`.

addons/chatmon/chatmon.lua
```diff
@@ -47,42 +47,36 @@ local function play_sound(sound)
...
-end
-
-local function check_triggers(from, text)
+local function check_triggers(from, text, sender)
     if windower.has_focus() and settings.DisableOnFocus then
         return
     end
 
+    text = windower.convert_auto_trans(text)
+    text = text:strip_color()
```
z16, line 84: `:strip_format` is more appropriate, but depending on the text source it may not make a difference, but still the more correct generic solution.

addons/chatmon/trigger_class.lua
```diff
@@ -0,0 +1,67 @@
...
+end
+
+function field_checkers.notSender(trigger, event)
+    return windower.wc_match(event.sender, trigger.notSender)
+end
+
+function trigger_class:check(event)
+    for k in pairs(self) do
+        if field_checkers[k] and not field_checkers[k](self, event) then
+            print(k)
```
z16: I assume this was not meant to be here anymore?

addons/chatmon/trigger_class.lua
```diff
@@ -0,0 +1,67 @@
...
+local meta = {__index = trigger_class}
+
+local function get_match(match)
+    if match ~= '<name>' then
+        return match
+    end
+
+    -- this is done to have parity with the old plugin.
+    local player_name = windower.ffxi.get_player().name:lower()
+    return '* ' .. player_name .. '|'
```
z16, line 39: I still say we replace this with a simple regex match with `'\\b' .. player_name .. '\\b'`.

addons/chatmon/readme.md
```diff
@@ -21,28 +21,62 @@
 #### Tigger Fields:
 * `from`: a set of chat name you would like the trigger to happen on, the full list of chat modes can be found here [resources/chat.lua](https://github.com/Windower/Resources/blob/master/resources_data/chat.lua)
```
z16: I would adjust this to not mention "chat modes", since it may give people the impression that they need modes, not just names.

addons/chatmon/readme.md
```diff
@@ -21,28 +21,62 @@
...
 
   * `*` can be used as a wild card. Ex. `*Nif*` will trigger for any text containing with `Nif` such as `Hi Nifim` or `Hi Niflheim` or `lolNifim`
 
-  * `|` cab be used to seperate multiple words you want to tigger for such as `Jo|Yo`, will trigger `Jo` and `Yo`
+  * `|` can be used to seperate multiple words you want to tigger for such as `Jo|Yo`, will trigger `Jo` and `Yo`
 
   * `<name>` can be used as a special flag to cover your character name in a number of common contexts.
 
-  * when the trigger source is `tell`, `invite`, or `examine` the text evaluated for match is the **senders name** all other sources evaluate the message text.
+  * when the trigger source is `tell`, `emotes`, `invite`, or `examine` the text evaluated for match is the **senders name** all other sources evaluate the message text.
```
z16: Er, why is `tell` in that list? Should `tell` match the text, not the name? Applies to the below comments as well. Or is that plugin behavior we're trying to match?

Also, did we settle on `emotes` instead of `emote`? I probably prefer the latter 🤔 But open to suggestions.

addons/chatmon/readme.md
```diff
@@ -21,28 +21,62 @@
...
 
-#### Example Trigger:
+### Example Trigger:
 
+#### This example triggers when any player other than Arcon says something containing Nif but not if they Nifl:
 ```lua
-{ from = S{ "tell" }, notFrom = S{}, match = "*", notMatch = "", sound = "IncomingTell.wav"},
+{ from = S{ "say" }, notFrom = S{ "shout" }, match = "*Nif*", notMatch = "*Nifl*", notSender = "Arcon" sound = "IncomingTalk.wav"},
 ```
+triggers exampe:
```
z16: `example` misspelled.

addons/chatmon/readme.md
```diff
@@ -21,28 +21,62 @@
...
+> Iyroku : Hi Nif
```
z16: Not that it's important, but `Iryoku` misspelled :D

addons/chatmon/readme.md
```diff
@@ -21,28 +21,62 @@
...
+
+does not triggers example:
+> Arcon : Hi Nif
+
+> Iyroku : Hi Nifl
```
z16: Here as well.

addons/chatmon/readme.md
```diff
@@ -21,28 +21,62 @@
...
 
+#### This example triggers when any player other than Arcon says something containing Nif but not if they Nifl:
 ```lua
-{ from = S{ "tell" }, notFrom = S{}, match = "*", notMatch = "", sound = "IncomingTell.wav"},
+{ from = S{ "say" }, notFrom = S{ "shout" }, match = "*Nif*", notMatch = "*Nifl*", notSender = "Arcon" sound = "IncomingTalk.wav"},
 ```
+triggers exampe:
+> Iyroku : Hi Nif
+
+does not triggers example:
```
z16: Both here and below you say `does not triggers` instead of `does not trigger`. I'd prefer a more proper prose throughout the readme, including proper sentence capitalization, but I'll leave that up to you.

addons/chatmon/trigger_class.lua
```diff
@@ -0,0 +1,67 @@
...
+            return false
+        end
+    end
+
+    return true
+end
+
+function trigger_class:new(o)
+    setmetatable(o, meta)
+    return o
```
z16: Can be shortened to `return setmetatable(o, meta)`.

addons/chatmon/readme.md
Nifim: well, tell does match the name of the sender... but I apparently misinterpreted the plugin code which does indeed go off of the text of the message 🤔 I kinda want it to do both.

Also, you are correct it should be `emote`, which is what the code uses.

addons/chatmon/chatmon.lua
Nifim: strip_format, strips the color and the auto translate, but the way it strips auto translate does not convert the auto translate into text that could be matched against.

So I call `convert_auto_trans` on the string converting all the auto translate into their plain text representation. the text at the time i call `:strip_color` can't contain auto translate so `:strip_format` seemed weird to use there.

addons/chatmon/trigger_class.lua
Nifim: I honestly am not sure it is necessary to begin with, i think it could just be `*player_name*`

addons/chatmon/chatmon.lua
z16: It can - if the text is evaluated with the auto-translate code resolved, `convert_auto_trans` does nothing anymore, but it will contain the auto-translate braces. To cover every use case, you need both `convert_auto_trans` and `strip_format`.

addons/chatmon/chatmon.lua
```diff
@@ -1,6 +1,34 @@
...
+LOSS OF USE, DATA, OR PROFITS; OR BUSINESS INTERRUPTION) HOWEVER CAUSED AND
+ON ANY THEORY OF LIABILITY, WHETHER IN CONTRACT, STRICT LIABILITY, OR TORT
+(INCLUDING NEGLIGENCE OR OTHERWISE) ARISING IN ANY WAY OUT OF THE USE OF THIS
+SOFTWARE, EVEN IF ADVISED OF THE POSSIBILITY OF SUCH DAMAGE.
+]]
+
 _addon = {}
 _addon.name = 'chatmon'
-_addon.version = '1.0.4'
+_addon.version = '1.1,0'
```
z16: A comma snuck in! :D

addons/chatmon/trigger_class.lua
z16: The only problem with just using `player_name` directly, is that it would match it if it was part of a string. So if someone was named `Bob` it would trigger on `I bobbed for apples yesterday`. Hence regex with `\\b` around the player name for boundary matching seems most reasonable.

PR #2407, opened 2024-12-15
New addon: Armory. Send storage contents to an html datatable for easy viewing and filtering.

BluesummersC:
Armory reads the exported lua file from the FindAll addon, modifies it slightly and sends it to a endpoint, driven by Flask, that will cross-reference the item-ids to build a fully descriptive account of all the items in that characters storage and serve it up on a webserver into a sortable, searchable and filterable datatable.  
  
In order for the frontend to reference the correct character, 2 fields are created: Name (taken from windower.get_player().name) and a randomized 8 digit Passkey created on database insert, which is then added to the reply back to the addon. The addon adds these as url arguments and open a page to the site which automatically requests the info and populates the datatable. The Passkey is kept in data/settings.xml for reuse in upserts.

### DIFF
--- a/addons/Armory/armory.lua
+++ b/addons/Armory/armory.lua
@@ -1,0 +1,209 @@
+--[[
+Copyright © 2024, Caleb Bluesummers
+All rights reserved.
+
+Redistribution and use in source and binary forms, with or without
+modification, are permitted provided that the following conditions are met:
+
+    * Redistributions of source code must retain the above copyright
+      notice, this list of conditions and the following disclaimer.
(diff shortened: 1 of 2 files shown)


z16:
I don't think this is a good fit for the launcher. It seems too specific to your website, and relies on sending data to an external URL. There's no way for us to make security guarantees of this addon that we're normally trying to make. You're obviously free to distribute it on your own, but we're not in a position to take responsibility for this.

Sorry this comes a bit late, I've been putting the decision off, trying to think of a way to integrate this in a way we could distribute it, but I see no good solution for this within our ecosystem.

PR #2410, opened 2024-12-30
Fix url

lili-ffxi:
Update an url that has never been updated


z16:
That's not quite the URL it was meant to link to, but I honestly think the entire section should be replaced with a link to the wiki and a minor blurb.

PR #2412, opened 2024-12-31
[Trusts] fix: Fixes nil index error

0xv1n:
This change adds logic to error out if no set name is provided rather than spitting a nil-index error that might be incomprehensible to most users.

### DIFF
--- a/addons/Trusts/Trusts.lua
+++ b/addons/Trusts/Trusts.lua
@@ -27,7 +27,7 @@ SOFTWARE, EVEN IF ADVISED OF THE POSSIBILITY OF SUCH DAMAGE.
 ]]
 _addon.name='Trusts'
 _addon.author='from20020516'
-_addon.version='1.1'
+_addon.version='1.1.1'
 _addon.commands={'trusts','tru'}
 
 config = require('config')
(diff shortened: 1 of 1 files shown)


RubenatorX:
@0xv1n
If you change the version to 1.2 and use `error()` instead of `log()` I'll accept it.

0xv1n:
> @0xv1n If you change the version to 1.2 and use `error()` instead of `log()` I'll accept it.

@RubenatorX - done.

z16:
I'd like to move the checking to the command handler. I think it should also change from `error`, which reports to console, to `windower.add_to_chat`, to make it congruous with the `help` command. In fact, I think invoking the `help` command would not be the worst thing to do in this case, but I'll leave that decision up to you.

0xv1n:
@z16  - just moved all the other error() to add_to_chat in the process and stripped out the help function. I'd rather the error just tell you directly how to run the command correctly rather than spit the entire list of commands personally. But if it's an issue, it's now possible to call `show_help()`. Not sure why that wasn't its own function anyways.

addons/Trusts/Trusts.lua
```diff
@@ -69,14 +69,7 @@ end)
...
     if cmd[1] == 'help' then
-        local chat = windower.add_to_chat
-        local color = string.color
-        chat(1,'Trusts - Command List:')
-        chat(207,'//tru '..color('save <setname>',166,160)..' --Save trusts in current party.')
-        chat(207,'//tru '..color('<setname>',166,160)..' --Calls trusts you saved.')
-        chat(207,'//tru '..color('list',166,160)..' --Lists your saved sets.')
-        chat(207,'//tru '..color('random',166,160)..' --What\'s your fortune today?')
-        chat(207,'//tru '..color('check',166,160)..' --List of unlearned trusts. gotta catch \'em all!')
+		show_help()
```
z16: Use spaces here, to match the format of the rest of the file.

z16:
> @z16 - just moved all the other error() to add_to_chat in the process and stripped out the help function. I'd rather the error just tell you directly how to run the command correctly rather than spit the entire list of commands personally. But if it's an issue, it's now possible to call `show_help()`.

It's fine, I actually prefer this, too. Just didn't care too much either way.

> Not sure why that wasn't its own function anyways.

Doesn't really need to be, it's fine in the command handler. Not an issue that you moved it out though, so don't need to go back. Doesn't matter much either way.

PR #2413, opened 2025-01-02
[GearSwap] Differentiate between item usage on NPCs versus player.

lili-ffxi:
- Using an item on an NPC initiates a trade packet, while using an item on a player initiates an item use packet. This commit makes gearswap better aware of the difference.

- Also adds a field `range` with value of `25` to Ranged Attacks that previously would not have a `range` field in the associated `spell` table. (old change that I forgot to commit sooner)

### DIFF
--- a/addons/GearSwap/triggers.lua
+++ b/addons/GearSwap/triggers.lua
@@ -120,6 +120,7 @@ windower.register_event('outgoing text',function(original,modified,blocked,ffxi,
                 storedcommand = command..' "'..windower.to_shift_jis(r_line[language])..'" '
             elseif unified_prefix == '/ra' then
                 r_line = copy_entry(resources_ranged_attack)
+                r_line.range = '25' -- temp
                 storedcommand = command..' '
             end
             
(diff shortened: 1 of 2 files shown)


addons/GearSwap/triggers.lua
```diff
@@ -120,6 +120,7 @@ windower.register_event('outgoing text',function(original,modified,blocked,ffxi,
                 storedcommand = command..' "'..windower.to_shift_jis(r_line[language])..'" '
             elseif unified_prefix == '/ra' then
                 r_line = copy_entry(resources_ranged_attack)
+                r_line.range = '25' -- temp
```
z16: Is this meant to be `temp`?

lili-ffxi: Excellent catch. No, this is a mistake because I updated the wrong (single) file, instead of two files. This change should be in this file at this line: https://github.com/Windower/Lua/blob/48fea756025a93da4d34927ac9c883fe50e0db9f/addons/GearSwap/statics.lua#L134

it's supposed to add a `range` field to ranged attacks, which is the only case where the `spell` table does not have a range value. What's the best thing to do here? I remove this line here and attach the other file to the PR? Remove this line entirely since ranged attacks have a consistent and implicit range of 25 no matter what situation the player is in?

lili-ffxi: @z16 awaiting for your preference so I can move forward with this

lili-ffxi: Removed and moved to statics.lua where that pseudo-resource line is initially (and statically) defined.

PR #2417, opened 2025-01-10
Add debuginfo addon.

lili-ffxi:
Add `debuginfo` addon that prints some diagnostic info and looks for common windower installation problems.


addons/debuginfo/debuginfo.lua
```diff
@@ -0,0 +1,59 @@
...
+_addon.author = 'Lili'
+_addon.version = '1'
+_addon.command = 'debuginfo'
+
+require('logger')
+
+local       get_dir = windower.get_dir
+local windower_path = windower.windower_path
+local     ffxi_path = windower.ffxi_path
+local      pol_path = windower.pol_path
```
z16: I don't know why you hate me like this when I show you nothing but love.

addons/debuginfo/debuginfo.lua
```diff
@@ -0,0 +1,59 @@
+_addon.name = 'debuginfo'
+_addon.author = 'Lili'
+_addon.version = '1'
+_addon.command = 'debuginfo'
```
z16: Honestly... I wouldn't mind just calling (both the command and the addon) just `debug`, and possibly expand it to debug other issues. I could even see something like letting a user troubleshoot their own FFXI installation via guided commands. I don't mind it as it is, the `info` just feels a little extra. But I'll leave it up to you.

addons/debuginfo/debuginfo.lua
```diff
@@ -0,0 +1,59 @@
...
+
+local       get_dir = windower.get_dir
+local windower_path = windower.windower_path
+local     ffxi_path = windower.ffxi_path
+local      pol_path = windower.pol_path
+-- local launcher_path = windower.launcher_path 
+
+local dlls = { 'd3d8.dll', 'dxgi.dll', 'ddraw.dll', 'd3dimm.dll', 'd3d9.dll', '.*%.conf' }
+
+windower.register_event('addon command',function(arg)
```
z16: SPAAAAAAAAACE

addons/debuginfo/debuginfo.lua
```diff
@@ -0,0 +1,59 @@
...
+    local arg = arg:lower()
+
+    local windower_settings = windower.get_windower_settings()
+
+    local str = ''
```
z16: Use a table or list and add strings, then join at the end. Not only faster, it also avoids all those `\n` at the start of a string, which is not only harder to look at and visually parse but also impairs contextual search. You won't even have to change the `:append`s if you use a list!

addons/debuginfo/debuginfo.lua
```diff
@@ -0,0 +1,59 @@
...
+
+    -- Look for graphics enhancement dlls and config files
+    for i, folder in pairs({ ffxi_path, pol_path, launcher_path }) do
+        local files = get_dir(folder)
+
+        for _, name in pairs(files) do
+            for _, pattern in pairs(dlls) do
+                local found = name:imatch(pattern)
+                if found then
+                    str = str:append('\n%s found in %s':format(found, folder))
```
z16: In a future version, it could also check if the files are in the right folder, instead of just printing them out.

addons/debuginfo/debuginfo.lua
```diff
@@ -0,0 +1,59 @@
...
+                end
+            end
+        end
+    end
+
+    log(str)
+
+    if arg == 'copy' or not windower.ffxi.get_info().logged_in then
+        windower.copy_to_clipboard(str:enclose('```'))
```
z16: I don't like treating login screen and in-game differently. I would either always copy to clipboard, even with no argument, or always respect the arguments, and then I'd probably allow something like
```
debuginfo copy|log
```
To display in log as well as copy. Could then also add a `console` and/or `file` option, for the people who are not logged in but don't want to copy to clipboard (I often don't want my clipboard messed up).

addons/debuginfo/debuginfo.lua
```diff
@@ -0,0 +1,59 @@
...
+        end
+    end
+
+    -- Look for graphics enhancement dlls and config files
+    for i, folder in pairs({ ffxi_path, pol_path, launcher_path }) do
+        local files = get_dir(folder)
+
+        for _, name in pairs(files) do
+            for _, pattern in pairs(dlls) do
+                local found = name:imatch(pattern)
```
z16: You really want me to never deprecate that function, huh? :cry:

I think I would not pattern match here at all and just check some files fully, and some with `:endswith`. If you do want to go with pattern matching, I would lowercase the `name` here first, then use a proper function. I would also properly escape the `.` in the file list, see comment above.

addons/debuginfo/debuginfo.lua
```diff
@@ -0,0 +1,59 @@
...
+    for _,name in pairs(addons) do
+        if name:endswith('-master') then
+            if not windower.file_exists('%s%s\\%s.lua':format(addons_path, name, name)) then
+                str = str:append('\nWrong folder name: ' .. name)
+            end
+        end
+    end
+
+    -- Look for graphics enhancement dlls and config files
+    for i, folder in pairs({ ffxi_path, pol_path, launcher_path }) do
```
z16: I would use `ipairs` even though `pairs` should behave identically here, just to guarantee the order.

addons/debuginfo/debuginfo.lua
```diff
@@ -0,0 +1,59 @@
...
+    str = str:append('\n Launcher version: ' .. windower_settings.launcher_version)
+    str = str:append('\n Hook version: ' .. windower_settings.hook_version)
+    str = str:append('\nFFXI path: ' .. ffxi_path)
+    str = str:append('\n Client version: ' .. windower_settings.ffxi_version)
+
+    -- Look for cases of mismatched foldername\addonname.lua
+    local addons_path = windower_path .. 'addons\\'
+    local addons = get_dir(addons_path)
+    for _,name in pairs(addons) do
+        if name:endswith('-master') then
```
z16: I change this check to this:
```lua
if name ~= 'libs' then
```

That's because the following check is *always* valid and can find misnamed directories even outside of the `-master` issue. If a directory with name `<name>` exists, it *must* contain a `<name>.lua` inside, otherwise it's not a valid addon (unless it's the `libs` directory).

z16 (changes requested):
Aside from missing spaces and some suggestions and stuff, looks fine. Missing the license text though.

addons/debuginfo/debuginfo.lua
lili-ffxi: It's pleasing to the eye mkay!

Changed.

addons/debuginfo/debuginfo.lua
lili-ffxi: `debug` it is.

addons/debuginfo/debuginfo.lua
lili-ffxi: I did it this way because originally I wanted the option to _not_ add a newline if needed, if for whatever reason I wanted a longer message on a single line. Right now I can't come up with a reason why to keep that, so I'll do the list thing.

z16: You can still build an individual string over multiple lines of code, then add it to the table, when the line finishes.

addons/debuginfo/debuginfo.lua
lili-ffxi: On third thought tho, this doesn't seem much of a debugging tool as a diagnostic one. Should it be `diagnostic` or similar?

addons/debuginfo/debuginfo.lua
lili-ffxi: "Right folder" is highly contextual to what is happening and why, so I decided to make the tool just report the information so that an human can parse it, rather than have the tool (attempt to fix|tell the user how to fix) the issue.
Plus, the correct location, or even if they are allowed to be there at all, depends on the specific file (some are just useless, some actively cause crashes). Thoughts?

addons/debuginfo/debuginfo.lua
lili-ffxi: Fair. I'll add a log|copy|file argument, which defaults to `log`. 
I do think that `log` should print to console if the character is not logged in (and thus the chatlog is unavailable), tho, what do you prefer?

addons/debuginfo/debuginfo.lua
lili-ffxi: What comment above? 😄 

Pattern matching offers a really simple way to have partial name matches without having specific logic for it, which I like because EXPANDABLE. I don't feel strongly about it tho so I could change `dlls[]` to be a table and have a different behavior specific for each file, but it feels more clunky to me that way.

addons/debuginfo/debuginfo.lua
lili-ffxi: Implemented.

addons/debuginfo/debuginfo.lua
z16: I mean, working under the assumption that only addons and libs are under `Windower/addons/` (I guess some people could try to hide their porn there, but not sure the debug tool should cater to them), and all libs being confined to the `libs` sub directory, that means every other directory under `Windower/addons/` should be an addon. And for it to be a valid addon, there must be a file with the directory's name (with the `.lua` extension) under that directory. If that is not the case, it's *definitely* something that does not belong in there.

I suppose some people could use it for their *own* shared libraries, that they use between their own addons. If so, a) they shouldn't, and b) it's not like this tool would delete it, it would only let us know that this exists. Worst case, we tell the user to double check that, and the user tells us, that it's not a mistake, we move on. That was my thought process here.

That said, it was definitely just a suggestion and I don't care too much either way. Would just allow us to detect more naming issues than just the `-master` one. For example, if someone downloads an addon from somewhere, that does *not* use the `-master` branch. Especially now, that `main` is the default git branch name.

addons/debuginfo/debuginfo.lua
z16: I don't think it's necessary, but I wouldn't mind if you implemented it. Should definitely be able to specify the console as well though, even if the chat log is available.

addons/debuginfo/debuginfo.lua
z16: Well look at that, I forgot to add the comment above. I was talking about the `dlls` table, which is used via pattern matching, but only the last entry in there (which isn't a DLL, just saying) escapes the period character.

My idea was to have two variables, `full` and `ends` or something, one containing full matches, the other partial end-matches, but I'm not too happy about this either, so pattern matching is fine. But please no `imatch`.

addons/debuginfo/debuginfo.lua
lili-ffxi: Do you want the `console` flag to be exclusive to the `log` flag, or can they happen concurrently?

z16: I think all four should be treated equally. Ideally, I'd want all of this to work:
```
debug -- use some default, probably console, since it always works
debug log|copy
debug log|console|file|copy
```

But honestly, if it's too much work, just accepting one is fine, too. Was just an idea to allow multiple outputs. I'm not sure how often that will be used at all, so maybe it's not worth the effort.

Either way, `console` should be unrelated to `log`.

addons/debuginfo/debuginfo.lua
lili-ffxi: Right. Well, if the user has a `d3d8Zdll` file, it's still likely there's some problem there 😄 

I'm not so much a fan of having specificity like that. It's also likely we find a new use case where we want to detect files _starting_ with a certain string instead. I think I want to make it detect % and treat the string as a pattern in that case. How do you feel about that?

addons/debuginfo/debuginfo.lua
z16: Don't care too much, I think as far as users are concerned, the difference is probably insignificant. Whichever you want, if you do `diagnostic`, then make a short command `diag`, though, since `diagnostic` is too much to type :)

addons/debuginfo/debuginfo.lua
lili-ffxi: I already made all flags able to work at the same time, the only doubt was if to make console and log exclusive to each other. In the end I did not, so you can specify all 4 flags and have a chorus of information thrown at you!

addons/debuginfo/debuginfo.lua
lili-ffxi: Renamed and command changed.

PR #2422, opened 2025-01-16
Add missing trailing slash to path

lili-ffxi:


z16:
I will fix this in the launcher.

PR #2425, opened 2025-02-04
azuresets tested/updated

Neqsal:
subjob support, code changes. original rework by Exclore.


Neqsal:
Errors have persisted after long term use. I cant seem to debug this error and so I plan to eventually do a complete rewrite of azuresets.

PR #2435, opened 2025-03-07
Fixed bug in return of completed data for assault extdata

Traelblaser:
Tweaks return value of completed as follows: 
From: (buggy code)
Screenshot 2025-03-07 at 2 15 14 PM

To: (what was intended)
Screenshot 2025-03-07 at 2 16 06 PM

### DIFF
--- a/addons/libs/extdata.lua
+++ b/addons/libs/extdata.lua
@@ -1972,87 +1972,87 @@ end
 
 function decode.AssaultLog(itemid, str) 
     local missions = T{
-        [2491] = T{
+        [2491] = {
             -- "Leujaoam Log"
-            [1] = "Leujaom Cleansing",
-            [2] = "Orichalcum Survey",
-            [3] = "Escort Professor Chanoix",
-            [4] = "Shanarha Grass Conservation",
-            [5] = "Counting Sheep",
-            [6] = "Supplies Recovery",
-            [7] = "Azure Experiments",
-            [8] = "Imperial Code",
-            [9] = "Red Versus Blue",
-            [10] = "Bloody Rhondo"
-    
+            [1] = {name="Leujaom Cleansing"},
+            [2] = {name="Orichalcum Survey"},
+            [3] = {name="Escort Professor Chanoix"},
+            [4] = {name="Shanarha Grass Conservation"},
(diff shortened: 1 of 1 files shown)


Traelblaser:
No more nesting of the set inside another set
Screenshot 2025-03-07 at 2 33 43 PM

addons/libs/extdata.lua
```diff
@@ -1972,88 +1972,88 @@ end
...
+    for key, val in ipairs(flags) do
+        names[key].completed = val
+    end
+
     local rettab = {
         type='Assault Log',
-        completed = S{data},
+        completed = names,
         flags = flags,
     }
```
z16, line 2053: Not too happy with the structure of the result table. I would return something like this:
```lua
return {
    type = 'Assault Log',
    area = 'Ilrusi', -- will need to be looked up based on ID
    log = {
        { name = 'Golden Salvage', completed = false },
        ...
    },
}
```

With my above suggestion, this could be:
```lua
return {
    type = 'Assault Log',
    area = missions[item_id].name,
    log = table.range(10):map(function(index)
        return {
            name = missions[item_id][index],
            completed = flags[index],
        }
    end),
}
```

addons/libs/extdata.lua
```diff
@@ -1972,88 +1972,88 @@ end
...
-            [3] = "Escort Professor Chanoix",
-            [4] = "Shanarha Grass Conservation",
-            [5] = "Counting Sheep",
-            [6] = "Supplies Recovery",
-            [7] = "Azure Experiments",
-            [8] = "Imperial Code",
-            [9] = "Red Versus Blue",
-            [10] = "Bloody Rhondo"
-    
+            [1] = {name="Leujaom Cleansing"},
```
z16, line 1977: Would change this back to being raw strings, and assemble the `name` attribute later. Also, please move the `missions` table outside of the function and close over it, like this:
```lua
do
    local missions = {
        [2491] = { name = 'Leujaoam',
            [1] = 'Leujaoam Cleansing', -- also fix Leujaoam in this name please
            [2] = 'Orichalcum Survey',
            ...
        },
    }

    function decode.AssaultLog(item_id, str)
        -- use the `missions` variable here
    end
end
```

PR #2437, opened 2025-03-28
Added AutoDispel addon

designly1:
Addon that watches for mob abilities to be cast and then automatically casting dispel (or other specified magic or job ability).

### DIFF
--- a/addons/AutoDispel/AutoDispel.lua
+++ b/addons/AutoDispel/AutoDispel.lua
@@ -1,0 +1,132 @@
+_addon.name = 'AutoDispel'
+_addon.author = 'Jay Simons'
+_addon.version = '1.0.1'
+_addon.commands = { 'adisp' }
+
+require('luau')
+require('functions')
+require('config')
+
+defaults = T {}
+defaults.WatchFor = T {}
+defaults.DoCommand = '/ma "Dispel"'
+defaults.ExecDelay = 2
+
+settings = config.load(defaults)
+
+local do_command = string.format('input %s <t>', settings.DoCommand)
+
+local function trim(s)
+    return s:match("^%s*(.-)%s*$")
+end
+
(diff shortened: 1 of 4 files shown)


Nifim:
Add-ons that respond to non-player actions to initiate an action will not be added to the repository.

designly1:
That's an odd policy. May I ask why?

z16:
Addons that perform actions in response to non-player input (such as game events or enemy actions) fall under our definition of botting, which we don't want to support. This has been our policy since forever, although we were, admittedly, not great at enforcing it back when we initially opened the repository to the public. There are some addons in this repository that we wouldn't accept today, and we may even remove those in the future. We have already removed such addons in the past, like AEcho (which did something similar to this submission) which we [just removed a few months ago](https://github.com/Windower/Lua/commit/588ec66cc47519415094f960c471c3bfc2f4e9a9). But these days, we are pretty strict about this and we will not accept any such addons.

designly1:
Ok, well you have many addons that do just that.

Examples:
autojoin
autoinvite
enternity
gearswap

I mean... gearswap bro.. That's botting. Do you think anyone would use windower without it? lolz

Nifim:
Arcon said above: 
> although we were, admittedly, not great at enforcing it back when we initially opened the repository to the public.

pointing out examples of "hypocrisy" just gives us more reason to clean up our repository, as we did with autoechos

And gearswap does not fall under what we describe as a "bot" (performing actions in response to non-player input), some gearswap scripts can do that because the authors simply write their own code to execute on various portions of the windower API. if gearswap was a bot it would have functions such as `mob_precast`, `mob_midcast`, and `mob_aftercast` which trigger when someone other than the player takes an action.

At any rate if you want to continue this discussion you can do so in our Discord development channel (we have had it many times)

PR #2440, opened 2025-05-07
[DressUp] Update self model on job change regardless of Combat blinking setting

shastaxc:
In some situations, you can end up with your gear locked to a previous job's lockstyle even after changing jobs.

A specific example is when you have DressUp settings set to lock in Combat but not Always, during Odyssey after doing a fight and getting sent back to the lobby, the game client sometimes thinks you're still in combat even though you're in the lobby. In this case, changing gear while not having a job or character profile set up in DressUp's settings (instead preferring in-game lockstyle) your character will be stuck looking like the previous job.

I modified the if statement on the job change event to always update the model even if there's no profile to load for the new job. It still attempts to load a profile.

image

### DIFF
--- a/addons/DressUp/DressUp.lua
+++ b/addons/DressUp/DressUp.lua
@@ -85,11 +85,13 @@ windower.register_event('logout', function()
 end)
 
 windower.register_event('job change',function(job)
-    local isProfileLoaded = load_profile(res.jobs[job].name)
-    update_model(info.self.index)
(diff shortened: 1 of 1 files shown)


shastaxc:
After looking into it some more, I think my change should actually not be having any effect. Maybe my testing was not thorough enough.
I'm gonna close my PR and work on this some more.

PR #2445, opened 2025-06-04
fix scoreboard loading error: end expected to close function

Mekaider:


z16:
It is, indeed, the right target branch.

PR #2459, opened 2025-08-13
[itemizer] Refactor command logic and add `move` functionality.

lili-ffxi:
Added `move` and `moves` commands, to reflect that the vanilla client now supports moving items across non-inventory bags directly.

Refactored command logic a litle.

Added in placeholders in preparation for the future inventory management library (when we get to finalize it).

Tweaked help text slightly.

### DIFF
--- a/addons/itemizer/itemizer.lua
+++ b/addons/itemizer/itemizer.lua
@@ -5,7 +5,6 @@ _addon.command = 'itemizer'
 
 require('luau')
 local packets = require('packets')
-local libitems = {}  -- preparation for future item library
 
 defaults = {}
 defaults.AutoNinjaTools = true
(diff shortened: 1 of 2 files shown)


Nifim:
Items can be moved from recycle but not too it directly, that said i don't really have a problem with just ignoring the bag for the sake of simplicity

lili-ffxi:
> Items can be moved from recycle but not too it directly, that said i don't really have a problem with just ignoring the bag for the sake of simplicity

So it can be a source bag but not a destination bag? Huh, TIL.
That said, it's still something I'm not happy to mess with programmatically, it feels to me like it should be all manual user side.
Now I wonder if anyone relies in the functionality tho, as currently I don't think there's anything that blocks it.

PR #2464, opened 2025-08-23
Closet Cleaner updates

Fox214:
Removed print() statements, all message are add_to_chat() or to a file.  Added support for Selindrille luas
fixed some bugs reading xml file


Fox214:
Added a slips report.

PR #2467, opened 2025-09-09
[DistancePlus] Fixed an error in distance calculation for Flourishes II.

LycorisXI:
When Flourishes II is present in the ability list, DistancePlus repeatedly throws the following error in the console:
`Lua runtime error: distanceplus/distanceplus.lua:130: attempt to perform arithmetic on field '?' (a nil value)`

This happens because in `res/job_abilities.lua`, Flourishes II is defined with `range=0` and `targets=4`, causing DistancePlus to fail when calculating distances while updating the ability list.

I have updated the code so that abilities with a `range` of `0` are skipped when checking whether a target is within range.

### DIFF
--- a/addons/DistancePlus/DistancePlus.lua
+++ b/addons/DistancePlus/DistancePlus.lua
@@ -122,16 +122,17 @@ function displayabilities(distance,master_pet_distance,s,t)
     local list = 'Abilities:\n'
     if abilitylist then 
       for key,ability in pairs(abilitylist) do
-        ability_en = res.job_abilities[ability].name
+        ability_en = res.job_abilities[ability].en
+        ability_name = res.job_abilities[ability].name
         ability_type = res.job_abilities[ability].type
         ability_targets = res.job_abilities[ability].targets
         ability_distance = res.job_abilities[ability].range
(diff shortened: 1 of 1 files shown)


lili-ffxi:
This feels more like an error or mismatch in resources. Two of the Flourish II are self-target _only_, while one is enemy-target and has about 5' range.

Nifim:
It seems at this was attempted to be addressed by adding `and ability_en ~= "Flourishes II"` to the line right before the one you changed

so i wouldn't think flourishes ii should be making it into that section of the code

LycorisXI:
The `ability_en` field stores the Japanese name `フラリッシュII` in the Japanese client. Therefore, `and ability_en ~= “Flourishes II”` will not function correctly in the Japanese client.

Nifim:
Ah, now that seems like the problem then.
https://github.com/Windower/Lua/blob/3e6af24ac1388c7a18f859380ee1a98518004525/addons/DistancePlus/DistancePlus.lua#L125C9-L125C53
```lua
 ability_en = res.job_abilities[ability].name
```

this needs to be changed to get the english name of the ability `res.job_abilities[ability].en`, or we just check if the `ability == 213`  rather than checking the name.

LycorisXI:
I had overlooked that the condition `and ability_en ~= "Flourishes II"` was already present.
I changed `ability_en` to hold the English name for checks and added `ability_name` to store the display name.

Issue #2475, opened 2025-10-15
Rolltracker events go to the wrong chat category/window


SnickySnacks:
When using dual windows, rolltracker events end up in the same window as chat logs, due to using `add_to_chat(1,`

I've adjusted this locally by changing this to `add_to_chat(21,` which gives a pleasant grey color and puts it into the battle window, as I would expect.

However, I don't have a good understanding of what the various color values actually mean when used in add_to_chat.
I assume they correspond to different chat categories and thus would get filtered to window 1 or 2 as appropriate.

**Is there any information on what chat colors correspond to which chat categories so that these messages could be categorized correctly, if it even really matters?**

It's also possible that I'm misunderstanding how the chat window categorization actually works.

I can raise a PR for this (assuming the change in window behavior is wanted) but I'd like to make sure I'm not doing something silly by picking an arbitrary color here, but as it stands it doesn't make a lot of sense to me to have the buffs show up in the middle of what is, normally, just my party chat.

Any thoughts?

RubenatorX:
"puts it into the battle window, as I would expect."
There's no "battle window" and "chat window". You can in-game configure your battle messages to go in the other window if you want.
We wouldn't accept a PR for this change that is purely subjective.
If you want to do a PR, you'd need to use the `config` library to add settings and then set the default setting to the channel that it is now, and add a command to select a custom channel/mode.

SnickySnacks:
I understand what you're saying about there not being a fixed "battle" and "chat" window, but my issue is that the way the add_to_chat is currently set, it does not get grouped with other buffs and does not go to the window that I have set to display buffs, but instead ends up in the window with my say/party/yell chats.

This goes back to my question of "is there a way to determine what category the various chat colors get grouped into so that rolltracker could use the correct color to make this categorized as a buff, rather than as a normal chat"

Currently there isn't a way to fix this in-game by modifying my chat filters, as I already have them set up to *only* send say/tell/party/yell to my "chat" window, but rolltracker ends up there anyways.

Edit: I'm open to making it configurable, but my concern is mostly about picking a sane default rather than just picking a random color that seems to end up in the right window as if the color I picked (21 in this case) actually means something like "enemy attack" then people with different filter settings than mine could end up with a similar issue.

RubenatorX:
<https://github.com/Windower/Resources/blob/ee6e87bcde557915358b84b62f40c9e362a76cf0/resources_data/action_messages.lua#L423-L428>
Rolls are typically `101`
[...]

PR #2476, opened 2025-10-16
RollTracker: Settings based color customization

SnickySnacks:
Allows colors to be customized via RollTracker commands.

### DIFF
--- a/addons/rolltracker/rolltracker.lua
+++ b/addons/rolltracker/rolltracker.lua
@@ -25,7 +25,7 @@
 --SOFTWARE, EVEN IF ADVISED OF THE POSSIBILITY OF SUCH DAMAGE.
 
 _addon.name = 'RollTracker'
-_addon.version = '1.8.1.0'
+_addon.version = '1.8.1.1'
 _addon.author = 'Balloon'
 _addon.commands = {'rolltracker','rt'}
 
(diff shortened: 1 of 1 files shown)


RubenatorX:
Thanks for doing this but, can I request a few small changes?
1) I'm fine if you call the custom colors "colors" but when using the `add_to_chat` first parameter, that is a "channel" or "mode", not a color. Could you adjust the variables/verbiage to support that? thanks
As a matter of explanation, every channel has it's own color associated with it -- some of these colors are customizable via the in game config -- such as changing the color of "party" chat text, etc. So when you use coloring within text, you are using the color associated with a particular channel number. But when you use a number as the first parameter of `add_to_chat`, you are specifying which channel to use for the entire message, which determines which window selections and chat filters apply to it (among other things).
So when you do
`windower.add_to_chat(settings.color.warn,'Attempting to Doubleup on a Lucky Roll: Re-double up to continue.')`
you are not selecting a color, you are selecting a channel (which happens to have a color associated with it).
whereas when you do:
`luckChat = string.char(31,settings.color.lucky).." (Lucky!)"`
you are coloring the text.
2) I see you're using `string.char` to color stuff -- the official way to do that is with `string.color` via the `chat` library:
<https://github.com/Windower/Lua/blob/6bed1ba3234ea6999ccd7ee1651b9e9aeae6eeb2/addons/libs/chat.lua#L48-L59>
For example:
"(Lucky!)":color(settings.color.lucky)
Using this function also properly uses the color end tag character

SnickySnacks:
Ah, yeah I was just copying what was there.
I'll update it with the requested changes when I get a minute.

Thanks for the feedback.

SnickySnacks:
@RubenatorX  How about this?

RubenatorX:
> @RubenatorX How about this?

Looking really good! I'll have to test it myself for a bit! :)

lili-ffxi:
<Rubenator> oh uh, I forgot it existed and it needs tested but, probably safe to toss onto dev

tossing onto dev so it can be field tested. Also resolves #2475

PR #2481, opened 2025-11-14
[gearswap] Fix player.target.type being not set sometimes.

lili-ffxi:
Let target_complete() deal with being passed `nil` tables directly, so it can properly set the `player.target.type` field when the specified target does not return any mob.


Nifim:
Link to discord discussion: https://discord.com/channels/338590234235371531/501099842097905675/1439001753482821795

PR #2499, opened 2025-12-12
[GearSwap] Only do item trade check once per code path

Zirk40:
At present for the pretarget code path, the bit.band spawn type check gets performed twice, once in outgoing text, and again in pretarget. We should only need to perform this check once. In addition, the determination if the current action type is a trade or not is only performed in outgoing text, before the userfile might change the target to something different in pretarget. If the target did get changed in pretarget to something with a different spawn type, the initial trade action type that was determined in outgoing text might no longer be correct. This would be the case if the initial target was an NPC, and the new target was non-NPC, or vice versa. So this commit determines the item action type for the pretarget code path only after the target may have been changed to ensure it is correct, and moves the outgoing text item action type check to inside the target ID code path to avoid it being performed twice for pretarget code path.

### DIFF
--- a/addons/GearSwap/flow.lua
+++ b/addons/GearSwap/flow.lua
@@ -228,6 +228,7 @@ function equip_sets_exit(swap_type, ts, val1)
                 if val1.prefix == '/item' then
                     -- Item use packet handling here
                     if bit.band(val1.target.spawn_type, 2) == 2 and find_inventory_item(val1.id) then
+                        val1.action_type = 'Trade'
                         -- 0x36 packet
                         if val1.target.distance <= 6 then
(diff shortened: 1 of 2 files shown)


lili-ffxi:
As discussed, the idea is solid, but as it is this commit would result in `spell.action_type` being incorrect during the pretarget user event when using a command like `/item "Mars Orb" <t>` on an npc, which isn't exactly ideal. If we decide that a certain type of target results in an action that is a Trade, and not an item use, then it should be labeled correctly during all user events.

Zirk40:
I agree pretarget user event needs to be able to see if the current target is making the action type a trade, as that might be the specific piece of info they are checking to determine if target needs to be changed or cancel spell etc.

PR #2503, opened 2025-12-14
[rollracker] Update  version and enhance HUD features

aragan:
Updated version and author information, added HUD functionality for displaying roll information, and improved command handling.

image

### DIFF
--- a/addons/rolltracker/rolltracker.lua
+++ b/addons/rolltracker/rolltracker.lua
@@ -25,14 +25,57 @@
 --SOFTWARE, EVEN IF ADVISED OF THE POSSIBILITY OF SUCH DAMAGE.
 
 _addon.name = 'RollTracker'
-_addon.version = '1.8.1.1'
-_addon.author = 'Balloon'
+_addon.version = '1.9.0.0'
+_addon.author = 'Balloon, Aragan' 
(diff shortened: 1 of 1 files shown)


z16:
We cannot accept the PR in this state. You need to make sure that existing functionality is not changed, to avoid disruption to users. You should re-add all the deleted chat output code, and work your new code into that. The HUD should be toggleable, and ideally default to `off`.

Further, you need to remove the `_addon.update` field, as that table only has specific fields you should assign to, and this is not one of them.

You should also stick to the established code style within that file. Remove the `rt_` prefix from names, and use camelCase, as the rest of the file does. Preserve the command handler structure and the help text. Remove excessive comments and condense multiple blank lines in a row to a single blank line. This goes for all submissions to our repo.

Nifim:
I assume you are still working on this, but in case you are waiting for more feedback.

You still need to fix the command handler to put it back to the way it was, you are adding functionality try to minimize the changes so that all you are doing is adding code and avoid changing what was already there.

You also need to put all the settings you remove back, the end user is using these settings to control the existing functionality for printing to the chat log.

Nifim:
Closings due to inactivity, we can reopen if you address the issues mentioned in the discussion

PR #2505, opened 2025-12-17
Proposed addon: useall

lili-ffxi:
With this addon enabled, if the user is holding down a modifier key while using an item (default: shift), then the addon will try to use all copies of that item that are present in the user's inventory. This is similar to the functionality of `pouches`, but it's activated by using the item rather than by typing the item name.

Future planned functionality:
- `//useall @current` - leverage `windower.get_item_display()` to use all copies of the item currently highlighted by the user in their inventory  
- `//useall itemname` - same functionality as pouches

Starting the PR for the purpose of code review/considerations on what to do with an addon that is effectively duplicating the functionality of another existing one; I'll add a readme and the entry in addons.xml at a later point.

### DIFF
--- a/addons/useall/useall.lua
+++ b/addons/useall/useall.lua
@@ -1,0 +1,131 @@
+--Copyright © 2025, Lili
+--All rights reserved.
+
+--Redistribution and use in source and binary forms, with or without
+--modification, are permitted provided that the following conditions are met:
+
+--    * Redistributions of source code must retain the above copyright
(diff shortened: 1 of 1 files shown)


aragan:
lili u steal my idea addon autocoffer and convert it to u addon pls respect rights

z16:
As we explained in Discord, this addon has been around for almost a year. It was just never submitted to this repository. The code and mechanics are also very different, so no one is ripping off anyone here. We will review both addons.

PR #2506, opened 2025-12-17
create new addon autocoffer

aragan:
AutoCoffer automatically opens all similar containers (Coffer / Pouch / Sack / Case / Box / Parcel / Codex) from the main inventory when you manually use one of them once.


Nifim:
Too similar to existing addon

PR #2507, opened 2025-12-17
Proposed addon: atreplace

lili-ffxi:
Adds the `atreplace` addon, which lets the user input autotranslate terms by using the `_(string)` syntax in their outgoing text.

The `//atreplace copy <string>` functionality seems to not work for all terms, probably due to an encoding issue?

### DIFF
--- a/addons/atreplace/atreplace.lua
+++ b/addons/atreplace/atreplace.lua
@@ -1,0 +1,136 @@
+--Copyright © 2024, Lili
+--All rights reserved.
+
+--Redistribution and use in source and binary forms, with or without
+--modification, are permitted provided that the following conditions are met:
+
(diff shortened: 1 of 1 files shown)


Nifim:
as an english speaker `atreplace` reads as `at replace`. I have no suggestions for a better name, but feel like it could benefit from a name that more clearly conveys the purpose.

lili-ffxi:
> as an english speaker `atreplace` reads as `at replace`. I have no suggestions for a better name, but feel like it could benefit from a name that more clearly conveys the purpose.

Well, it does stand for AutoTranslate Replace, so `at replace` is technically not incorrect. I'm open for better suggestions but I myself couldn't really come up with anything else :-\

PR #2508, opened 2025-12-19
Add subjob support and update version to 1.24.3

Gojito:
Updated version to 1.24.3 and added subjob support for Blue Mage spells. Enhanced functions for job detection and spell management.


Nifim:
Wrong target branch, need to target `Windower:dev` not `Windower:live`

PR #2510, opened 2025-12-20
[AutoInvite] add mode freeinv Enhance AutoInvite functionality and up…

aragan:
…date version

[AutoInvite] add mode freeinv Updated AutoInvite to version 1.3 with enhanced features and commands.

### DIFF
--- a/addons/autoinvite/autoinvite.lua
+++ b/addons/autoinvite/autoinvite.lua
@@ -26,12 +26,24 @@ ON ANY THEORY OF LIABILITY, WHETHER IN CONTRACT, STRICT LIABILITY, OR TORT
 SOFTWARE, EVEN IF ADVISED OF THE POSSIBILITY OF SUCH DAMAGE. 
 ]]
 
+
+--[[
+    AutoInvite (Enhanced Version)
+    Author: Registry + Modified by Aragan
+    Description:
+      - AutoAdd is enabled by default (invites anyone who sends you a /tell)
+      - Does not invite someone already in the party
+      - Supports on/off/status commands
+      - Fully English messages
+      - All original commands are available (whitelist / blacklist / keywords)
+]]
+
 require('luau')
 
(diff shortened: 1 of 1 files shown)


z16:
Don't mess up the formatting, don't add needless command descriptions, don't call it "Enhanced Version" and don't change the order of functions for no reason, it makes the diff hard to read, and don't update versions with random numbers, a `1.0` should be followed by a `1.1`. Open a new PR with a proper adjustment, so I can review it.

aragan:
> Don't mess up the formatting, don't add needless command descriptions, don't call it "Enhanced Version" and don't change the order of functions for no reason, it makes the diff hard to read, and don't update versions with random numbers, a `1.0` should be followed by a `1.1`. Open a new PR with a proper adjustment, so I can review it.

It was released 11 years ago. I hope you understand
It is missing many modifications

Nifim:
You must make focused changes so that we can review them. If you change every line when you simply want to add this feature, you make it very hard to see what was changed for the purpose of adding the feature rather than a seemingly random decision to change something completely unrelated.

A glaring example of this is why did you change the command prefix? Can you even justify doing such a thing? As you said this is 11 years old people are using that prefix you can't go and change it without reason

aragan:
> You must make focused changes so that we can review them. If you change every line when you simply what to add this feature you make it very hard to see what was changed for the purpose of adding the feature rather than a seemingly random decision to change something completely unrelated.
> 
> A glaring example of this is why did you change the command prefix? Can you even justify doing such a thing? As you said this is 11 years old people are using that prefix you can't go and change it without reason

I added the mod without changing the old code , and the add-on itself didn't change; only the mod was added.

prefix  The feature you're referring to conflicts with some of the addon I use, such as autoitem , ai.

This can be addressed in another PR or by updating this request.

Nifim:
An addon outside the official repo is not a good justification to change the prefix of autoinvite.

PR #2511, opened 2025-12-21
Fixing packet fields bugs

kaytecat:
Fixed the following packets fields:

Outgoing 0x05B
Corrected the option index from a short to an int
Updated the _unknown2 name and comment to reflect that it's actually upper bit padding for a short that includes the 'Automated Message' bool (all 16 bits will either be a 1 or 0 value, but technically could be any value). Discussion on discord suggested to keep this as is to not break addons, but really the bool and char should be combined into a single short called 'Mode' similar to below.

Outgoing 0x05C
Corrected the _unknown1 value to its actual usage, an option index mimicing the usage in 0x05B.
Updated the _unknown2 name to 'Mode' due to not being limited by previously called 'Automated Message' bool, but can change this to match the above bool+char if desired (though it would lock into this paradigm on both).

Incoming 0x033
Corrected the ctype of the menu parameters from char[32] to data[32] as the other packets use, which also fixes issues when there is a 0 value that would terminate processing of the string.

Incoming 0x0BF
Added missing packet data for instance reservations. Description already present in data.lua.

### DIFF
--- a/addons/libs/packets/fields.lua
+++ b/addons/libs/packets/fields.lua
@@ -3457,7 +3457,7 @@ fields.incoming[0x0B5] = L{
 }
 
 -- Reservation Response
-packets.raw_fields.incoming[0x0BF] = L{
+fields.incoming[0x0BF] = L{
     {ctype='unsigned short',    label='_unknown1'},                             -- 04
     {ctype='unsigned short',    label='Result'},                                -- 06   Success is 4, other values are failures
     {ctype='unsigned int',      label='_unknown2'},                             -- 08
--- a/addons/libs/packets/fields.lua
+++ b/addons/libs/packets/fields.lua
@@ -672,8 +672,9 @@ fields.outgoing[0x05B] = L{
     {ctype='unsigned int',      label='Target',             fn=id},             -- 04
     {ctype='unsigned int',      label='Option Index'},                          -- 08
     {ctype='unsigned short',    label='Target Index',       fn=index},          -- 0C
-    {ctype='bool',              label='Automated Message'},                     -- 0E   1 if continuing the current menu, 0 if ending
-    {ctype='unsigned char',     label='_padding'},                              -- 0F   Upper bits of a Mode short including the previous bool
+    {ctype='bool',              label='Continue',                               -- 0E   Determines which mode the menu uses
+                                alias='Automated Message'},
+    {ctype='unsigned char',     label='_padding'            const=0},           -- 0F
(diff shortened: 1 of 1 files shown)


addons/libs/packets/fields.lua
```diff
@@ -670,11 +670,10 @@ fields.outgoing[0x05A] = L{
...
 fields.outgoing[0x05B] = L{
     {ctype='unsigned int',      label='Target',             fn=id},             -- 04
-    {ctype='unsigned short',    label='Option Index'},                          -- 08
-    {ctype='unsigned short',    label='_unknown1'},                             -- 0A
+    {ctype='unsigned int',      label='Option Index'},                          -- 08
     {ctype='unsigned short',    label='Target Index',       fn=index},          -- 0C
-    {ctype='bool',              label='Automated Message'},                     -- 0E   1 if the response packet is automatically generated, 0 if it was selected by you
-    {ctype='unsigned char',     label='_unknown2'},                             -- 0F
+    {ctype='bool',              label='Automated Message'},                     -- 0E   1 if continuing the current menu, 0 if ending
+    {ctype='unsigned char',     label='_padding'},                              -- 0F   Upper bits of a Mode short including the previous bool
```
RubenatorX: Let's not rename something to padding when we know it's not padding.

kaytecat: It is padding due to the choice to use the first 8 bits of the 16 bit value as the bool, so the other half will always be unused padding for that value as implemented currently. Ideally we'd combine this value and the previous bool into a short, but given that would break things, understandable to keep them separate?

Byrth: I assume you reached this conclusion by modifying `_unknown2` and observing that the Automated Message boolean didn't work properly anymore?

It's weird that Mode is 1 byte in 0x05C and 2 bytes in 0x05B.

kaytecat: It is indeed, but that seems to be what the data structs are. Kind of silly they even gave it 16 bits in the first place if they were only going to use 1 and 0 values (and 0x05C only ever sends 1) but SE gonna SE.

Honestly I guess just to make it explicit I could const the padding to 0 just to make sure people don't muck with the upper bits of the short as well.

kaytecat:
Per Discord conversation, renamed the Automated Message field to Continue and standardized between 0x05B and 0x05C. Uses the new alias feature just added by Arcon. Removed the comment on the padding and const it to 0 so it doesn't get used.

PR #2514, opened 2025-12-21
Proposed addon: DoorHelper

aragan:
DoorHelper is a Windower 4 addon that automatically interacts with nearby doors open


z16:
We don't want to have an addon that automatically opens doors in the main repository. Such addons have been made in the past, and were deliberately not included.

PR #2516, opened 2025-12-22
[autocontrol] add multiple commands for equipset

aragan:
add more options shortcut for command equipset to e and equip


Nifim (changes requested):
This is a good change, just need to do 1 more thing. 

In the section that prints the help text update it to show the new short command additions.

PR #2519, opened 2026-01-10
Update AutoRA version and modify commands

aragan:
Update AutoRA version and modify commands

windower.send_command('bind ^d ara toggle')
windower.send_command('bind ^f ara haltontp')

to save not use more binds


Nifim:
Denied, you can't change user facing feature like the existing binds without a good reason

PR #2524, opened 2026-02-03
Correct label for Hex Index

lili-ffxi:
Changes mentions of `Hex ID` to the more correct label of `Hex Index`


z16:
This was deliberately done this way to keep parity with various sources like wikis and old forums that refer to an "ID" for things like placeholder hex indices. Also, this would be a breaking change.

PR #2526, opened 2026-02-17
DistancePlus: Enhancements and Bug Fixes. v. 2.0.1

voliathon:
Below is a summary of the bugs fixed and enhancements implemented compared to the original version of the script.

**Enhancements**
- Visual Text Labels: The distance display now includes text descriptions (e.g., "14.50 Distance Penalty") instead of just a number, making it immediately clear what your current status is.
- "Critical Penalty" Zone: Added a Red warning zone for being too close (Point Blank) to the target, which causes severe accuracy/damage penalties often overlooked by players.
- Accessibility Improvements: * Changed the "True Shot" color from dark blue (unreadable) to Deep Sky Blue (readable).
- Fixed the "Yellow" color code to ensure it appears visible in the chat log.
- Performance Throttling: The math calculations now run 10 times a second instead of 60 times a second. This reduces CPU load by ~83% while keeping the display visually smooth.
- New Help Menu: Completely rewrote //dp help to include a color legend and tips on Weapon Skill mechanics (Physical vs. Magical distance rules).

**Bug Fixes & Code Cleanup**

- Global Variable Pollution: Fixed a critical issue where variables like settings, distance, and self were declared globally. They are now local, preventing crashes or conflicts with other Windower addons.
- Memory Optimization: The displayabilities function now uses table insertion instead of string concatenation, preventing unnecessary garbage collection cycles and memory churn.
- Logic Refactor: Removed duplicate, copy-pasted code blocks for Gun, Bow, and Xbow logic. These are now handled by a single, smarter get_ranged_status function, making the code cleaner and easier to maintain.
- Static Table Fix: Moved the range_mult table definition outside of the main loop so it is created only once upon load, rather than being recreated every single frame.

### DIFF
--- a/addons/DistancePlus/DistancePlus.lua
+++ b/addons/DistancePlus/DistancePlus.lua
@@ -1,5 +1,6 @@
 --[[
 Copyright © 2017, Sammeh of Quetzalcoatl
+Copyright © 2026, Voliathon of Bahamut
 All rights reserved.
 
 Redistribution and use in source and binary forms, with or without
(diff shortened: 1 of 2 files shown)


RubenatorX:
I'm gonna take a closer look at this tomorrow but, before I go to bed, I'm gonna point out the most important issue I saw skimming --
You can't just remove the model size from the distance calculation. It's an essential part of accurate distance math. Especially when you're fighting a giant dragon instead of a goblin or a chigo

RubenatorX:
Also, do you have any evidence for this? 
> "Critical Penalty" Zone: Added a Red warning zone for being too close (Point Blank) to the target, which causes severe accuracy/damage penalties often overlooked by players.

Frankly it sounds made up to me (did you use AI?)
Or are we only talking about ranged attacks?

z16:
What Rubenator said, also you use some very values that, I suspect, were made without regard for model size, namely the `RangedData` table.

Fair warning, we would *really* appreciate it if you split this up into various separate PRs. Long PRs like this tend to not get accepted, because the work on them takes very long, and often the involved people lose the motivation to keep going.

z16:
Also, please use a consistent code style. You use `snake_case`, `camelCase`, `PascalCase` and `squishedcase` throughout the code, which makes it really disorienting to read.

voliathon:
Further testing will go into the models size and RangedData.

[https://classicffxi.fandom.com/wiki/Distance](https://classicffxi.fandom.com/wiki/Distance) - Take notice of the steeper penalty being up close when using Ranged Physical Weaponskills and Ranged Attacks when up close in the Point Blank range.  This add-on does not take into any consideration and never has Sharpshot removing the distance penalty. 

Understood on the consistend code style. None was taken in the original sauce. I didn't either until just mentioned.

voliathon:
Closed due to wrong information provided in regards to Distance and other issues.

PR #2529, opened 2026-03-28
[EquipViewer] Add draggable option for UI elements

aragan:
Add draggable option for UI elements

### DIFF
--- a/addons/equipviewer/equipviewer.lua
+++ b/addons/equipviewer/equipviewer.lua
@@ -104,6 +104,7 @@ local defaults = {
     hide_on_zone = true,
     hide_on_cutscene = true,
     left_justify = false,
+    draggable = false,
 }
 settings = config.load(defaults)
 config.save(settings)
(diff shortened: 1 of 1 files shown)


RubenatorX:
It's helpful if you talk to an addon Author before submitting a PR, but I'll let you know what I was planning and what I need changed in order to accept it.
Also, in the future please always include the name of the addon/thing you're changing in square braces at the start of your PR (in this case) `[EquipViewer]` (I have edited the title as such).
1) You wrote an entire function, `position_all_ui_elements`, that is basically a non-destructive duplication of `setup_ui`. Instead, should add an optional flag to setup_ui to make it nondestructive, and call it with the flag.
2) I won't accept `draggable` as being a save-able setting. it should be a temporary addon state, toggleable with a command `move` instead (instead of drag/draggable), and only save the config after it is toggled off. Also should not have an off/on command option. Relevant instructions and feedback should be conveyed to the user when this is toggled.

PR #2532, opened 2026-04-16
Remove all spaces from "Addon Name" field for all addons

GoblinGuide:
There was a Windower version change in late February 2026 that required users to type the content of the "addon name" field of the addon to unload an addon with `lua u AddonName` (rather than typing the file name of the addon.) Doesn't look like it was reverted with that whole update-and-rollback, so **this is a functional change**, but I think new players shouldn't get this experience:
Screenshot 2026-04-16 120824

**Affected addons**: AutoEnterKey, BattleStations, EmpyPopTracker, JobChange, MountMuzzle, NoCampaignMusic, XIVBar
**Changed code:** Literally just the spaces in the "addon name" fields.
...and also Github appears to have strong opinions about the final lines of MountMuzzle and BattleStations, so I assume it's some kind of carriage return thing I don't understand becaues I'm not very bright.


RubenatorX:
We have a large patch to windower coming very soon pending some final bugfixes that will resolve this issue entirely.

PR #2537, opened 2026-05-16
Rename addon to 'xivbar'

EvanSwinehart:
Changes the name of the addon to match the folder structure, so that loading and unloading can be done with the same name


Nifim:
This is a bug that has been fixed on dev

PR #2538, opened 2026-05-16
Add wineprefix check for Linuxfix, fix misnamed windows key

Kilgannon2113:
Added a wineprefix check to hopefully prevent future troubleshooting on windows. Corrected an old bug where the windows key was misnamed as win.

There may still be an issue with the ; key but not sure what the correct name is for it.


Kilgannon2113:
hmm didn't increment version.

z16:
It appears this addon is not needed anymore when playing on Wine prefixes we recommend, so I will close this, and we will probably remove the addon itself soon.

PR #2539, opened 2026-05-17
findAll: defer slips initialization (~20MB RAM savings)

christopherkmoore:
## Summary
- Defers `slips.storages:map()` from module load time to first `search()` call
- Prevents ~20MB of `res.items` entries from materializing into memory at addon load
- Result is cached after first search, so subsequent searches have no additional cost

## Problem
At module level (line 195-198), findAll iterates every storage slip ID and accesses `res.items[id]` for each to build `merged_storages_orders`. This forces Windower's lazy-loading resource system to materialize hundreds of item entries into memory (~20MB).

This data is only needed when performing a search, not at idle.

## Solution
Wrap the slips order initialization in `ensure_merged_storages_orders()`, called at the top of `search()`. Built once on first search, cached for subsequent searches.

## Impact
- Idle memory: ~27MB → ~700KB
- First search: ~235ms one-time cost for cold `res.items` lookups (imperceptible)
- Subsequent searches: no change
- Multiboxers (6+ clients): saves ~150MB+ total RAM

## Testing
- Verified `//findall` searches return correct results across all storage types including slips
- Verified tracker functionality unaffected
- Benchmarked cold `res.items[id]` lookups at ~0.2ms each (1000 lookups in 235ms)

### DIFF
--- a/addons/findAll/findAll.lua
+++ b/addons/findAll/findAll.lua
@@ -192,12 +192,20 @@ storages_order         = S(res.bags:map(string.gsub-{' ', ''} .. string.lower ..
 
     return index1 < index2
 end)
-local storage_slips_order = slips.storages:map(function(id)
-    return 'slip ' .. res.items[id].english:lower():match('^storage slip (.*)$')
-end)
-merged_storages_orders = storages_order + storage_slips_order + L{'key items'}
+-- Deferred: slips iteration accesses res.items[id] for every slip, materializing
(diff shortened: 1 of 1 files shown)


christopherkmoore:
I didn't do too much testing on this other than works on my machine. Claude helped me clean this up and it seems sensible with much better memory footprint after his change -- so I'll leave it up to you all to decide.

PR #2545, opened 2026-06-06
DressUp: Fix client freeze on invalid face replacements, More face ID, and transition help menu to UI window

voliathon:
**Description**: This PR addresses a fatal client crash bug, expands available Face ID presets, and replaces the chat-spamming help commands with a clean, draggable in-game UI.  It was difficult to understand the previous help that displayed in the console. More instructions meant losing view of addon help in the console.

Key **Changes**:

1. Client Crash Prevention: Fixed an infinite loop in the //du replacements face <old> <new> command parser. When users entered an invalid face identifier (e.g., //du replacements face 3b 9a), the while #_models ~= 2 loop would run indefinitely, completely freezing and crashing the PlayOnline/FFXI client. Adding validation and immediate returns resolves this.
2. Face ID Mapping Additions: Added explicit lowercased keys for all standard faces. Extends the face ID selection ranges inside DressUp past the standard 1-16 (1a-8b), Fomor (29), and mannequin (30) options to support raw Face IDs up to 35 (e.g., matching the extended HumeM face lists and other expanded models available in the Place section).
_Importance_: Corrects a critical client-crashing bug. Previously, supplying a raw value higher than the FFXI client limits or outside the hardcoded ranges could cause the model engine to panic, triggering a client crash. This code caps raw inputs safely at 35 and yields an error message if an out-of-bounds ID is supplied, allowing players to safely utilize all available client face assets.
**//du self face 1-35     <--Allows for more faces to display in game beyond mannequin and fomor.**

3. UI Help Window: Replaced the heavy console printout for //du or //du help with a draggable on-screen text box using the Windower texts library. Toggling the help command now shows/hides this window, leaving the player's active chat log clean.

### DIFF
--- a/addons/DressUp/DressUp.lua
+++ b/addons/DressUp/DressUp.lua
@@ -26,11 +26,12 @@
 
 _addon.name = 'DressUp'
 _addon.author = 'Cair'
-_addon.version = '1.31'
+_addon.version = '1.4'
 _addon.commands = {'DressUp','du'}
 
 
 packets = require('packets')
+texts = require('texts')
 require('luau')
 require('helper_functions')
 require('static_variables')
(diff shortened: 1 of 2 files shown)


Nifim:
Denying this primarily for the addition of the UI for displaying the help text.
but also the change from the set to a series of ands for the face id condition is not a good choice you end up calling tonumber 3 times instead of once.

PR #2557, opened 2026-07-19
GilTracker: add separator setting and periodic refresh

FFNeko:
Add configurable gil separators and periodic refresh support for updates not covered by existing inventory events.

### DIFF
--- a/addons/giltracker/giltracker.lua
+++ b/addons/giltracker/giltracker.lua
@@ -28,7 +28,7 @@
 
 _addon.name = 'giltracker'
 _addon.author = 'sylandro'
-_addon.version = '1.0.0'
+_addon.version = '1.0.1'
 _addon.language = 'English'
 
 config = require('config')
@@ -51,6 +51,7 @@ local is_hidden_by_key = false
 
 defaults = {}
 defaults.hideKey = SCROLL_LOCK_KEY
+defaults.thousandsSeparator = ','
 defaults.gilText = {}
 defaults.gilText.bg = {}
 defaults.gilText.bg.alpha = 100
@@ -163,6 +164,12 @@ windower.register_event('keyboard', function(dik, down, _flags, _blocked)
     toggle_display_if_hide_key_is_pressed(dik, down)
 end)
 
+windower.register_event('time change', function()
+    if inventory_loaded then
+        update_gil()
+    end
+end)
+
 function ready_if_valid_treasure_packet(packet_data)
     local p = packets.parse('incoming',packet_data)
     if (p.Count > 0) then ready = true end
@@ -192,7 +199,7 @@ end
 
 function update_gil()
     local gil = windower.ffxi.get_items('gil')
-    gil_text:text(comma_value(gil))
+    gil_text:text(format_value(gil))
 end
 
 function show()
@@ -205,13 +212,36 @@ function hide()
     gil_image:hide()
 end
 
-function comma_value(amount)
-    local formatted = tostring(amount)
-    while true do
-        formatted, k = string.gsub(formatted, "^(-?%d+)(%d%d%d)", '%1,%2')
-        if (k==0) then break end
+function format_value(amount)
+    local value = tostring(amount)
+    local separator = settings.thousandsSeparator
+
+    if separator == 'space' then
+        separator = ' '
+    elseif separator == 'none' then
+        separator = ''
+    elseif separator == nil or separator == '' then
+        separator = ','
     end
-    return formatted
+
+    local sign = ''
+    if value:sub(1, 1) == '-' then
+        sign = '-'
+        value = value:sub(2)
+    end
+
+    local first_group = #value % 3
+    if first_group == 0 then
+        first_group = 3
+    end
+
+    local formatted = value:sub(1, first_group)
+
+    for index = first_group + 1, #value, 3 do
+        formatted = formatted..separator..value:sub(index, index + 2)
+    end
+
+    return sign..formatted
 end
 
 function is_cutscene(status_id)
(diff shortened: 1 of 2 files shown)


addons/giltracker/giltracker.lua
```diff
@@ -205,12 +212,31 @@ function hide()
...
-    while true do
-        formatted, k = string.gsub(formatted, "^(-?%d+)(%d%d%d)", '%1,%2')
-        if (k==0) then break end
+function format_value(amount)
+    local value = tostring(amount)
+    local separator = settings.thousandsSeparator
+
+    if separator == 'none' then
+        return value
+    elseif separator == 'space' then
```
RubenatorX: The setting really probably should accept the separator character to use, not just the name of the character to use. meaning `,` should be valid.
And, please use a map:
```lua
--somewhere near the top
local seperator_map = {
  none = '',
  period = '.',
  comma = ',',
}

function get_seperator()
  return seperator_map[settings.thousandsSeperator] or settings.thousandsSeperator
end

--somewehere
local seperator = get_seperator()

File: addons/giltracker/giltracker.lua [hunk 2]
```diff
@@ -205,12 +212,31 @@ function hide()
...
+        separator = ' '
+    elseif separator == 'period' then
+        separator = '.'
+    elseif separator == 'comma' or separator == '' or not separator then
+        separator = ','
     end
+
+    local first_group = #value % 3
```
RubenatorX: You took the nice gsub call and replaced it with a bunch of math and stuff.
I also hate the while loop so.. I'd probably do something more like:
```lua
function format_value(amount)
  return tostring(amount):reverse():gsub("%d%d%d", "%1"..get_seperator()):gsub(get_seperator()..'$',''):reverse()
end
```
although this is acceptable:
```lua
function format_value(amount)
  local formatted = tostring(amount)
  local k
  while k ~= 0 do
      formatted, k = string.gsub(formatted, "^(-?%d+)(%d%d%d)", '%1'..get_seperator()..'%2')  end
  end
  return formatted
end
```
or
```lua
function format_value(amount)
  local formatted = tostring(amount)
  while true do
      formatted, k = string.gsub(formatted, "^(-?%d+)(%d%d%d)", '%1'..get_seperator()..'%2')
      if (k==0) then break end
  end
  return formatted
end
```

addons/giltracker/giltracker.lua
```diff
@@ -163,6 +164,12 @@ windower.register_event('keyboard', function(dik, down, _flags, _blocked)
     toggle_display_if_hide_key_is_pressed(dik, down)
 end)
 
+windower.register_event('time change', function()
```
RubenatorX: This event ultimately results in this being called faster than once every 5 seconds.
I'd probably prefer a timed loop, with a configurable wait time (meaning, a setting).

RubenatorX (commented):
In addition to the code comments, please update the README :D
Thank you!

addons/giltracker/giltracker.lua
FFNeko: Updated to use a separator map while preserving literal custom separators

addons/giltracker/giltracker.lua
FFNeko: Updated the formatter to use the suggested gsub approach

addons/giltracker/giltracker.lua
FFNeko: Replaced time change with a configurable timed refresh loop

FFNeko:
> In addition to the code comments, please update the README :D Thank you!

README updated with the new separator and refresh settings

addons/giltracker/giltracker.lua
```diff
@@ -108,6 +128,15 @@ local gil_text = texts.new(settings.gilText)
 local inventory_loaded = false
 local ready = false
 
+local function refresh_gil_loop()
+    while true do
+        coroutine.sleep(math.max(tonumber(settings.refreshInterval) or defaults.refreshInterval, 1))
```
RubenatorX, line 133: Have you tested if this implodes on char-select/title screen on logout? (or zoning?)

addons/giltracker/giltracker.lua
```diff
@@ -28,7 +28,7 @@
 
 _addon.name = 'giltracker'
 _addon.author = 'sylandro'
-_addon.version = '1.0.0'
+_addon.version = '1.0.1'
```
RubenatorX: Up to you but, as this is a feature add and not just a bugfix, `1.1.0` is probably more appropriate but, not gonna stop me from accepting it.

addons/giltracker/giltracker.lua
```diff
@@ -205,12 +235,22 @@ function hide()
...
+    local k
+
+    if separator == '' then
+        return formatted
+    end
+
     while true do
-        formatted, k = string.gsub(formatted, "^(-?%d+)(%d%d%d)", '%1,%2')
-        if (k==0) then break end
+        formatted, k = string.gsub(formatted, '^(%d+)(%d%d%d)', function(prefix, group)
```
RubenatorX: Seems like this could/should be a string instead of a function no? (I can see why it coulda/woulda been a function for debugging)

RubenatorX (commented):
Thanks! just a few more minor things to poi0nt out

addons/giltracker/giltracker.lua
FFNeko: Updated version

addons/giltracker/giltracker.lua
FFNeko: Changed to a replacement string and escaped % so custom separators remain literal

addons/giltracker/giltracker.lua
FFNeko: Tested zoning, character select, multiple character logins, and logout. Loading... appears briefly during inventory initialization and clears normally. No errors observed.

PR #2560, opened 2026-08-22
feat: make send @all / @others work when logged out

Byrth:
When I use `@all` or `@others`, it doesn't affect players that are logged out.

That's kind of wacky, because even an unknown (logged out) character is a member of the set of all / other characters.


lili-ffxi (approved):
Looks good to me. I guess eventually it'd be nice to have a `@logged` pronoun to target only logged in characters, but today it's not that day.

PR #2564, opened 2026-09-11
fix: a bunch of small gearswap things

Byrth:
This includes a bunch of small fixes. I'm opening it now so I don't accidentally lose it to someone else's merge, but I'd like to leave it running for a few days before I conclude it's shippable.

Behaviorally, there are two changes that aren't pure bugfixes:
1. pet.tp was sometimes on [0,300] and sometimes on [0,3000] based on the path it took through the code. I standardized it on [0,3000].
2. windower.send_command was always prefixing with @ instead of only doing it when the user didn't provide it. I don't _think_ that will matter, but it could be that it caused bugs in weird situations and required users to work around it, and their (now unnecessary) workarounds may now cause unexpected behavior.


Byrth:
This seems good to go now

